PR #1, opened 2018-01-28
Linux fixes

vbmithr:
I fixed the build under Linux (modernized the autoconf files a bit), and added a pkg-config file.
Archlinux package here: https://aur.archlinux.org/packages/ta-lib-git/

* * *

Since there are a lot of space changes in the PR, to make it easier to review, the following way is suggested.

```
git fetch origin pull/1/head:fix-build
git checkout fix-build
git diff --ignore-space-change 48f5ed37b6e4b66eba6a9ab1dee78b23607117d9 HEAD
```


mckelvin:
Hey Vincent - this repo was just a mirror of the original SVN repo on SF. I'm not planning to maintain the project. But since the upstream is frozen for a long time, I'd like merge some compatible improvements to make it easier for community to use (like this PR).

.gitignore
```diff
@@ -0,0 +1,10 @@
+**/*.in
```
mckelvin: Usually `.in` files are use for generating some other files. I don't think we should ignore them.

.gitignore
```diff
@@ -0,0 +1,10 @@
...
+**/*.o
```
mckelvin: I think `*.o` is enough if you want to ignore all `.o` files in the repo. The same to the other lines in the file.

.gitignore
```diff
@@ -0,0 +1,10 @@
...
+**/*.lo
+**/*.la
+**/Makefile
```
mckelvin: There're several `Makefile`s in the repo, which is generated using `tmake`. I'd like to remove them and add this line if all existed `Makefile`s can be generated from the source (`make/gen_make/`). Otherwise I'd prefer to remove `Makefile` here.

src/tools/gen_code/gen_code.c
```diff
@@ -4369,18 +4371,18 @@ void genJavaCodePhase2( const TA_FuncInfo *funcInfo )
...
 #else
    /* The options are the quite same, but on linux it still outputs the #include lines,
 	didn't find anything better that to cut them with the sed ... a hack for now. */
    sprintf( buffer, TA_MCPP_EXE " -@compat -+ -z -P -I.." TA_FS_SLASH "src" TA_FS_SLASH "ta_common -I.." TA_FS_SLASH "src" TA_FS_SLASH "ta_abstract -I.." TA_FS_SLASH "include -D _JAVA .." TA_FS_SLASH "src" TA_FS_SLASH "ta_func" TA_FS_SLASH "ta_%s.c | sed '/^#include/d' >> .." TA_FS_SLASH "temp" TA_FS_SLASH "CoreJavaCode1.tmp ", funcInfo->name);
-   system( buffer );
+   ret = system( buffer );
 
    sprintf( buffer, TA_MCPP_EXE " -@compat -+ -z -P -I.." TA_FS_SLASH "src" TA_FS_SLASH "ta_common -I.." TA_FS_SLASH "src" TA_FS_SLASH "ta_abstract -I.." TA_FS_SLASH "include -D _JAVA .." TA_FS_SLASH "temp" TA_FS_SLASH "CoreJavaCode1.tmp | sed '/^#include/d' > .." TA_FS_SLASH "temp" TA_FS_SLASH "CoreJavaCode2.tmp " );
-   system( buffer );
+   ret = system( buffer );
```
mckelvin: Why adding the `ret`? I think it's never used.

src/tools/gen_code/Makefile.am
```diff
@@ -9,6 +8,6 @@ gen_code_CPPFLAGS = -I../../ta_common
 gen_code_LDFLAGS = -L../../ta_common -L../../ta_abstract -L../../ta_func
 gen_code_LDADD = -lta_common -lta_abstract_gc -lta_func -lm
 
-all-local:
-	$(MAKE) $(AM_MAKEFLAGS) gen_code
-	cp gen_code ../../../bin
+# all-local:
```
mckelvin: I'd prefer to leave it unchanged. Is it possible to not install the `gen_code` script in the `PKGBUILD`?

.gitignore
vbmithr: These files are autogenerated from `Makefile.am` files. They should not be commited in a repository.

.gitignore
vbmithr: Ok.

src/tools/gen_code/gen_code.c
vbmithr: I think it was triggering a warning (possibly error) in my machine. I remember thinking that it was indeed useless.

src/tools/gen_code/Makefile.am
vbmithr: This was triggering a recursive `make` that was screwing up the compilation when using parallel compilation.

.gitignore
mckelvin:
```
$ find . -name "*.in"
./ta-lib.dpkg.in
./ta-lib.pc.in
./ta-lib.spec.in
```

What about these files?

vbmithr: Fixed (and improved, now ignoring all generated files).

mckelvin:
Thanks @vbmithr !

PR #3, opened 2019-12-31
Add solution for Visual Studio 2019

hmG3:
Recently I used this library in my personal project so I'd like to contribute a solution file for VS 2019. I know that this is not an official repo, but, maybe, it would be useful for anyone, like me in my time.

Also, I did some basic source clean-ups using a code static analyzer tool.


mckelvin:
Hey Anatoliy,

Thanks for the PR. Right now there's no CI pipeline to make sure the code is working on every platform. In the very beginning I just wanna a modern way to clone the code and now I feel like everyone needs a modern way to use this software, like CI (test, lint), multiple platform build, ... Most of the commits LGTM except ba251392e0de81463bef892e5b73b9c9c3fde047 as I can't verify if this works for every platform. Probably we can do it after the CI pipeline is done. What do you think?

hmG3:
@mckelvin  It's a good improvement to set up a pipeline, especially for 15-year-old software. And there is no rush to complete this request. BTW To verify correctness, I tried to compile the source code on the latest Ubuntu but no luck. Neither _make_ nor _cmake_ worked because of errors in configuration files.

PR #16, opened 2024-01-05
Fix a warning

trufanov-nok:
Clang-tydy produces a lot of warnings but only this one seems to be a bug. 
retCode is always TA_INTERNAL_ERROR(131);


mario4tier:
Thanks for fixing this.

PR #28, opened 2024-01-22
Fix name typos of CdlStickSandwich

Frederisk:
Looks like `CdlStickSandwhich` is just a misspelling of `CdlStickSandwich`?


mario4tier:
Thanks for fixing this.

PR #33, opened 2024-05-01
Fix overflow in floating point multiplication

iglesias:


mario4tier:
Fixed on `main` ✅ — the single-precision (`TA_S_*`) functions computed their arithmetic in float before widening to the double output, so combining two inputs could overflow to infinity (e.g. `TA_S_MULT` of two large floats). They now widen each float input to double first, so `TA_S_<F>(float)` is bit-identical to `TA_<F>((double)float)`. Since TA-Lib's C is generated, the fix went into the code generator, so it covers the whole single-precision family, not just MULT.

See the [CHANGELOG](https://github.com/TA-Lib/ta-lib/blob/main/CHANGELOG.md) entry for #33. Scheduled for the **0.7.2** release (~second half of 2026). Thanks @iglesias for reporting this and for taking the initiative with the fix — casting to double before the multiply was exactly the right idea!

Issue #35, opened 2024-05-23
Add License to Repo


SeeRich:
The linked website in the project readme mentions BSD license. Could you add this to the repo to make it more easily identifiable.

PR #38, opened 2024-10-02
Add BSD 3 license file

major:
This will make packaging in Fedora much easier if the license file is included in the repository.

Fixes TA-Lib/ta-lib#35


mario4tier:
Thanks. 

I will modify it though to match the license in the source code (using my name instead of TA-Lib) because TA-Lib is not a legal entity (not sure if it matters).

PR #51, opened 2024-12-24
Allow for Debian 11 and Ubuntu 22.04 LTS support with lower version of CMake

vikasnkumar:
CMake 3.30 is too strict and the code builds just fine on Cmake 3.18 and 3.22 which are on Debian 11 and Ubuntu 22.04 LTS. 

The library has been in existence for  decades and forcing it to only work on most recent CMake's is a detriment to long term users and downstream applications that use this library.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.30)
+cmake_minimum_required(VERSION 3.18)
 
 project(ta-lib)
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,7 @@
 cmake_minimum_required(VERSION 3.18)
+if (CMAKE_VERSION VERSION_LESS "3.30" AND WIN32)
+    message(FATAL_ERROR "For Windows systems you need CMake Version 3.30")
+endif (CMAKE_VERSION VERSION_LESS "3.30" AND WIN32)
 
 project(ta-lib)
 


mario4tier:
I know. It is a dilemma and had to make a though call.

Why 3.30?
- it was the minimum needed to get the .msi generation (CPack) to work for windows.
- at least 3.22 was needed for another reason that I forgot (did use a tool verifying the minimum required base on what is done in the CMakeLists).

Why bending to windows?
I will be a bit judgmental here: window users typically struggle more for building the lib, so the .msi is a big step forward to them. On the opposite side, Linux devs often have more "technical agility".

Here are some alternatives for Linux:
  - autotools works
  - the release now includes multiple .deb variant... and .rpm is next.
  - Personally, I did work around this by building the latest CMake on my ubuntu machine (was not hard to do).
  - There is no "forcing": 0.4.0 is still usable until the distributions catch-up.

Conclusion
==========
Will revisit for solutions (e.g. have two CMakeLists.txt files?) only if many complains or it is blocking to the large ta-lib-python crowd.
 
I did not succeed to get CMakeLists specify different minimum version (base on platform)... and will welcome a PR that makes that possible.

vikasnkumar:
Hi @mario4tier 

I updated the PR just now to handle the Windows 3.30 requirement.

Thanks.

mario4tier:
The .msi is produced by the more recent Wix5 toolset which is supported only by the more recent version of CMake. Specifying the higher version is not without reason.

Just to be clear, I did not meant in my response to diminish the importance of your apps. It was a tough choice for me to choose who it will break for.

Thanks for the contribution, I will merge your change (may be tomorrow). I need first to change the github action to modify the CMakeLists.txt for when building for windows.

vikasnkumar:
thanks. I appreciate what you and your team does.  Let me know if you need help with anything related to supporting other languages with TA-Lib. I hope to keep maintaining it for another ten years in Perl at least.

PR #66, opened 2025-05-30
Update merge-ta-lib-temp.py

JackDonwel:


mario4tier:
Looks AI generated :smile_cat: 

Nothing wrong with that, except that the changes are too broad, and not clear what problem is being solved here.

I would prefer that you submit one small fix at the time.

PR #73, opened 2025-10-20
docs: adding notes about the Ruby and Zig wrappers

mrjbq7:
The Zig wrapper works pretty well, but doesn't yet have the abstract API.  Still, might be worth mentioning.

And also the Ruby wrapper.


mario4tier:
Changes are on main branch.

PR #75, opened 2025-10-29
add functions documents

nehemiah888:
Add functions/ directory in docs/ directory, and the functions are grouped. every function has calculation/formula, interpreter, example sections. And every function has a chinese version too. 
I tried to find such documents but failed, so I generated these documents by AI, and let differnet AI model to have a double check on the calculation/formula part to align to the codes.
I wish these documents could be merged to original project, so anybody could review/contribute these documents, and then make these documents better and better.


mario4tier:
@nehemiah888 

Documentation had to be handled differently because TA-Lib is intended to be distributed in various way, the website being only one of them.

So I am trying to develop a "Single source of truth" from which various format can be generated (in particular Rust/NPM).

The solution so far is to have a doc markdown alongside the "code" in ta_codegen/input.

Example for sma, the input file is: 
  https://github.com/TA-Lib/ta-lib/blob/main/ta_codegen/input/sma/sma.md

... this file is used by ta_codegen to maintain the website. The output is:
   https://ta-lib.org/functions/sma/

So effort now should be directed to maintain the markdown files under ta_codegen/input.

I am sorry that I went a different way than your approach. Your idea and initiative was good. 

If interested, there is still iterative work to be done on the input file. In particular, the formula section generally "look bad" and we have to find an AI friendly way to express these (Latex?)

mario4tier:
Thank you for putting this together, @nehemiah888! 🙏

TA-Lib's documentation is generated: each page under `docs/functions/` is produced by our code generator from a single source-of-truth file (`ta_codegen/input/<name>/<name>.md`). Hand-placed pages in grouped subfolders would be overwritten on the next regeneration, and we don't yet have a mechanism for the Chinese variants.

Rather than lose your contribution, I merged this into `dev` for attribution and cherry-picked the load-bearing facts into the canonical source: author and originating-publication references now appear on the DEMA, TEMA, T3, MFI, ULTOSC, KAMA and TRIX pages (each citation verified against primary sources before inclusion).

See the [CHANGELOG](https://github.com/TA-Lib/ta-lib/blob/main/CHANGELOG.md) entry for #75. Scheduled for the **0.7.2** release (~second half of 2026). Thanks again @nehemiah888 !

PR #86, opened 2026-06-05
Add basic Conan support

CaptainTrunky:


mrjbq7:
What is your motivation for this? When would you use this instead of the binary releases that the project ships. 

How would it be tested?

How popular is Conan, I've never used it. Do they have a formula registry that could keep this file instead of here in our project.

CaptainTrunky:
Personally, I'm using a different platform, for which I have to build it from src. While make/cmake are well-known and established, they miss some nice modern features, in my opinion.

I find conan and it's alternatives (e.g. mesos), to be way more convenient - dependency management, versioning, supporting different platforms and even isolated self-contained & reproducible builds are way more comfortable using it.

Sure, you can achieve the same yourself using cmake, but you would write tons of boilerplate and/or end up with your own handmade conan/mesos.

I'm pretty sure it's the next most popular option after plain cmake. It's a high-level generator on top of cmake and 100% compatible with it. It has its own registry (conan.io), but you have to provide conan config in your project.

As for testing, I think adding CI step that uses conan for building and running tests is a good starting point.

Hope I sound rather convincing :)

mrjbq7:
Thank you. Which platforms does this not build properly on? I could be confused but I don't see how this Conan script does anything different than running cmake?

CaptainTrunky:
cmake/make builds just fine; it's not the issue itself (Fedora43, clang-21, gcc 15). Major benefits for conan integration are not building the project itself, but ease to integrate your software - instead of cmake's fetchcontent or bash's scripts that do "git clone blah" it's a trivial python script:

```
        self.requires("benchmark/1.8.4")
        self.requires("gtest/1.14.0")
        self.requires("fmt/12.0.0")
        self.requires("liburing/2.6")
```

It will work on the most imaginable platforms in the same way. Plus, conan checks for consistency of your dependencies. Basically, conan plays the same role as cargo for rust or pip for python.

mario4tier:
Thanks for your contribution. I am open to include this. Seems to be low maintenance and non-conflicting with the rest.

Will try to get an AI agent review and follow up... as a learning experience for me. Bare with us if it gets weird :smile:

mario4tier:
The agent was a bit buggy (could not push to the dev branch and could not publish the summary here). So I am cut&pasting the summary here.
-=-=-=
Thanks @CaptainTrunky for the contribution! 🎉

This has been merged into dev via PR #87.

Two changes were made on top of the original PR during integration:

- **Rebased onto dev** — the original PR targeted main; the branch was rebased to merge cleanly into dev.
- **Version sync integration** — Add auto-maintenance of the version string. See changes in scripts/utilities/versions.py for the workflow so the version stays in sync with all other files when scripts/sync.py is run (manually or by CI/CD)

mario4tier:
@CaptainTrunky Feel free to submit/maintain TA-Lib at conan.io 

I did not look into it, but if it is like brew (it is all automated and building directly from our repos) then this would be great!

Thanks.

CaptainTrunky:
Thanks, @mario4tier! Yes, I'll publish the package there soonish.

PR #96, opened 2026-07-05
Adopt FMA: faithful-algorithm-within-tolerance numerical contract

mario4tier:
**Decision: TA-Lib adopts fused multiply-add (FMA), under a clearer contract** — each function faithfully implements its algorithm **within a documented tolerance** (every result within ~1e-9 of before, in practice the last bit), rather than promising bit-identical output forever.

**Why:** ~25% faster recursive indicators on every CPU since 2013, and — for the first time — identical results across x86 and ARM. Chasing bit-for-bit floating-point across evolving hardware isn't sustainable.

**What changes for users:** effectively nothing visible. Measured fused-vs-current across 27M outputs on real and adversarial data: continuous values shift by at most ~1e-10 (usually the last bit); discrete signals (candlestick patterns, SAR, trend mode) didn't change once, though patterns fired 23k+ times. Evidence: `docs/fma-readiness-audit.md`.

**For user keeping stored reference calculated with older TA-Lib version:** expect one tiny bounded shift, then stability again.

Landing in 0.8.1 after the current optimization pass.


mario4tier:
I will leave that here in case someone care, but my mind is already set to modernize the code unless someone "insist" that TA-Lib should remain 100% bitwise-matching like the last 25 years...

To be clear, we are talking here of very small output differences (likely insignificant to most), in exchange with faster/modern code.

BTW, this debate is not specific to TA-Lib, there are similar recuring discussion in the Rust community about what "default" means and if it should adapt to more modern processor.

Forever matching floating point output is... complicated.

mrjbq7:
I'm in favor! Let's move into the future!

mario4tier:
@mrjbq7 Great timing! I was coming back here to see how I should bubble this up to the largest python community.

Thanks for concurring. 

Will probably move to FMA in ~~late 2027~~ release 0.8.1. I want to do first a bunch of fixes and algo-level optimization while we still can do bit-perfect comparison against the 0.6.4 reference.

mario4tier:
### Implementation complete — merging to `dev`

This PR opened as the RFC/decision and now carries the **full implementation** of the FMA numerical contract (commit `74c18731`). Summary of what was done:

**Contract & generator**
- Explicit `fma()` at the ~26 recursive fused indicators, driven by one shared fusion detector so **C / Rust / Java fuse the identical sites and stay bit-identical** (correctly-rounded IEEE-754 fma); .NET inherits via P/Invoke. Faithful within ~1e-9 relative vs 0.6.4, not bit-for-bit.
- `-ffp-contract=off` (CMake `NOT MSVC` + autotools accept-check) makes the explicit `fma()` the **only** fusion → deterministic across compilers/platforms/languages.

**Runtime CPU dispatch (portable, no `-mfma`)**
- C batch variants marked `target_clones("default","fma")`, gated on `__x86_64__ && __GLIBC__ && __GNUC__ && !__clang__`. On **manylinux** a baseline build dispatches to a hardware-`vfmadd` clone at load via an ifunc — portable, **no SIGILL** on pre-2013 CPUs.
- `glibc` is the only libc that implements ifunc, so **musl (musllinux) / Windows / macOS-Intel correctly fall through to a plain software `fma()`** (correct, just unaccelerated). Streaming is left unmarked — benchmarked, the per-bar dispatch is a wash.

**Verification / guards**
- `fuzz-064` one-time transition gate (1e-9 relative; re-freeze to hash-exact after tagging).
- New `musl-build` nightly job + `scripts/verify-musl-fma.sh` — regression guard proving the library builds on musl with dispatch correctly excluded (`target_clones` requires GNU ifunc; musl supports it in no version).
- Green at merge: glibc build (104 ifunc) + reference suite + `fuzz-064` (0 fail, max 4.13e-11), regen oracle clean, 551 generator tests, musl build proven, and the **manylinux wheel path smoke-tested** (ifunc-dispatched `TA_MACD` resolves + computes correctly under a Python `dlopen`).

**Known, accepted regressions** (all correct, SIGILL-free, within the 1e-9 contract): software `fma` on the ~15–20 recursive fused indicators for **Windows-x86_64 / macOS-Intel / musllinux-x86_64**; **manylinux-x86_64 + every ARM target** get the hardware-FMA win / baseline.

_Follow-ups tracked separately (non-blocking): a future portable manual-`cpuid` dispatch if hardware FMA on Windows/musl is ever wanted; ta-lib-python's own CI wheel `import` at the `TALIB_C_VER` bump._

PR #128, opened 2026-07-18
perf(minmax): unroll batch rescan loops

dexhunter:
## What changed

This replaces the earlier monotonic fast-path proposal in this PR, which you
correctly rejected as a synthetic case. The rework targets the common
random-walk domain instead and does **not** change the algorithm.

`TA_MIN`/`TA_MAX` already cache the current window extreme and rescan the
window only when that extreme expires. On random-walk input that rescan runs
regularly, and it is where the batch functions spend a large share of their
time. This change emits an explicit unroll hint (`#pragma clang loop
unroll_count(4)` / `#pragma GCC unroll 4`, guarded to Clang and GCC >= 8) on
that existing rescan loop, cutting its loop-control overhead:

- no comparison is added anywhere, in particular not in the per-sample tight
  loop;
- the algorithm, outputs, and streaming implementation are untouched — the
  generated streaming source is byte-identical;
- MSVC and older GCC see no pragma and compile exactly the code they compile
  today.

The hint is emitted by the C backend of `ta_codegen` for MIN/MAX batch bodies
only, and generation fails closed unless each MIN/MAX batch body contains
exactly one rescan-unroll site, so the directive cannot silently land on the
wrong loop or vanish. `src/ta_func/ta_MIN.c`, `ta_MAX.c`, and the sources
digest are regenerated with the official generator; generation is idempotent.

## Measurements (instruction counts, not wall-clock)

Setup: x86_64, GCC 13.3.0 `-O2` Release, Callgrind (valgrind 3.22.0)
instruction counts (Ir), which are deterministic. Baseline is pristine `dev`
(`cae231d8`). 168 paired cases: MIN and MAX, periods 14/30/50, n=20,000,
seven random-walk regimes (plain, low-vol, high-vol, drift up/down,
mean-reverting, tick-quantized), four seeds — two of which were held out and
never used while tuning.

- Total: 134,392,343 -> 119,490,628 instructions (**-11.09%**)
- Regressed cases: 0 of 168
- Per-regime range: -9.38% (mean-reverting) to -12.22% (drift up)

Wall-clock samples on this host were too noisy to be meaningful (baseline
spread up to 43% across repetitions), so we make no wall-clock claim — this
is an instruction-count result on x86_64 GCC. Happy to run any benchmark
setup you prefer.

Correctness evidence: 3,150 double-API semantic pairs plus 2,016 float-API pairs were bit-exact vs pristine `dev` (all
seven random-walk regimes plus NaN, infinities, ties, plateaus, signed zero,
in-place aliasing, sub-ranges, guarded/unguarded, periods
1..257); full `ta_regtest` under GCC 13.3 and Clang 18.1 (`-Werror`);
ASan+UBSan `ta_regtest --function=MIN,MAX`; generator test suite and clippy
clean; two idempotent generation passes.

## On an alternative algorithm

We looked, and the honest answer is that we did not find one that beats the
current cached-extreme algorithm under your (correct) constraint that the
tight loop must not gain comparisons:

- **Branchless rescan** (ternary compare/select so the compiler can emit
  cmov): GCC already generates equivalent code — measured instruction counts
  were identical to baseline.
- **Latest-index tie-keeping** (keep the most recent extreme on ties so it
  survives longer and rescans trigger less often on tick-quantized data):
  changes which index the cached extreme tracks, and failed our bit-exact
  matrix on tie/zero patterns, so it silently alters behaviour.
- **Pairwise/tournament rescan** (halve the compare-update dependency chain):
  failed bit-exactness on NaN patterns due to reordered comparisons.
- Dequeue and Van Herk/Gil-Werman you had already ruled out; we agree — their
  per-sample bookkeeping is exactly the added tight-loop work that hurts.

So this PR deliberately keeps your algorithm and only tightens the code the
compiler emits for the rescan that already exists.

### DIFF
--- a/include/ta_common.h
+++ b/include/ta_common.h
@@ -151,7 +151,7 @@ TA_LIB_API TA_RetCode TA_Shutdown( void );
  * This value is updated whenever a make, cmake or any source files
  * modification should trig a repackaging of TA-Lib.
  */
-#define TA_LIB_SOURCES_DIGEST 2f2afcea23151d63b3957a0636e8df83
+#define TA_LIB_SOURCES_DIGEST 25d4faee56d7415dac044d8591341c80
 
 #ifdef __cplusplus
 }
--- a/src/ta_func/ta_MAX.c
+++ b/src/ta_func/ta_MAX.c
@@ -135,6 +135,11 @@ TA_LIB_API TA_RetCode TA_MAX( int    startIdx,
          highestIdx = trailingIdx;
          highest = inReal[highestIdx];
          i = highestIdx;
+#if defined(__clang__)
+#pragma clang loop unroll_count(4)
+#elif defined(__GNUC__) && __GNUC__ >= 8
+#pragma GCC unroll 4
+#endif
          while( ++i <= today )
          {
             tmp = inReal[i];
@@ -202,6 +207,11 @@ TA_LIB_API TA_RetCode TA_MAX_Unguarded( int    startIdx,
          highestIdx = trailingIdx;
          highest = inReal[highestIdx];
          i = highestIdx;
+#if defined(__clang__)
+#pragma clang loop unroll_count(4)
+#elif defined(__GNUC__) && __GNUC__ >= 8
+#pragma GCC unroll 4
+#endif
          while( ++i <= today )
          {
             tmp = inReal[i];
@@ -280,6 +290,11 @@ TA_RetCode TA_S_MAX( int    startIdx,
          highestIdx = trailingIdx;
          highest = (double)inReal[highestIdx];
          i = highestIdx;
+#if defined(__clang__)
+#pragma clang loop unroll_count(4)
+#elif defined(__GNUC__) && __GNUC__ >= 8
+#pragma GCC unroll 4
+#endif
          while( ++i <= today )
          {
             tmp = (double)inReal[i];
@@ -344,6 +359,11 @@ TA_RetCode TA_S_MAX_Unguarded( int    startIdx,
          highestIdx = trailingIdx;
          highest = (double)inReal[highestIdx];
          i = highestIdx;
+#if defined(__clang__)
+#pragma clang loop unroll_count(4)
+#elif defined(__GNUC__) && __GNUC__ >= 8
+#pragma GCC unroll 4
+#endif
          while( ++i <= today )
          {
             tmp = (double)inReal[i];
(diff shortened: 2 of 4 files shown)


mario4tier:
(The intent of this reply is to test AI response)
@dexhunter @aiden-weco

Thanks for the PR. The proposed optimization is for a synthetic test that would not happen normally in TA-Lib context.

Therefore, I prefer to not add this complexity for an improbable scenario.

I would prefer you attempt to optimize for the common cases in our domain (e.g. random-walk input).

For min/max, some alternative were already explored (dequeue and Van-Herk) but were found to not be more performant. Most of the time, adding a comparison or two in the tight loop is just enough to affect negatively the performance.

Can you think of an alternative algo for min/max?

dexhunter:
Thanks for the direction. I replaced the monotonic fast path with a common-case random-walk optimization. It keeps the existing cached-extreme algorithm and adds no comparisons to the tight loop; the generator emits an unroll hint only for the existing batch rescan. Across 168 random-walk cases, deterministic Callgrind instructions fell from 134,392,343 to 119,490,628 (-11.09%) with 0 regressions. I also reran the full GCC and Clang regression suites, ASan plus UBSan, generator tests and clippy, 3,150 double-API semantic pairs, and 2,016 float-API pairs bit-exact against current dev. Wall-clock samples were too noisy, so I am not claiming a wall-clock speedup. The updated PR description includes the alternatives explored and why they were rejected.

mario4tier:
Merged — thanks @dexhunter and @aiden-weco.

Your commit landed as 65e49d72 (merge 0365a69a), and the follow-up work is d661b9ae on `dev`.

What we kept is the core insight: the expired-extreme rescan is the hot part, and it can be sped up without touching the algorithm or the outputs. On top of your commit, d661b9ae expresses the hint as `TA_UNROLL(n)` written directly in `ta_codegen/input/<f>/<f>.c` (so the source of truth stays where it belongs), extends it past MIN/MAX to the 18 sites across 11 functions that carry the same rescan, and enables it on the toolchains where a code-layout-swept benchmark showed it actually pays: roughly -8% on GCC x86-64 and -27% on Apple clang arm64, warm. Outputs are bit-identical, and the Rust/Java/.NET backends are byte-identical.

Please keep digging. We particularly value **algorithmic** changes worth **>15%** that keep outputs within **1e-12** of the current results (some tolerance for floating-point re-ordering is fine). Two things that make a proposal easy to accept: measure on random-walk input rather than synthetic best cases, and sweep code layout — we have seen identical machine code move -13% to +2% on placement alone.

PR #143, opened 2026-07-25
perf(mavp): bound each ma() pass at its period's last use

dexhunter:
## What this changes

`TA_MAVP` normalizes every requested period into `localPeriodArray`, then walks that array:
for each output whose period is still flagged, it calls `ma()` over the **whole remaining
range** and scans the rest of `localPeriodArray` to find and copy every other output sharing
that period. Two independent inefficiencies live there: the selection scan is `O(U*n)` for `U`
distinct periods over `n` outputs, and every `ma()` pass runs all the way to `endIdx` even when
its period's last user is much earlier.

This change makes the existing normalization pass group the outputs, and then uses that
grouping to bound each `ma()` pass:

* `localPeriodArray` now holds, for each output, the index of the **next** output using the
  same period (`-1` ends the chain). It is the same allocation as before, reused.
* A new `periodIndex` array holds the **first** and **last** occurrence of each representable
  period (`periodIndex[period]` and `periodIndex[optInMaxPeriod+1+period]`).
* Selection becomes one `O(n)` build plus one `O(n)` total chain traversal, replacing the
  per-period rescan.
* Each `ma()` call now ends at the last output that uses its period, rather than at `endIdx`.

Both parts contribute measurable time, and the second is described on its own below because it
is the one change that goes beyond restructuring the period selection.

### Each `ma()` call now ends at the last output that uses its period, not at `endIdx`

Every moving average in TA-Lib warms up before `startIdx` and then fills forward, so an output
at index *k* depends only on input up to bar `startIdx + k`; ending the call earlier cannot
change any output at or before that point. The inner call's own `outBegIdx`/`outNBElement` are
written but never read by `TA_MAVP`, and the values `TA_MAVP` reports to *its* caller are
unchanged.

This is worth roughly a fifth of the improvement below: without it, the same patch measures
≈0.78 rather than ≈0.73 against the same control. It pays off on period series whose values
stop being used before the end of the range (as wandering or regime-driven period inputs
produce); on layouts where every period is used until the end it does strictly less or equal
work and changes nothing.

Because the `ma()` range is now bounded, `localOutputArray` is only partially written (indices
`0..lastOccurrence`). Every index read is on that period's chain and is therefore
`<= lastOccurrence`, so there is no uninitialized read; this was checked under ASan, UBSan and
LSan across the full comparison matrix.

## Result agreement

One `ma()` call per distinct clamped period is still made and the arithmetic inside those
passes is untouched, so no summation, division or accumulation order changes anywhere. The
calls are now issued in ascending period order rather than first-occurrence order, and each
output is written from the same pass that produced it before, so the outputs are
**bit-identical** rather than merely within tolerance.

Direct comparison against the unpatched library over 20,000-bar random walks — max periods
30/100/500, seven price regimes, and batch, in-place (`outReal == inReal`) and subrange calls —
is **bitwise identical: max absolute delta 0.0**, comfortably inside the 1e-12 agreement bar.
This was verified twice with two independently written harnesses: 10,804,965 values compared
over nine MA types, and 8,904,640 values compared over **all ten MA types reachable through
`TA_MAVP`** — `SMA, EMA, WMA, DEMA, TEMA, TRIMA, KAMA, MAMA, T3, HMA`. Zero differing values
in both, including every return code, `outBegIdx` and `outNBElement`. The second harness also
repeated its full matrix under a non-default `TA_SetUnstablePeriod(TA_FUNC_UNST_ALL, 7)`, with
zero differences, which is the sharpest check on the bounded `ma()` range described above.

## Measured improvement

Paired p50 wall-clock ratio against the unpatched build, measured under the protocol requested
in #128 — random-walk inputs, swept over eight code layouts, fresh per-sample seeds, 84
`TA_MAVP` SMA calls over 20,000-bar series across seven price regimes, four series seeds and
max periods 30/100/500:

| Variant | Paired p50 ratio (lower is better) | vs control |
| --- | --- | --- |
| Untouched control | ≈1.00 | — |
| Period grouping only | ≈0.782 | ≈22% |
| Grouping + bounded `ma()` range (this PR) | ≈0.730 | ≈27% |

The shipped patch measures a mean of **0.7304** across five repeats (0.7292–0.7313), against a
control of ≈1.00 measured in the same window on the same host. The three groups are cleanly
separated: none of their ranges overlap, and each group's spread is an order of magnitude
smaller than the gaps between them.

> **The timing figures in this section are being re-verified.** They were measured on a
> shared host where a concurrent job could perturb wall-clock results, so I do not want them
> standing here as settled. They are being re-measured under exclusive single-tenant
> measurement and this section will be updated with the confirmed numbers — including if they
> differ from what is written above. The correctness results are unaffected: the bit-identical
> comparison, the test suites and the codegen verification are exact checks, not timings.

## Verification

* `scripts/build.py test` — full C reference suite passes, including the 200-pair in-place
  alias gate and the 168-function four-variant parity gate (bit-identical).
* `ta_regtest --function=Moving` — the `All Moving Averages` group (which is what covers
  `TA_MAVP`) passes.
* `ta_regtest --codegen-only --language=c,rust` — codegen verification passes for both,
  161/161 functions each, against the frozen `reference-pre-cutover` oracle.
* `scripts/build.py generate` is idempotent: a second run produces a byte-identical tree.
* Regenerated surfaces committed alongside the template: the C library, the Rust crate and
  the Java `Core_MAVP` fragment plus `Core.java`. The Rust crate compiles (`cargo build`).
* Java and .NET server verification was **not** run — this machine has no JDK or .NET SDK, so
  those two server builds fail here on the unpatched tree as well. The Java source is
  regenerated and committed; it has not been compiled.

## Notes for review

* The added allocation is `2*(optInMaxPeriod+1)` ints. It is sized by the configured period
  range rather than by the data, so for a very large `optInMaxPeriod` with a very short output
  range it is larger than the previous scratch. It is null-checked and returns `TA_ALLOC_ERR`
  with full cleanup of the earlier allocations.
* The `TA_MAVP_Unguarded` / `TA_S_MAVP_Unguarded` entry points skip the guarded API's
  `optInMinPeriod >= 1` range check by design. A caller passing `optInMinPeriod <= 0` through
  those would now reach a negative `periodIndex` index, where the baseline instead passed the
  same nonsensical period into `TA_MA_Unguarded`. Both are undefined on that input and the
  guarded public API rejects it; flagging it since it is the one new memory-unsafe shape.
* `ta_codegen/input/mavp/mavp.c` is the only hand-edited file; everything else in the diff is
  generator output.

Follow-on to #128, which invited algorithmic changes worth more than 15% with output agreement
within 1e-12, random-walk inputs and code-layout sweeps.

Supplementary autoresearch: the optimization direction was explored with an automated search over the same sealed benchmark — one authoritative run, 11 integrity-valid scalar-emitting steps, best node a 27.0% improvement against the paired-ratio baseline: [public trajectory](https://dashboard.weco.ai/share/BL2XnwhXLK6H-nUDu7ASKQjtbmK9uheb). The submitted patch is a manual minimal rebuild of that best node's idea, independently re-verified as above.

### DIFF
--- a/src/ta_func/ta_MAVP.c
+++ b/src/ta_func/ta_MAVP.c
@@ -84,12 +84,14 @@ TA_LIB_API TA_RetCode TA_MAVP( int    startIdx,
                                double        outReal[] )
 {
    int i;
-   int j;
    int lookbackTotal;
    int outputSize;
    int tempInt;
    int curPeriod;
+   int lastOccurrence;
+   int chainIdx;
    int *localPeriodArray;
+   int *periodIndex;
    double *localOutputArray;
    double *localFinalArray;
    int finalIsAllocated;
@@ -167,6 +169,19 @@ TA_LIB_API TA_RetCode TA_MAVP( int    startIdx,
    /* Allocate intermediate local buffer. */
    localOutputArray = malloc(outputSize * sizeof(double));
    localPeriodArray = malloc(outputSize * sizeof(int));
+   /* Occurrence index for the period grouping below, one pair of entries per
+    * representable period: periodIndex[period] is the first output using that
+    * period and periodIndex[optInMaxPeriod+1+period] the last one.
+    */
+   periodIndex = malloc(2 * (optInMaxPeriod + 1) * sizeof(int));
+   if( periodIndex == NULL )
+   {
+      free(localOutputArray);
+      free(localPeriodArray);
+      *outBegIdx= 0;
+      *outNBElement= 0;
+      return TA_ALLOC_ERR;
+   }
    /* In-place defence (issue #130): each ma() pass below re-reads inReal over
     * the full range, so with outReal==inReal the results are staged in a
     * scratch buffer and copied once at the end. A regular call writes
@@ -181,9 +196,17 @@ TA_LIB_API TA_RetCode TA_MAVP( int    startIdx,
    {
       localFinalArray = outReal;
    }
-   /* Copy caller array of period into local buffer.
-    * At the same time, truncate to min/max.
+   /* Read the caller array of period, truncate to min/max, and group the
+    * outputs by that truncated period in the same pass. localPeriodArray now
+    * holds, for each output, the next output using the same period (-1 ends
+    * the chain), so the outputs of one period form a linked list rooted at
+    * periodIndex[period]. This replaces the flag-and-rescan below, which
+    * walked the rest of the range once per distinct period.
     */
+   for( i = 0; i <= optInMaxPeriod; i += 1 )
+   {
+      periodIndex[i] = 0 - 1;
+   }
    for( i = 0; i < outputSize; i += 1 )
    {
       tempInt = (int)inPeriods[startIdx + i];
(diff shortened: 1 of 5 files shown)


mario4tier:
Merged — thank you @dexhunter! This PR is the basis of what shipped on dev (`e9848f1e`, with the merge commit crediting this PR's commit directly).

The core idea — ending each per-period `ma()` pass at that period's last use, bit-exact because every dispatched MA fills forward — held up under everything we threw at it (bitwise A/B across all MA types, unstable periods, in-place and subrange calls, sanitizers, the full cross-language sweeps) and is unchanged in the landed version.

We did rework the selection machinery around it before landing: the linked-list grouping measured slower than the old rescan on some period layouts — a constant/near-constant period series (one distinct period) came in ~1.7x slower, and a large `optInMaxPeriod` with a small used range ~3.2x, from the O(maxPeriod) table setup plus the pointer-chasing chain walk. The committed version keeps your bounded `ma()` range and swaps the grouping for a counting sort sized by the used period range, with a single-distinct-period fast path and a contiguous-run block copy. Net effect: faster than the previous code in every layout we measured (roughly 20–60% less time depending on the period series, ~30% for wandering periods in the default range), with no regression case left. A dedicated MAVP regression group now pins all of these paths bitwise across the C, Rust, Java and .NET backends.

Also appreciated: the measurement protocol write-up and the candor about re-verifying the timing numbers — it made the review straightforward. The CHANGELOG credits you for this and for #128.

More PRs like these are very welcome!

Issue #148, opened 2026-07-29
Rust backend emits no range check for `real` optional parameters — 19 documented bounds across 11 functions unenforced (T3 optInVFactor, MAMA limits, SAR acceleration, CDL optInPenetration)


kevinlincg:
The Rust backend validates integer optional parameters but skips `real` ones entirely.
C rejects every documented out-of-range value; Rust returns `RetCode::Success` and a
number. Refs at `144c635b`.

### Evidence — documented vs actual vs C

T3 is the clearest case. The bound is documented twice — `src/ta_func/t3.rs:83-84` and
`:120-121`, *"(default 0.7, range 0..=1)"* — and `t3.rs:131` promises
`RetCode::BadParam` "when an optional parameter is outside its documented range". No
entry point checks it: `t3()` (`:169`) validates only `optInTimePeriod` at `:183-186`;
same for `t3_lookback` (`:91-92`), `t3_open_internal` (`:574-577`), `t3_open_and_fill`
(`:815-818`). (`t3_unguarded`, `:362`, is validation-free by design — not the issue.)
C checks at all three tiers: `ta_T3.c:76-79`, `:122-125`, `:448-451`.

Not T3-specific. `ta_codegen/input/*/*.yaml` declares `type: real` with a bounded
`range:` for **19 parameters across 11 functions**, none checked in Rust:
`optInPenetration` ≥0 in the seven `CDL*` functions, `optInFastLimit`/`optInSlowLimit`
0.01..=0.99 in MAMA, `optInAcceleration`/`optInMaximum` ≥0 in SAR, seven acceleration
params in SAREXT, `optInVFactor` in T3. Five further real params (BBANDS
`optInNbDevUp`/`Dn`, STDDEV/VAR `optInNbDev`, SAREXT `optInStartValue`) are
`TA_REAL_MIN..TA_REAL_MAX`, which C skips deliberately — unaffected.

### Repro

python TA-Lib 0.7.1 (C) vs generated Rust 0.8.1, same 252-point sine series:

| call | C via python | Rust |
|---|---|---|
| `T3(x, 5, 1.05)` / `T3(x, 5, -0.5)` | `TA_BAD_PARAM` | `Success`, last = 99.337 / 93.188 |
| `MAMA(x, 1.5, 0.05)` | `TA_BAD_PARAM` | `Success`, last = 97.250 |
[...]

PR #149, opened 2026-07-29
fix(ta_codegen): #148 — range-check real optional params in the Rust backend

kevinlincg:
Closes the Rust half of the gap reported in #148, and with it item 2 of
`ta_codegen/generator/CLAUDE.md`'s own **Known Code Quality Issues** list —
*"`gen_opt_param_validation` silently skips Real/Enum optional params"*.

## What this changes

`backends/rust_lang.rs::gen_opt_param_validation_with` wrapped its entire body in
`if opt.param_type == ParamType::Integer` (`:1021` on `dev`), so every `real` optional
parameter reached the generated Rust with no validation at all — no range check, and no
default-sentinel substitution either. The C backend has had the arm the whole time
(`backends/c.rs:266-282`), as has Java (`backends/java.rs:552-570`).

The `if` becomes a `match` and gains the `ParamType::Real` arm, emitting the same shape
the other two backends emit:

```rust
if optInVFactor == -4e37 {
    optInVFactor = 7e-1;
} else if (optInVFactor < 0e0) || (optInVFactor > 1e0) {
    return RetCode::BadParam;
}
```

against C's

```c
if( optInVFactor == -4e37 )
   optInVFactor = 0.7;
else if( optInVFactor < 0e0 || optInVFactor > 1e0 )
   return TA_BAD_PARAM;
```

The "skip unbounded ranges" magnitude test (`c.rs:275`) comes across unchanged, so the
five parameters declared `[TA_REAL_MIN, TA_REAL_MAX]` get the sentinel substitution and
no range check — exactly as in C.

Bounds are emitted in exponent form (`0e0`, `9.9e-1`) rather than plain Display, because
`0` and `1` would be integer literals on the right-hand side of an `f64` comparison and
would not compile.

## Scope, measured from the regenerated output

**24** real optional parameters now honour the `-4e37` sentinel. **19** of them, across
**11** functions, additionally reject out-of-range values:

| Function(s) | Parameter(s) | Declared range |
| --- | --- | --- |
| `CDLABANDONEDBABY`, `CDLDARKCLOUDCOVER`, `CDLEVENINGDOJISTAR`, `CDLEVENINGSTAR`, `CDLMATHOLD`, `CDLMORNINGDOJISTAR`, `CDLMORNINGSTAR` | `optInPenetration` (7) | `[0, TA_REAL_MAX]` |
| `MAMA` | `optInFastLimit`, `optInSlowLimit` | `[0.01, 0.99]` |
| `SAR` | `optInAcceleration`, `optInMaximum` | `[0, TA_REAL_MAX]` |
| `SAREXT` | `optInOffsetOnReverse`, `optInAccelerationInitLong`, `optInAccelerationLong`, `optInAccelerationMaxLong`, `optInAccelerationInitShort`, `optInAccelerationShort`, `optInAccelerationMaxShort` (7) | `[0, TA_REAL_MAX]` |
| `T3` | `optInVFactor` | `[0, 1]` |

The remaining five — `BBANDS.optInNbDevUp`/`optInNbDevDn`, `STDDEV.optInNbDev`,
`VAR.optInNbDev`, `SAREXT.optInStartValue` — are `[TA_REAL_MIN, TA_REAL_MAX]` and get
sentinel substitution only.

Every validated entry point is covered: `*_lookback` (which fails with `usize::MAX`), the
batch call, and both streaming entry points. The `_unguarded` variants stay validation-free
by design, exactly as they already were for integer parameters.

## Behavior change

**This rejects input that previously succeeded, and it is worth calling out explicitly for
the `ta-lib` crate's semver.** A real optional value outside a documented bound now returns
`RetCode::BadParam` (or `usize::MAX` from a lookback) instead of `RetCode::Success` and a
number. The rejection itself is not new to the library — the C implementation has always
performed it — but it is new to the Rust crate.

Measured on the built artifacts, same 60-point sine series, before and after this commit,
plus the C library from this tree for reference (`TA_SUCCESS = 0`, `TA_BAD_PARAM = 2`):

| Call | Rust before | Rust after | C |
| --- | --- | --- | --- |
| `t3(x, 5, 1.05)` | `Success`, 36 values | `BadParam` | `2` |
| `t3_lookback(5, 1.05)` | `24` | `usize::MAX` | `-1` |
| `sar(h, l, -1.0, 0.2)` | `Success`, max\|out\| = **6.29e66** | `BadParam` | `2` |
| `mama(x, 5.0, 0.05)` | `Success`, 28 values | `BadParam` | `2` |
| `cdldarkcloudcover(…, -3.0)` | `Success`, 49 values | `BadParam` | `2` |
| `t3(x, 5, -4e37)` | `Success`, first = **1.005e112** | `Success`, first = 108.38931656904083 | first = 108.38931656904 |

The last row is the second half of the change, and it goes beyond the direction suggested
in the issue. The issue argued Rust needed only the range check, since the generated docs
said only integer parameters accept a default sentinel. That reading was circular: the docs
said so *because* reals had no handling at all. Adding the range check alone would have made
`-4e37` — the value every other binding accepts as "use the default" — start returning
`BadParam`, replacing one cross-language divergence with a new one. With the sentinel branch,
`t3(x, 5, -4e37)` now produces exactly what `t3(x, 5, 0.7)` produces, matching C.

## Docs corrected in the same pass

Two statements in the tree were false, both about this code path:

1. The per-function rustdoc sentence *"Integer parameters accept `i32::MIN` to select their
   default value"* was emitted unconditionally for any function with optional parameters.
   SAR and MAMA have **no** integer optional parameter, so both documented an API they do not
   have. It is now gated on the kinds a function actually takes and names the real sentinel
   where one applies (`rust_doc.rs::default_sentinel_sentence`, shared by the batch and
   lookback doc emitters). The crate-level doc in `main.rs` gets the same treatment.

2. `docs/param-boundary-audit.md` closed with *"the library does not range-check real optional
   parameters"*. That was never true of C (`src/ta_func/ta_T3.c:76-79`;
   `backends/c.rs:266-282`). Rewritten to say what the library does and what the boundary
   sweep does not yet assert.

Note that the generated `# Errors` section already promised `RetCode::BadParam` *"when an
optional parameter is outside its documented range"*, and `t3.rs` already documented
`optInVFactor` as *"range 0..=1"*. The docs were describing the intended contract; only the
emitter was missing.

## Why `Enum` is not covered

`CLAUDE.md`'s note says "Real/Enum". This PR does Real only, deliberately.

All **13** enum optional parameters in `ta_codegen/input/*/*.yaml` (`optInMAType` and friends)
declare **no `range:` key**. C's `ParamType::Integer | ParamType::Enum(_)` arm therefore emits
no range check for any of them either — `ta_MA.c:73-74` is default substitution and nothing
else. An enum arm in the Rust backend would emit dead code.

What *is* genuinely missing for enums is the `i32::MIN` default substitution: C maps
`0x80000000` to the declared default, Rust does not, so `Core::ma(.., i32::MIN)` falls through
the dispatch `match` to the `_` arm instead of selecting SMA. That is a different behavior
change (sentinel semantics, not bounds) on a different set of functions, so it does not belong
in a PR titled "range-check real optional params". The known-issues note in
`ta_codegen/generator/CLAUDE.md` is narrowed to exactly that residue rather than deleted. Happy
to send it as a follow-up if you want it.

## Tests — the gate for both halves of the parameter contract (`test_codegen.c`)

You were right that the verification table below passed equally well without the
generator change. It does not any more. `ta_regtest --xlang-hash` now goes from
`PASS / 0 mismatches` to `FAIL / 13 functions` if either half of the arm is removed,
and each half is proven independently.

Everything is inside `ta_regtest`; the existing `fuzz_build_vectors` sweep is extended,
no new harness. Two candidate classes, both driven off `ta_abstract` metadata:

| Class | Candidates | Contract |
| --- | --- | --- |
| `FUZZ_VEC_REJECT` (`fuzz_add_out_of_range`) | below-min, above-max — per **bounded** side, integer ranges as well as real | the call must be **rejected** (`TA_BAD_PARAM` / an invalid lookback) |
| `FUZZ_VEC_SENTINEL` (`fuzz_add_default_sentinel`) | `TA_REAL_DEFAULT` (`-4e37`), `TA_INTEGER_DEFAULT` (`INT_MIN`) — per **Range** parameter, bounded or not | the call must **succeed with the declared default's result** |

Because it stays metadata-driven, the 19 bounded reals and every bounded integer range
come for free, and so does anything bounded added later.

**Pointed at `:4670` only.** `fuzz_build_vectors(..., frozenOracle=1)` — the `:3942`
v0.6.4 path — emits neither class, exactly as you asked. Evidence that they are really
separated: `--fuzz-064` reports `comparisons: 171230   matches: 148119   benign: 30
cci-tolerated: 8434   fma-tolerated: 14647   failures: 0` **byte-identical** with and
without this commit.

**Both tiers**, since lookback carries its own copy of the validation. The batch call
reuses the existing verdict machinery (`codegen_hash_compare` / `codegen_compare_tol`
diff `retCode` first and short-circuit to MATCH once both sides agree on the same
non-success code). The lookback tier gets `xlang_lookback_leg`, one
`abstract_get_lookback` per parameter vector — lookback is a pure function of the
parameters, so it needs no data. C and Java report a rejected lookback as `-1`, the Rust
crate as `usize::MAX`; both normalize to `-1` before comparing.

**The sentinel is asserted inside each server**, not against the C golden: the same call
with the sentinel and with the explicit default must agree bit-for-bit. "The sentinel
selects the default" is a property of one implementation, and asserting it
language-internally needs no tolerance, so Java's fdlibm transcendentals (MAMA,
LINEARREG_ANGLE) stay bitwise here like everything else. Nothing is lost — `vec[0]`
already gates the *default's* result across languages in the same sweep, so
`server(sentinel) == server(default) ≈ golden(default) == golden(sentinel)` closes the
loop transitively.

**Three self-checks keep it from passing vacuously:** an out-of-range vector the
in-process C library *accepts* fails the run (the candidate has drifted from what the
generated code validates, so its parity assertion would be meaningless); a sentinel
vector where C does not reproduce the explicit default's result fails the run; and an
unfiltered run producing zero cases in either class at either tier fails as
`VACUOUS CONTRACT LEG`.

Summary line on a clean run:

```
lookback tier: 1128 case(s) compared (192 out-of-range, 109 sentinel)
parameter-contract parity (issue #148):
  rejection — 15552 batch + 192 lookback case(s) on below-min / above-max values of every bounded optional parameter; C rejected all of them
  sentinel  — 8829 batch + 109 lookback case(s) on TA_REAL_DEFAULT / TA_INTEGER_DEFAULT; C resolved every one to the explicit default's result
```

### The new assertion fails without the backend change

Each half proven separately, `ta_regtest --xlang-hash --language=rust`, on the same
tree with only `backends/rust_lang.rs::gen_opt_param_validation_with` varied and
`scripts/build.py generate` re-run:

| `ParamType::Real` arm | Result |
| --- | --- |
| **as merged** | `PASS — … Rust: 226470 cases, 0 mismatch(es)` (exit 0) |
| **deleted** (the `dev` state) | `FAIL — 3129 output mismatch(es) + 0 input-port mismatch(es) across 13 function(s)` (exit 84) |
| **range check kept, sentinel substitution removed** (#148's original proposal) | `FAIL — 1855 output mismatch(es) + 0 input-port mismatch(es) across 13 function(s)` (exit 84) |

Full output is in the PR comment below. The 13 functions are the 11 with a bounded real
parameter plus **BBANDS** and **STDDEV**, whose reals are `[TA_REAL_MIN, TA_REAL_MAX]` —
substitution is their only contract, so they are reachable only by the sentinel class.

### Two things worth flagging

1. **A latent server hazard this surfaced, not fixed here.** Every generated server runs
   an *unguarded* timing rerun after the guarded call, unconditionally — and "every
   optional parameter resolved (never a sentinel) and in-range" is an unguarded
   **precondition** (`src/tools/ta_regtest/CLAUDE.md`, VARIANT gate). Both contract
   classes violate it by construction, and the Java server duly died twice:
   `ArrayIndexOutOfBoundsException: Index 40` (below-min period) and `Index -2147483634`
   (the raw `INT_MIN` sentinel), both inside `linearRegAngleUnguardedInternal`. The gate
   works around it by always taking the hash path for contract vectors, which returns
   right after the guarded call. The robust fix is for the servers to skip the unguarded
   rerun when the guarded call failed — a generator change across four backends, so I
   left it out of this PR. Happy to send it separately.
2. **Cost.** Contract vectors run on the full range only (validation precedes all
   `startIdx`/`endIdx` logic, so the contract is subrange-independent; shapes × seeds ×
   sizes still repeats each 81 times). Full two-server `--xlang-hash` goes 63.8s → 72.8s
   on this host, +14%.

`src/tools/ta_regtest/CLAUDE.md` is updated with the spec for all of the above.

## Verification

Run on macOS (Apple clang 21), toolchain 1.97.0 as pinned by `rust-toolchain.toml`.

| # | Check | Result |
| --- | --- | --- |
| 1 | `python3 scripts/build.py generate`, then `git status --porcelain` | Clean at HEAD (the CI `regen-check` gate). Before commit: **15** Rust output files changed, **0** files under `src/`, `include/`, `ta_codegen/output/java`, `ta_codegen/output/dotnet` |
| 2 | `cargo test` (generator) | `47 passed; 0 failed` |
| 3 | `cargo test` (`ta_codegen/output/rust`) | `23 passed; 0 failed` + `340 passed; 0 failed` doctests |
| 4 | `python3 scripts/build.py clippy` | `Clippy clean (generator + generated crate).` — `--all-targets -D warnings`, pedantic |
| 5 | `python3 scripts/build.py test` | `* All tests succeeded. *` — includes the `PERIOD1/BOUNDARY`, `MAVP/GROUPING`, all-candlesticks and `UNGUARDED,TA_S_,VARIANT` legs |
| 6 | `ta_regtest --codegen-only --language=c,rust,java` vs the frozen `reference-pre-cutover` oracle | `C-ref 161/0`, `C 161/0`, `Rust 161/0`, `Java 161/0` — *"All 3 language(s) passed codegen verification"* |
| 7 | Behavior before/after + C reference | The table under **Behavior change**, above |
| 8 | `ta_regtest --xlang-hash --language=rust` (the new gate) | `PASS`, `Rust: 226470 cases, 0 mismatch(es)`; **fails on both sabotages** — see **Tests**, above |
| 9 | `ta_regtest --xlang-hash` (Rust + Java) | Rust `0 mismatch(es)`; Java `14 mismatch(es)`, **identical to the same run without this commit** — a pre-existing `HT_TRENDMODE` integer-output divergence on this host's Temurin 17.0.20 fdlibm, on shape 3 only. Not introduced or masked here |
| 10 | `ta_regtest --fuzz-064` | `comparisons: 171230 … failures: 0`, **byte-identical** with and without this commit — the frozen v0.6.4 oracle at `:3942` is provably untouched |
| — | `cargo doc --no-deps` | Warning-free (the doc emitter changed) |
| — | `python3 scripts/build.py format-check` | `all 171 input file(s) already formatted` |
| — | `python3 scripts/build.py check-source-lists` | `lists agree across CMake and autotools (38 files). OK.` |

Check 6 needed `git fetch origin tag reference-pre-cutover` first; a shallow or
single-branch clone does not carry the tag, and `scripts/regtest.py` builds the oracle
worktree from it.

### Declared limitation

**.NET was not compiled locally** — this host has .NET SDK 7.0.200 against a `net10.0`
target, so that leg of `scripts/build.py servers` fails. The same limitation was declared
in #143 and accepted. It is low-risk: .NET P/Invokes the C library, it is not a row in
`--xlang-hash`, and check 1 shows its output bytes are untouched.

**Java was** — since this PR now adds a test that drives the Java server, leaving it
unverified would have been a poor trade. A Temurin 17.0.20 JDK was fetched into a scratch
directory (no system install) so `scripts/build.py servers` completes the Java legs, and
checks 6 and 9 above are real Java runs. That is how the pre-existing `HT_TRENDMODE`
divergence was identified as pre-existing rather than mine.

## How this surfaced

We vendored `ta_codegen/output/rust/library` into a Rust trading engine and ran all 161
functions point-by-point against python TA-Lib 0.7.1 across three parameter sets each. T3's
non-default set produced 564 points where python raised and Rust returned a value. Chasing
that one function led to the backend, where it turned out not to be a T3 problem at all.

### DIFF
--- a/docs/param-boundary-audit.md
+++ b/docs/param-boundary-audit.md
@@ -73,6 +73,11 @@ so the change is comment-only there.
   lookback formula) are left to the ASan/UBSan job rather than forced by the
   boundary sweep: they are values no caller passes and are unsafe to provoke in a
   plain build. The sweep uses an overflow-safe "large" period near the data length.
-- **Real-parameter range rejection**: the library does not range-check real
-  optional parameters, so the sweep exercises real endpoints for the finite-output
-  scan only, not for `TA_BAD_PARAM` behavior.
+- **Real-parameter range rejection**: the library *does* range-check real optional
+  parameters that declare a finite bound — 19 parameters across 11 functions
+  (`ta_T3.c:76-79` is the shape; the emitter is `backends/c.rs`, mirrored in
+  `backends/java.rs` and `backends/rust_lang.rs`). The remaining 5 real optional
+  parameters declare `[TA_REAL_MIN, TA_REAL_MAX]`, which constrains nothing and
+  emits no check. The sweep still exercises real endpoints for the finite-output
+  scan only, not for `TA_BAD_PARAM` behavior — extending it to assert the rejection
+  is open work, not a property the library lacks.
--- a/src/tools/ta_regtest/CLAUDE.md
+++ b/src/tools/ta_regtest/CLAUDE.md
@@ -365,36 +365,109 @@ Architecture (see `fuzz_data.h`, the Rust port in
 - **Coverage:** every function × 9 shapes × 3 seeds × 3 sizes × parameter
   vectors × 3 subranges ≈ 182k comparisons **per server**, ~94% with non-empty
   output (a non-vacuity guard fails the run if nothing produced output — an empty
-  output hashes the same on both sides).
+  output hashes the same on both sides). Plus the rejection leg below.
 
-Scope rules (deliberate):
-- **No 0.6.4, no waivers; one tolerance + one ill-conditioning skip.** This is
-  current-vs-current across languages, so — unlike `--fuzz-064` — there are none
-  of the `#98`/`#107`/FMA-transition carve-outs. Every case is bitwise except
-  Java's transcendental calls (1e-9, above). A non-tolerated mismatch is a real
-  fusion-site / codegen divergence to fix.
-- **The one ill-conditioning skip: HT_DCPHASE / HT_SINE on the constant shape,
-  Java only.** These two derive their output from `atan2` of the Hilbert
-  transform's in-phase/quadrature components. On `FUZZ_CONSTANT` (flat O=H=L=C,
-  zero variance) those components are floating-point noise (~0), so the phase is
-  `atan2(≈0,≈0)` — chaotically sensitive to the last bit of every transcendental
-  step. C and Rust share the system libm and stay **bit-identical** there (Rust:
-  0 mismatches on every shape); Java's fdlibm differs by ~1 ULP and this
-  ill-conditioning amplifies it to whole *degrees* (~2.9° on HT_DCPHASE). It is
-  not a codegen bug — all 8 non-degenerate shapes agree within 1e-9, and `atan2`
-  of a null signal is mathematically undefined — so no fixed tolerance can
-  separate it from fdlibm noise. `xlang_java_illcond` skips exactly these two
-  functions on exactly the constant shape for the Java leg; the count is reported
-  in the summary. Rust still gates HT_DCPHASE/HT_SINE bitwise on the constant
-  shape, so the C computation itself stays covered there.
-- **period == 1 is in scope** (no 0.6.4 to trip on it), though the shared
-  `fuzz_build_vectors` currently floors periods at 2; period-1 parity is also
-  covered by the `--codegen` edge sweeps.
-- Why this is expected GREEN (PR #96): every backend fuses the identical `a*b+c`
-  sites via the shared `backends/fma.rs` detector and builds with
-  `-ffp-contract=off`, and `fma`/`mul_add` are IEEE correctly-rounded → bit-
-  identical for equal operands (Java `Math.fma` included; only its fdlibm
-  transcendentals need the tolerance).
+### The parameter-contract legs — both classes, both tiers (issue #148)
+
+Everything above compares **values**. Until #148 nothing compared what a call is
+supposed to *do with its parameters*, because `fuzz_build_vectors` never produced
+a vector that exercised either half of the contract: integer candidates are
+clamped into `[min, max]`, real candidates are the `suggested_*` values (in-range
+by construction), and no candidate was ever a `TA_*_DEFAULT` sentinel. That blind
+spot let the Rust backend ship with **no validation at all** for `real` optional
+parameters — neither range check nor default substitution, at any tier — through
+every gate in this file. The whole verification table of the fix passed equally
+well *without* the fix.
+
+Two candidate classes now close it, both metadata-driven off `ta_abstract`, so
+they cover every range-typed parameter today and anything added later on the day
+it lands:
+
+| Class | Candidates | Contract |
+|---|---|---|
+| `FUZZ_VEC_REJECT` (`fuzz_add_out_of_range`) | below-min, above-max — per **bounded** side | the call must be **rejected** (`TA_BAD_PARAM` / an invalid lookback) |
+| `FUZZ_VEC_SENTINEL` (`fuzz_add_default_sentinel`) | `TA_REAL_DEFAULT` (`-4e37`), `TA_INTEGER_DEFAULT` (`INT_MIN`) — per **Range** parameter, bounded or not | the call must **succeed with the declared default's result** |
+
+- **Scope: `--xlang-hash` only.** `fuzz_build_vectors(..., frozenOracle=1)` — the
+  `--fuzz-064` path — emits neither class. That oracle certifies *numeric*
+  agreement with a shipped binary; the parameter contract is a separate question
+  and is deliberately not tangled into it. (`--fuzz-064`'s comparison counts are
+  byte-identical before and after this feature, which is the check that proves
+  the separation.)
+- **An unconstrained side contributes no reject candidate.** `TA_REAL_MIN/MAX`,
+  `TA_INTEGER_MIN/MAX` reach the backends as the widest representable value and
+  every backend skips the comparison for them (the shared unbounded-range
+  magnitude test in `backends/c.rs`, `rust_lang.rs`, `java.rs`), so there is no
+  rejection to compare. `optInPenetration` is `[0, TA_REAL_MAX]`: below-min is a
+  real gate, above-max is not. The **sentinel** class has no such exemption —
+  which is exactly what gates the five `[TA_REAL_MIN, TA_REAL_MAX]` reals
+  (`BBANDS.optInNbDevUp/Dn`, `STDDEV`/`VAR.optInNbDev`, `SAREXT.optInStartValue`):
+  substitution is their *only* contract.
+- **`IntegerList` (the MAType-style enums) is excluded from both**, for different
+  reasons. It declares no range, so there is no bound to exceed (the equivalent
+  out-of-**list** injection lives in the stream vector builder). And C maps
+  `TA_INTEGER_DEFAULT` to the declared default for enums while the Rust backend
+  does not — a known residual divergence tracked in
+  `ta_codegen/generator/CLAUDE.md`, on a different set of functions.
+- **Both tiers.** The lookback entry point carries its own copy of the validation,
+  so a backend can diverge at one tier and not the other. `xlang_lookback_leg`
+  sweeps `abstract_get_lookback` once per parameter vector (lookback is a pure
+  function of the parameters, so it needs no data). C and Java signal rejection
+  with `-1`, the Rust crate with `usize::MAX`; both normalize to `-1` before
+  comparing.
+
+**How each class is asserted**
+
+- *Reject* — against the C golden, reusing the existing verdict machinery:
+  `codegen_hash_compare` / `codegen_compare_tol` diff `retCode` first and
+  short-circuit to MATCH once both sides agree on the same non-success code, so a
+  language that computes a number where C returns `TA_BAD_PARAM` is a hard
+  mismatch.
+- *Sentinel* — **inside each server**, not against the golden: the same call with
+  the sentinel and with the explicit default must agree bit-for-bit
+  (`xlang_hash_call` twice). "The sentinel selects the default" is a property of
+  one implementation, and asserting it language-internally needs no tolerance, so
+  Java's fdlibm transcendentals stay bitwise here like everything else. Nothing is
+  lost: the cross-language correctness of the *default* result is already gated by
+  `vec[0]` in the same sweep, so
+  `server(sentinel) == server(default) ≈ golden(default) == golden(sentinel)`
+  closes the loop transitively.
+
+**Three self-checks keep it from passing vacuously**
+
+1. An out-of-range vector the in-process C library *accepts* fails the run — the
+   candidate, not the implementation, is wrong (the `ta_abstract` range has
+   drifted from what the generated code validates).
+2. A sentinel vector where C does *not* reproduce the explicit default's result
+   fails the run — otherwise "both languages agree" would stop meaning the
+   contract holds.
+3. An unfiltered run that produces **zero** cases in either class, at either tier,
+   fails as `VACUOUS CONTRACT LEG`.
+
+**Contract vectors always take the hash path, and only the full range.** Hash mode
+(`want_hash` / `gen_present`) makes the server return right after the **guarded**
+call; the tolerance path instead falls through to the server's *unguarded* timing
+rerun, where "every optional parameter resolved (never a sentinel) and in-range"
+is a **precondition** (see the VARIANT gate above) that both classes violate by
+construction. Sending them down that path killed the Java server twice —
+`ArrayIndexOutOfBoundsException: Index 40` (below-min period) and
+`Index -2147483634` (the raw `INT_MIN` sentinel), both inside
+`linearRegAngleUnguardedInternal`. The servers would be more robust skipping the
+unguarded rerun when the guarded call failed; that is a generator change across
+four backends, so this gate simply does not ask them to do it. And because
+validation runs before any `startIdx`/`endIdx` logic, the contract is
+subrange-independent — the contract vectors run on the full range only (still 81
+repetitions each across shapes × seeds × sizes), which keeps the added cost to
+about +14% wall clock on a two-server run.
+
+**Sabotage-proven, both halves independently.** Against `dev` +
+`--language=rust`:
+
+| `rust_lang.rs::gen_opt_param_validation_with` | Result |
+|---|---|
+| as merged | `PASS`, `Rust: 226470 cases, 0 mismatch(es)` |
+| `ParamType::Real` arm deleted (the `dev` state) | `FAIL — 3129 output mismatch(es) ... across 13 function(s)` |
+| Real arm kept, **sentinel substitution only** removed | `FAIL — 1855 output mismatch(es) ... across 13 function(s)` |
 
 ## `server_verify` — bitwise C⇄server on the hard-coded tests (issue #115)
 
--- a/src/tools/ta_regtest/test_codegen.c
+++ b/src/tools/ta_regtest/test_codegen.c
@@ -3445,15 +3445,154 @@ static unsigned long long fuzz_parse_hash(const char *resp)
     return strtoull(h, NULL, 16);
 }
 
-/* Parameter vectors: defaults + one-param-varied boundary/list sweeps. */
+/* Parameter-contract vector classes (issue #148). A vector is either an ordinary
+ * value sweep, or one of the two CONTRACT classes below -- the ones a correct
+ * implementation must do something specific with rather than merely compute. */
+#define FUZZ_VEC_NORMAL   0
+#define FUZZ_VEC_REJECT   1  /* below-min / above-max: must be REJECTED          */
+#define FUZZ_VEC_SENTINEL 2  /* TA_*_DEFAULT: must resolve to the declared default */
+
+/* Out-of-range candidates for one bounded optional parameter: exactly the two
+ * values the documented range excludes at its edges -- below-min and above-max.
+ *
+ * They are the ONLY candidates this builder produces that a correct
+ * implementation must REJECT. Every other candidate is deliberately pulled
+ * INSIDE the range (integers clamp to [min,max]; reals take suggested values,
+ * which are in-range by construction, and `continue` otherwise), so the sweep
+ * could only ever assert that VALUES agree -- never that REJECTIONS agree.
+ * That blind spot is why #148 (the Rust backend emitting no validation at all
+ * for `real` optional parameters, at any tier) survived every cross-language
+ * gate in this file.
+ *
+ * Emitted for the current-vs-current path only (frozenOracle == 0). Error-path
+ * parity against the frozen v0.6.4 binary is a separate question from the
+ * numeric agreement that oracle exists to certify.
+ *
+ * A side is skipped when the metadata declares it unconstrained
+ * (TA_INTEGER_MIN/TA_INTEGER_MAX, TA_REAL_MIN/TA_REAL_MAX): such a bound
+ * reaches the backends as the widest representable value and every backend
+ * skips the comparison for it (the shared "unbounded range" magnitude test in
+ * backends/c.rs, rust_lang.rs and java.rs), so there is no rejection to
+ * compare. optInPenetration is [0, TA_REAL_MAX]: below-min is a real gate,
+ * above-max is not. Nothing emitted here can collide with the "use the
+ * default" sentinels (TA_INTEGER_DEFAULT == INT_MIN, TA_REAL_DEFAULT ==
+ * -4e37) -- both lie outside the clamps below.
+ *
+ * IntegerList/RealList (the MAType-style enums) declare no range at all, so
+ * they have no bound to exceed and contribute nothing here; the equivalent
+ * out-of-LIST injection lives in the stream vector builder. Range parameters
+ * are exactly the population the "below-min / above-max" contract applies to.
+ *
+ * Metadata-driven: every bounded parameter is covered today, and any bounded
+ * parameter added later is covered the day it lands, with no list to maintain. */
+static void fuzz_add_out_of_range(const TA_OptInputParameterInfo *oi,
+                                  double cand[FUZZ_MAX_CAND],
+                                  char candKind[FUZZ_MAX_CAND],
+                                  int *nc, int *overflow)
+{
+    double oor[2];
+    int    noor = 0;
+
+    if( oi->type == TA_OptInput_IntegerRange )
+    {
+        const TA_IntegerRange *r = (const TA_IntegerRange *)oi->dataSet;
+        if( !r ) return;
+        if( (long long)r->min - 1 >= (long long)TA_INTEGER_MIN )
+            oor[noor++] = (double)((long long)r->min - 1);
+        if( (long long)r->max + 1 <= (long long)TA_INTEGER_MAX )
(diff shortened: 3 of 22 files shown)


mario4tier:
The fix is right, and the write-up is exemplary. Reviewed against the tree.

**Approach accepted, including the `-4e37` half.** You're right that #148's "range check only" direction was circular, and right to go past it: `ta_T3.c:76-77` substitutes the sentinel, so a bare range check would have made `-4e37` newly return `BadParam` in Rust and traded one divergence for another. Matching C is the correct target. The behavior change is real and worth the release note you flagged — the rejection is one the C library has always performed, so this is Rust catching up, not new library semantics.

The `match` shape, the unbounded-range magnitude test carried over verbatim, and the exponent-form literals are all fine. Confining Enum to a follow-up is the right scoping call, and narrowing the known-issues note to the residual `i32::MIN` substitution rather than deleting it is exactly what I'd have asked for. Yes to that follow-up when you're ready.

Not compiling Java and .NET locally is acceptable here for the reason you give — check 1 shows those output bytes untouched, and the nightly covers them.

## Blocking: this needs a test before it merges

Everything in your verification table passes equally well *without* the generator change. Check 6 in particular runs 161/161 on both sides of this commit, because no vector it drives is ever out of range. That's the gap that let the defect exist for as long as it did, and shipping the fix without closing it means the next refactor of `gen_opt_param_validation_with` can silently reopen it.

Please extend the sweep we already have rather than adding a harness. `fuzz_build_vectors` (`src/tools/ta_regtest/test_codegen.c:3448`) walks the `ta_abstract` metadata per function and builds one-param-varied vectors — integer ranges get min / min+1 / min+7 / default±1 / default+3, integer lists get every enum value, real ranges get suggested_start and suggested_end. Those feed the current-vs-current cross-language comparison at `:4670` and the frozen v0.6.4 oracle at `:3942`.

The reason it never fired: every candidate is deliberately pulled *inside* the documented range — integers clamp to `[min, max]`, reals `continue` when `v < r->min || v > r->max`, and suggested values are in-range by construction. `-4e37` is never emitted either. So the builder can express "these values agree" and cannot express "these values are both rejected" or "this sentinel resolves to the same default".

What the PR needs, then:

1. **Out-of-range candidates plus retcode parity**, below-min and above-max, asserted on the current-vs-current path at `:4670` — every server must agree on `TA_BAD_PARAM`. No reason to restrict it to reals while you're in there.
2. **A `-4e37` vector**, so the sentinel half you added is gated too. Right now it is exactly as untested as the range half was, and it's the half most likely to be "simplified" away by someone who reads the Rust API as sentinel-free.
3. **Point both at `:4670` only.** The frozen v0.6.4 oracle at `:3942` exists to certify numeric agreement against a shipped binary; error-path parity against it is a separate question and shouldn't get tangled into this PR.
4. **Cover the lookback tier**, not just the batch call — your own table shows `t3_lookback(5, 1.05)` was `24` before and `usize::MAX` after, so the divergence lived there too.

There's a precedent for the assertion, just not on this path: the stream vector builder around `:2210` injects one out-of-list value per enum parameter and requires batch and stream to agree on rejecting it. That's enum-only and batch-vs-stream inside a single server rather than across languages, so treat it as the shape to copy, not as coverage you already have.

Because it stays metadata-driven, your 19 parameters come for free and so does anything bounded added later — that's the property worth having, more than the fix itself.

When you push it, please confirm in a comment that the new assertion fails against `dev` without the `rust_lang.rs` change. That's the part that proves the gate has teeth.

kevinlincg:
Confirming what you asked for: **the new assertion fails against `dev` without the `rust_lang.rs` change**, and each half of the arm is proven independently.

Same tree, same commit, only `backends/rust_lang.rs::gen_opt_param_validation_with` varied, `python3 scripts/build.py generate` + the Rust server rebuilt between runs. All three are `ta_regtest --xlang-hash --language=rust`.

---

### 1. As merged — PASS

```
Output parity gate (168 function(s) x shapes x seeds x sizes x params x subranges)...

---------------------------------------------
golden cases: 225342   (in-process C library; 209790 value case(s), 197870 with non-empty output = 94% non-vacuous; 15552 out-of-range case(s) empty by design)
Rust: 226470 cases, 0 mismatch(es)
lookback tier: 1128 case(s) compared (192 out-of-range, 109 sentinel)
parameter-contract parity (issue #148):
  rejection — 15552 batch + 192 lookback case(s) on below-min / above-max values of every bounded optional parameter; C rejected all of them
  sentinel  — 8829 batch + 109 lookback case(s) on TA_REAL_DEFAULT / TA_INTEGER_DEFAULT; C resolved every one to the explicit default's result
PASS — every server matches the in-process C library: BIT-IDENTICAL (zero tolerance), Java transcendentals within 1e-09 (current-vs-current, all shapes, period>=2); retCode and lookback agree on every out-of-range parameter, and every default sentinel resolves to the same value at both tiers.
```
exit 0

---

### 2. `ParamType::Real` arm deleted — this is the `dev` state — FAIL

```rust
        ParamType::Real | ParamType::Enum(_) | ParamType::Price(_) => {}
```

```
Output parity gate (168 function(s) x shapes x seeds x sizes x params x subranges)...
  XLANG LOOKBACK MISMATCH TA_CDLABANDONEDBABY  C(golden) vs Rust  [out-of-range vector: both tiers must REJECT]  params: optInPenetration=-1
    lookback -1/12 (golden/Rust; -1 = parameters rejected)
  XLANG MISMATCH TA_CDLABANDONEDBABY  C(golden) vs Rust  [out-of-range vector: the two languages must agree on REJECTING it]  shape=0 seed=1 n=40 range=[0,39]  params: optInPenetration=-1
    retCode 2/0  begIdx 0/12  nbElem 0/28  hash d725c98e9509182e/ef9a813e60c978ad (golden/Rust)
  XLANG MISMATCH TA_BBANDS  C(golden) vs Rust  [sentinel vector: the two languages must agree on resolving it to the default]  shape=0 seed=1 n=40 range=[0,39]  params: optInTimePeriod=20 optInNbDevUp=-4e+37 optInNbDevDn=2 optInMAType=0
  XLANG LOOKBACK MISMATCH TA_MAMA  C(golden) vs Rust  [out-of-range vector: both tiers must REJECT]  params: optInFastLimit=-0.99 optInSlowLimit=0.05
    lookback -1/32 (golden/Rust; -1 = parameters rejected)
  XLANG MISMATCH TA_T3  C(golden) vs Rust  [out-of-range vector: the two languages must agree on REJECTING it]  shape=0 seed=1 n=40 range=[0,39]  params: optInTimePeriod=5 optInVFactor=-1
    retCode 2/0  begIdx 0/24  nbElem 0/16  hash d725c98e9509182e/74b884000d5d081d (golden/Rust)

---------------------------------------------
golden cases: 225342   (in-process C library; 209790 value case(s), 197870 with non-empty output = 94% non-vacuous; 15552 out-of-range case(s) empty by design)
Rust: 226470 cases, 3129 mismatch(es)
lookback tier: 1128 case(s) compared (192 out-of-range, 109 sentinel)
parameter-contract parity (issue #148):
  rejection — 15552 batch + 192 lookback case(s) on below-min / above-max values of every bounded optional parameter; C rejected all of them
  sentinel  — 8829 batch + 109 lookback case(s) on TA_REAL_DEFAULT / TA_INTEGER_DEFAULT; C resolved every one to the explicit default's result
FAIL — 3129 output mismatch(es) + 0 input-port mismatch(es) across 13 function(s).
```
exit 84 (excerpted — the reporter caps at 3 lines per function)

The 13: `BBANDS CDLABANDONEDBABY CDLDARKCLOUDCOVER CDLEVENINGDOJISTAR CDLEVENINGSTAR CDLMATHOLD CDLMORNINGDOJISTAR CDLMORNINGSTAR MAMA SAR SAREXT STDDEV T3`. Both tiers fire (17 `LOOKBACK MISMATCH` lines alongside the batch ones), and both classes fire (33 `out-of-range vector` lines, 6 `sentinel vector` lines).

---

### 3. Range check kept, **sentinel substitution only** removed — FAIL

This is #148's original "Rust has no `-4e37` sentinel, so the check is a plain `if`" proposal, and it is the shape you flagged as most likely to be simplified back in later:

```rust
        ParamType::Real => {
            if opt.default.is_some() {
                if let Some((lo, hi)) = opt.range {
                    if lo > f64::MIN / 2.0 || hi < f64::MAX / 2.0 {
                        out.push_str(&format!("{pad}if ({name} < {lo:e}) || ({name} > {hi:e}) {{\n"));
                        out.push_str(&format!("{pad}    {err_return}\n"));
                        out.push_str(&format!("{pad}}}\n"));
                    }
                }
            }
        }
```

produces, in `t3.rs`:

```rust
        if (optInVFactor < 0e0) || (optInVFactor > 1e0) {
            return usize::MAX;
        }
```

```
  XLANG SENTINEL MISMATCH TA_T3 [Rust]: the default sentinel does not select the declared default  shape=0 seed=1 n=120 range=[0,119]  params: optInTimePeriod=5 optInVFactor=-4e+37
    sentinel retCode 2 begIdx 0 nbElem 0 hash d725c98e9509182e
    default  retCode 0 begIdx 24 nbElem 96 hash 125b6d08fb54e6ca
  XLANG SENTINEL MISMATCH TA_BBANDS [Rust]: the default sentinel does not select the declared default  shape=0 seed=1 n=40 range=[0,39]  params: optInTimePeriod=20 optInNbDevUp=-4e+37 optInNbDevDn=2 optInMAType=0

---------------------------------------------
Rust: 226470 cases, 1855 mismatch(es)
FAIL — 1855 output mismatch(es) + 0 input-port mismatch(es) across 13 function(s).
```
exit 84

`retCode 2` where the default gives `retCode 0` and 96 values — precisely the new divergence that adding the range check alone would have introduced.

---

### Separation from the frozen oracle

`fuzz_build_vectors(..., frozenOracle=1)` emits neither class, so `:3942` is untouched. `ta_regtest --fuzz-064`, byte-identical with and without this commit:

```
comparisons: 171230   matches: 148119   benign(signed-zero): 30   cci-tolerated: 8434   fma-tolerated: 14647   failures: 0
PASS — current library is bit-identical to v0.6.4 at period>=2 (...)
```

### Regen + the rest

`python3 scripts/build.py generate` then `git status --porcelain` shows only the two files this commit touches — the `regen-check` gate stays green. `build.py test` → `* All tests succeeded. *`; `--codegen-only --language=c,rust,java` → `All 3 language(s) passed codegen verification`; generator and crate `cargo test` all green (`340 passed` doctests); `clippy` clean; `format-check` and `check-source-lists` clean. `test_codegen.c` compiles at `-Wall -Wextra -Wpedantic -Wshadow` with the same 3 pre-existing `unused-but-set-variable` warnings and no new ones.

Two notes I would rather raise than bury, both in the PR description under **Tests → Two things worth flagging**: the servers' unconditional unguarded timing rerun is a real hazard for any out-of-range or sentinel parameter (it killed the Java server twice while I was building this, inside `linearRegAngleUnguardedInternal`) — the gate routes around it rather than changing four backends, and I am happy to send that hardening separately; and the full two-server `--xlang-hash` goes 63.8s → 72.8s here, +14%.

Also: since this PR now drives the Java server, I stopped declaring Java unverified and fetched a Temurin 17 JDK into a scratch directory so the Java legs actually run. That is how I could establish that the 14 `HT_TRENDMODE` Java mismatches on this host are pre-existing — the identical 14 appear on the same tree with this commit stashed. .NET remains uncompiled locally (SDK 7.0.200 vs `net10.0`); it P/Invokes the C library and is not a row in `--xlang-hash`.

mario4tier:
Merged, with both your commits intact — thanks, this was a good catch and a well-argued fix. You were right to go past the direction #148 proposed: without the `-4e37` half, `t3(x, 5, -4e37)` would have started returning `BadParam` where every other binding takes it as "use the default", and the sentinel leg you added is now the only thing gating the five reals whose bounds constrain nothing.

Follow-up commits on `dev` trim the new comments back to the file's norm, drop some unreachable sentinel handling, prune the rejection sweep to one dataset (a rejected call's verdict comes from the parameter values alone, so 81 datasets bought nothing), and restore `src/tools/ta_regtest/CLAUDE.md`, whose `--xlang-hash` scope rules documented a carve-out that is still live. Nothing needed from you.

Your gate then found a second, older bug immediately. `parser/yaml.rs` mapped `TA_REAL_MAX` to `f64::MAX` while `backends/common.rs` had it as `3e37`, so the emitted checks used `1.7976931348623157e308` — and a "skip unbounded ranges" guard built on that premise dropped the check entirely for `[TA_REAL_MIN, TA_REAL_MAX]` params. `BBANDS(nbDevUp = +inf)` was returning `Success` with `inf` in the output where 0.6.x returned `TA_BAD_PARAM`. Fixed across C, Java and Rust, with the sentinels now pinned to `ta_defs.h`, and the generated code names them (`TA_REAL_DEFAULT`, `REAL_DEFAULT` in Rust) instead of repeating the numbers — Rust callers finally have a way to request a default without hardcoding `-4e37`.

If you have appetite for more: the Rust output is the youngest of the four backends and the least exercised by real users. The known residual is enum optional params — C maps `i32::MIN` to the declared default, Rust falls through its dispatch `match` instead, so `Core::ma(.., i32::MIN)` does not select SMA (noted in `ta_codegen/generator/CLAUDE.md`). Anything you hit vendoring the crate is worth an issue.

kevinlincg:
## arm64 cross-check

Re-ran the verification on Apple Silicon, because the sentinel half of the gate asserts bit-for-bit and everything above was measured on x86_64 only. **The gate is safe on aarch64.**

Apple M5 Max, macOS 26.5.2, `aarch64-apple-darwin`, rustc 1.97.0 (matching `rust-toolchain.toml`), Apple clang 21.0.0, Temurin 17.0.19.

The tri-state proof reproduces not just directionally but with identical counts:

| variant | x86_64 | arm64 |
|---|---|---|
| as merged | 226470 cases, 0 mismatch — PASS | **226470 cases, 0 mismatch — PASS** |
| `ParamType::Real` arm deleted | 3129 mismatches, 13 functions | **3129 mismatches, 13 functions** |
| range kept, sentinel removed | 1855 mismatches, 13 functions | **1855 mismatches, 13 functions** |

Case counts matching exactly is expected — the vector builder is metadata-driven and deterministic — but the mismatch counts matching means the failure modes are identical too, not merely present. `build.py test`, `build.py clippy`, generator `cargo test`, crate `cargo test` (23 + 340 doctests) and `--codegen-only` all pass.

**Java is cleaner here than on my x86 host.** Baseline `a8272d0c2`: 200907 cases, 0 mismatches. With this PR: 226416 cases, 0 mismatches — the new contract vectors add +25509 Java cases and all pass. The 14 pre-existing `HT_TRENDMODE` mismatches I reported on x86 do not reproduce on arm64 + Temurin 17.

**Could not run:** the .NET leg — `dotnet` is not installed on that machine and I did not install it.

### One thing to know before you run this on arm64 yourself

`--fuzz-064` and the codegen ref-sweep **fail on arm64**, and it is not this PR. Baseline without the PR produces bit-identical failure counts (549 failures across 9 functions, `comparisons: 171230`, same single `TA_CORREL` diff of 1.48e-09), which incidentally is direct empirical support for the claim that this change leaves the frozen-oracle path untouched.

The cause is a pre-existing repo issue that only manifests on arm64: the frozen oracle worktrees are configured without `-ffp-contract=off`, so on a target where FMA is baseline ISA clang fuses inside the oracle. Rebuilding them with the flag turns both into PASS. Filed separately as #150 rather than folded in here, since it predates this PR and touches the harness scripts rather than `rust_lang.rs`.

If you do verify on Apple Silicon, applying that flag first will save you chasing 549 failures that have nothing to do with the parameter contract.

Issue #146, opened 2026-07-29
Rust backend: `.to_vec()` in the scratch-buffer election makes the C aliasing optimization unreachable (BBANDS pays ~6-7% for it)


kevinlincg:
The Rust backend turns C's pointer-aliasing scratch-buffer election into a real
allocation + copy, making the optimization the surrounding comment promises
unreachable. Refs at `144c635b`, under `ta_codegen/output/rust/library/`.

### Evidence

`src/ta_func/bbands.rs:252-254` carries the C comment verbatim: *"Identify TWO
temporary buffers among the outputs so the calculation needs no memory
allocation; whenever possible make tempBuffer1 be the middle band output, saving
one copy operation."* C delivers exactly that — `src/ta_func/ta_BBANDS.c:169-183`
assigns pointers, so the copy-back guard at `ta_BBANDS.c:298` is false and the
`memcpy` is skipped.

Rust cannot: `bbands.rs:256-266` writes `tempBuffer1 = outRealMiddleBand.to_vec()`
in every branch, including the always-taken `else` at `:265-266`. `.to_vec()`
always allocates, so `bbands.rs:375`
(`tempBuffer1.as_ptr() != outRealMiddleBand.as_ptr()`) is **always true** — the
copy-back always runs and the promised saving is unreachable; the `== inReal`
guard at `:270` is correspondingly always false. All four branches are in fact
dead: `inReal: &[f64]` can never alias `outReal*: &mut [f64]` in safe Rust.

Cost per BBANDS(SMA) call vs C: 2 extra n-sized allocations, 2 memcpys whose
bytes are overwritten before they are ever read, 1 copy-back C skips. Origin:
`ta_codegen/generator/src/backends/rust_lang.rs:2105-2126` (`needs_to_vec`).

Same pattern at `mavp.rs:264` and `:482`, on the always-taken non-aliasing branch
— `mavp.rs:386-388` already documents that the copy-back must always run in Rust,
so the divergence is known; only correctness was addressed. In
`stoch.rs:336,506,832,1142` / `stochf.rs:312,473,787,1079` the `.to_vec()` is on
the aliasing branch only, unreachable in safe Rust — dead code, no runtime cost.

### Measurements

Interleaved A/B/C, 12 rounds, min ns/bar. Intel i7-10700K, rustc 1.97.0,
`opt-level=3, lto=true, codegen-units=1`. SMA/STDDEV are unmodified controls.

[...]

PR #151, opened 2026-07-30
perf(ta_codegen): #146 — resolve pointer scratch elections in the Rust backend

kevinlincg:
Closes the zero-allocation half of #146.

> **Stacked on #149.** At the time of opening, `dev` is at `a8272d0c2` and #149's two
> commits are not on it, so GitHub shows all three commits here. This PR's own commit
> is `perf(ta_codegen): #146 — resolve pointer scratch elections in the Rust backend`;
> the other two are #149's, unchanged. Once #149 lands the diff converges to the
> zero-alloc commit on its own — nothing to rebase. The two do not overlap in
> `rust_lang.rs` (#149 is in `gen_opt_param_validation_with`, this is a new pass plus
> the `gen_impl_block` hook).

Thank you for prototyping the shape before asking for it, and for the three traps —
they were the expensive part, exactly as you said. Trap 2 in particular is the one I
would not have found from a green test run.

## The rule

Stated over the IR, not over `BBANDS`. `backends::rust_lang::ScratchElection` is a
Rust-only pass hooked at the top of `gen_impl_block` and applied to both batch bodies:

1. match an `if`/`else if`/…/`else` chain whose **every** condition is an input↔output
   pointer equality and whose **every** arm consists only of `scratch = someOutput;`
   elections;
2. take the terminal `else`'s mapping — the binding safe Rust always reaches;
3. delete the chain and rename those scratch names to their elected outputs through
   the rest of the enclosing block;
4. drop any guard the rename has turned into a self-comparison.

What licenses it is Rust's own aliasing rule, which is also what makes C's aliasing
branches dead code: `inX: &[T]` and `outY: &mut [T]` are distinct parameters and can
never overlap, so every `inX == outY` test is statically false; two `&mut [T]` can
never overlap either, so electing a *second* output needs no further check. The entry
point's distinctness test on the three outputs is C parity, not the licence.

**Being general is not being greedy.** Requiring *every* arm to be an election is the
clause that declines `STOCH`, `STOCHF` and `MAVP` — each mixes an allocation and a
`…IsAllocated = 1;` flag into a branch, which is a genuine in-place defence with a
real buffer to allocate, not an election. `MAVP` is inverted as well, so it is
rejected on the first link before the `else` is even reached. I had `MAVP` folded in
on the first pass, by a matcher that only inspected the arm that *runs*; narrowing
the rule to your formulation dropped it, and the `MAVP` legs of the gate went with
it. Tolerating one allocating arm is a widening of the rule for a later change, not
something I bolted on here.

## The four red lines

Verified mechanically over the pass's 424 lines of executable code, doc comments
stripped (they do name `BBANDS`/`MAVP` explanatorily):

| red line | status |
|---|---|
| no function name in the transform or its conditions | **clean** — no `BBANDS`/`STOCH`/`STOCHF`/`MAVP`, any case |
| no per-function table, allowlist or `match func.name` | **clean** — no `func.name`, no name comparison, no table |
| no `tempBuffer1`/`tempBuffer2`/`outRealMiddleBand`/`outRealUpperBand` literals | **clean** — also no `outReal`/`inReal`/`localFinalArray` |
| nothing keyed on `optInMAType == TA_MAType_SMA` | **clean** — no `optInMAType`, `TA_MAType`, `MAType` or `SMA` |

The pass reads only `func.inputs`, `func.outputs`, the `VarType::RealPointer` /
`IntPointer` locals it collects from the body, and statement structure.

`BBANDS` is currently the only function in `input/` written in this shape, and that
is asserted rather than asserted-about: `rust_scratch_election_declines_arms_that_allocate`
sweeps **every** indicator and requires the pass to fire for `bbands` alone. A
hand-picked list would have let a future widening slip past a function the test
forgot.

## The three traps

**1. Match before you recurse.** `ScratchElection::block` runs `election_chain` on the
statement before `descend` walks into any child block, so an inner `else if` link can
never be collapsed out from under the chain. `election_chain` walks the `else` spine
itself rather than relying on the child walk.

**2. No catch-all in the renaming walker.** This one needed a second pass — my first
version had `other => other.clone()` in `rename_shallow`, `other => other` in
`descend` and `_ => {}` in `visit_stmts`. All three are now exhaustive over
`Statement`'s 16 variants, with the leaf variants spelled out. Proof that it bites:
adding a `ProbeLoop { body: Vec<Statement> }` variant to `Statement` produces E0004
at all three sites (`rust_lang.rs:4248`, `:4480`, `:4601`) instead of compiling. Your
description of the failure mode is why I went back for it — a skipped body-bearing
variant is well-typed, generates and compiles cleanly, and only fails on the first
index into a still-empty `Vec`.

**3. The copy-back has to be deleted, not left as dead code.** Clause 4 above.
Confirmed as E0502 rather than something the optimiser folds: after the rename the
guard is `if o.as_ptr() != o.as_ptr() { o[..n].copy_from_slice(&o[..n]) }`, so
dropping the statement is load-bearing. Clause 4 fires only where the rename actually
rewrote a name, which is how a function that elected nothing is never touched.

## Scope: one generated file

`generate` across all 168 functions changes exactly one generated indicator file.

```
$ python3 scripts/build.py generate && git diff --name-only <parent>
ta_codegen/generator/CLAUDE.md
ta_codegen/generator/src/backends/mod.rs
ta_codegen/generator/src/backends/rust_lang.rs
ta_codegen/generator/src/main.rs
ta_codegen/generator/templates/rust/scratch_election.rs
ta_codegen/generator/tests/backend_suite.rs
ta_codegen/output/rust/library/src/ta_func/bbands.rs      <- the only indicator
ta_codegen/output/rust/library/src/ta_func/mod.rs         <- +4, the test module decl
ta_codegen/output/rust/library/src/ta_func/scratch_election.rs  <- the gate, copied
```

`git status --porcelain` after `generate` on the committed tree is empty — the
`regen-check` gate stays green. `mavp.rs`, `stoch.rs` and `stochf.rs` are
byte-identical to their parent.

Other backends, zero changed files each: `src/ta_func/`, `src/ta_abstract/`,
`src/ta_common/`, `ta_codegen/output/java/`, `ta_codegen/output/dotnet/`, `include/`,
`website/`, `ta_func_api.xml`, `CMakeLists.txt`.

The general MA path and both stream paths keep their real allocations — four
surviving `tempBuffer1`/`tempBuffer2` pairs in `bbands.rs` (guarded batch, unguarded
batch, and one per stream variant), pinned by a generator shape test. No `.to_vec()`
remains anywhere in the file.

## The gate is resident, not a one-off probe

Your 4800 comparisons were a throwaway probe; this one runs in CI on every push.

`templates/rust/scratch_election.rs` is hand-written, copied verbatim into the crate
by `generate` (held by the Rust backend's `clean_keep`), and declared `#[cfg(test)]`
in the generated `mod.rs`, so nothing ships in a release build.

**It is a `#[cfg(test)]` module rather than `output/rust/library/tests/*.rs`
deliberately.** An integration test there survives `generate`, but no CI job executes
it — `dev-nightly-tests.yml` runs `cargo test --lib` and `--doc` only. It would have
been a gate that never runs. `templates/rust/types.rs` is the established precedent
for hand-written source shipping inside the generated crate, so this is a second
template alongside it (wired via `RUST_TEMPLATE_MODULES` / `RUST_TEST_ONLY_MODULES`
in `main.rs` and `clean_keep` in `backends/mod.rs` — adding another means touching all
three, which is now written down in `generator/CLAUDE.md`).

Three properties, 5752 bit-exact comparisons, over periods 2/3/5/20/33/100 ×
`nbDev` (2,2)/(0,3)/(-1,2)/(1.5,2.5) × MAType 0/1/2/3/8 × n from 1 to 5000:

| property | comparisons | what a mis-election breaks |
|---|---|---|
| `BBANDS` == `MA` + `STDDEV` recomposed through the public API | 3264 | a genuinely separate code path; covers the electing SMA path *and* the untouched general path |
| independent of the prior contents of the output slices | 2032 | electing an output means reading a buffer the caller owned |
| independent of the output slices' capacity | 456 | the pre-#146 shape sized scratch from the caller's slice |

**All 5752 pass with the transform reverted.** That is the point — they pin numeric
invariance, not the absence of an allocation.

Discrimination, since a gate that cannot fail is not a gate. Four probes, each run
against this tree:

| probe | result |
|---|---|
| standard deviation written into the middle band instead of the upper | **caught** — 2 of 3 legs fail |
| band loop reordered so the scratch is clobbered before it is read | **caught** — composition leg fails |
| matcher widened by dropping the `then`-arm requirement (my v1 rule) | **caught** — the whole-tree non-firing test fails on `MAVP` |
| pass disabled entirely | gate still green, `bbands.rs` byte-identical to parent, 16 `.to_vec()` back |

Standard deviation into the *lower* band is **not** caught, correctly — that is an
equally valid election.

## Verification

Every one actually run, on the committed tree.

| check | result |
|---|---|
| `build.py generate` then `git status --porcelain` | empty |
| `git diff` over C / Java / .NET / include / website | empty |
| `build.py test` | `* All tests succeeded. *` |
| `build.py clippy` | clean, both crates |
| generator `cargo test` | 620 passed |
| crate `cargo test --lib` / `--doc` | 26 / 340 passed |
| `ta_regtest --codegen-only --language=c,rust` | C 161/0, Rust 161/0 |
| `ta_regtest --xlang-hash --language=rust` over BBANDS/MAVP/STOCH/STOCHF + MA family | 93 288 cases, **0 mismatches** — bit-identical to the in-process C library |
| `build.py format-check`, `check-source-lists` | clean |

## Performance

Throwaway interleaved A/B harness outside the tree, per your note — no second
benchmark tool in the released tree. It links the generated library before and after
as two crates in one process, alternates which variant runs first, and checks every
timed shape bit-for-bit before timing it. `SMA` — identical code in both crates — is
the control, so the control delta bounds harness bias.

### x86_64 — i7-10700K, macOS 26.5, rustc 1.97.0, `lto = "thin"`, `codegen-units = 1`, min of 25 rounds

**Control: within 0.13%.**

Exactly-sized output slices, ns/call, period 20:

| bars | before | after | gain |
|---|---|---|---|
| 100 | 508.3 | 296.8 | **+71.3%** |
| 500 | 1924.6 | 1618.2 | **+18.9%** |
| 5000 | 22864.7 | 19977.3 | **+14.5%** |

Flat in the period, as you predicted: at 500 bars +17.7/18.5/18.9/19.8% for periods
5/14/20/50.

Capacity group — 500 bars of data, output slices allocated once at capacity, ns/bar,
median of 8 processes:

| capacity | before (min..max) | after |
|---|---|---|
| 500 | 3.82 (3.77..3.83) | 3.23 |
| 5 000 | 6.90 (6.55..86.89) | 3.23 |
| 50 000 | 98.92 (59.48..153.60) | 3.20 |

The `before` spread is allocator-state dependent — the libmalloc analogue of the
glibc effect you described — and is itself the point; `after` does not move.

### aarch64 — Apple M5 Max, macOS 26.5.2, rustc 1.97.0, median of 11 processes

**Control: within 0.06%.**

| bars | before | after | gain |
|---|---|---|---|
| 100 | 177.3 | 128.4 | **+38.1%** |
| 500 | 815.1 | 693.8 | **+17.2%** |
| 5000 | 9791.7 | 7881.8 | **+24% to +35%** |

The 5000-bar row is reported as a range because it has a period sensitivity I cannot
explain. `after` reads 8185/7589/7882/7493 ns for periods 5/14/20/50 while `before`
stays flat near 10 000 — a 9% spread in `after` alone. It reproduces (under 1% within
each cell) and the output length varies by only 1% across those periods, so it is
neither noise nor work volume. Most likely loop alignment at the 40 KB buffer size.
It does not change the direction or the order of magnitude. Every other row is flat
in the period: at 500 bars +16.1/16.3/17.2/17.4%.

Capacity group — 500 bars of data, ns/bar, 24 processes:

| capacity | before | after |
|---|---|---|
| 500 | **bimodal: 1.62–1.66 (14/24) or 3.85–3.98 (10/24)** | 1.39 (1.38..1.39) |
| 5 000 | 3.62 (3.34..3.69) | 1.39 (1.39..1.40) |
| 50 000 | 24.66 (23.25..24.78) | 1.39 |

No single median is quoted for cap=500 `before` because it is genuinely bimodal
across processes, and collapsing it to one number would misrepresent it.

`after` is **1.38–1.40 ns/bar in all 72 readings** (24 processes × 3 capacities). The
extended sweep takes it further: `before` is linear in the caller's capacity across
four orders of magnitude while `after` stays 1.39, so at capacity 1 000 000 against
500 bars of data `before` reads 397.6–401.6 ns/bar against 1.39 — **289×**. That is
the argument you made in the issue, and it is the one that matters to us.

### Honest comparison to your prototype

These gains run higher than your +9.1% / +9.4% / +8.7–12%. **That is platform and
harness difference, not a claim that this implementation is better than yours** —
different CPUs, macOS libmalloc rather than glibc (so no `mmap`-threshold
transition to fall off), and a min-of-rounds interleaved harness rather than a single
sweep. The generated code is the same shape you described.

## Not in scope

- `STOCH`/`STOCHF`/`MAVP`, per your scope note — reaching them needs a matcher that
  tolerates one allocating arm, which is a widening of the rule and a separate change.
  Their non-firing is what lets me assert the other three backends were untouched.
- The Rust enum-optional-param residual (`Core::ma(.., i32::MIN)` not selecting SMA),
  which you flagged on #149 — happy to take that next if you want it.

### DIFF
--- a/ta_codegen/generator/CLAUDE.md
+++ b/ta_codegen/generator/CLAUDE.md
@@ -36,6 +36,15 @@ include/ta_func.h        (generated public header)
 (which holds only the indicator definitions) and out of `output/` (100% generated):
 - `templates/rust/types.rs` — the `Core` / `RetCode` / `CoreBuilder` / `CandleSettings`
   scaffolding, copied verbatim into the Rust crate (`output/rust/library/src/ta_func/types.rs`).
+- `templates/rust/scratch_election.rs` — the value gate for the scratch-buffer
+  election (issue #146), copied verbatim and declared `#[cfg(test)]` in the
+  generated `mod.rs`, so it never ships in a release build. Run by
+  `cargo test --lib -p ta-lib`.
+
+Both Rust templates are listed in `main.rs`'s `RUST_TEMPLATE_MODULES` (and the
+test-only ones in `RUST_TEST_ONLY_MODULES`) and in `RustBackend::clean_keep`, so
+`generate` copies them in and never deletes them. Adding another one means
+touching all three.
 - `templates/c/ta_retcode.c.template` — spliced with `src/ta_common/ta_retcode.csv`
   (`backends/retcode.rs`) → `src/ta_common/ta_retcode.c`.
 - `templates/c/ta_abstract_serve.c` — hand-written abstract-serve handlers `#include`d
@@ -89,6 +98,11 @@ cd ta_codegen/generator && cargo clippy          # Strict pedantic lints enabled
 ```
 
 Tests are in `tests/backend_suite.rs` and `tests/integration_test.rs` — they verify IR-to-backend rendering, expression types, function signatures, and function variants across all backends.
+
+Value gates that need the *generated* library live in the crate itself, as
+`#[cfg(test)]` modules copied from `templates/rust/` (see
+`RUST_TEMPLATE_MODULES`); run them with `cargo test --lib -p ta-lib` in
+`ta_codegen/output/rust/`.
 
 ## Cross-Language Testing Architecture
 
@@ -196,6 +210,50 @@ asserts `Success`). Crate-level docs, Cargo.toml package metadata, and the crate
 README.md are emitted by `generate_rust_crate_scaffolding` in `main.rs`. Verify
 with `cargo doc --no-deps` (must be warning-free — prose escaping of `[`/`<` is
 load-bearing) and `cargo test --doc` in `ta_codegen/output/rust/`.
+
+### Scratch-buffer election (issue #146)
+
+Several C bodies elect one of their own *output* buffers as the scratch the
+calculation runs in — `BBANDS` opens with `tempBuffer1 = outRealMiddleBand;` so it
+needs no allocation at all. In C that is a pointer assignment. Rust has no pointer
+to assign, so a naive rendering emits `outRealMiddleBand.to_vec()`: an allocation
+plus a copy of bytes that are overwritten before they are read, sized by the
+*caller's slice* rather than by the data range.
+
+`backends::rust_lang::ScratchElection` (a Rust-only pass, applied to both batch
+bodies at the top of `gen_impl_block`) restores C's shape by renaming the local to
+the elected output. It leans on Rust's own aliasing rules — `&[T]` and `&mut [T]`
+parameters can never overlap, and neither can two `&mut [T]` — which is also what
+makes C's aliasing arms, its input-alias guard and its copy-back statically dead
+here.
+
+The rule is stated over the IR, and nothing in the pass names a function, a buffer
+or an MA type: match an `if`/`else if`/…/`else` chain whose *every* condition is an
+input↔output pointer equality and whose *every* arm is only `scratch = someOutput;`
+elections; take the terminal `else`'s mapping; delete the chain and rename through
+the rest of the enclosing block; drop any guard that became a self-comparison.
+Clause 4 is load-bearing rather than cosmetic — left in place, `BBANDS`' copy-back
+would read and write the same `&mut` slice in one statement, which is E0502.
+
+Being general is not the same as being greedy. Requiring *every* arm to be an
+election is what declines `STOCH`, `STOCHF` and `MAVP`: each mixes an allocation
+and a `…IsAllocated = 1;` flag into a branch, which is a genuine in-place defence,
+not an election (`MAVP` is inverted too — allocation in the `then`, election in the
+`else`). Tolerating one allocating arm would reach them; that would be a widening
+of this rule for a later change, never a per-function case. An election also stops
+at the end of the block holding it, so `BBANDS`' general MA path and both stream
+paths keep their real `vec![0.0; ...]` allocations, and the pass backs off entirely
+if the local is assigned again while still in scope.
+
+`BBANDS` is currently the only function in `input/` written in this shape.
+`rust_scratch_election_declines_arms_that_allocate` in `tests/backend_suite.rs`
+pins that by sweeping every indicator and asserting the pass fires for `bbands`
+alone.
+
+The C, Java and .NET backends need none of this — they assign the pointer or
+reference directly — so the transform must never change their output. `generate`
+followed by `git diff` over `src/ta_func/`, `output/java/` and `output/dotnet/` is
+the check. `templates/rust/scratch_election.rs` is the value gate.
 
 ### Debug-safe decrements
 
--- a/ta_codegen/generator/src/backends/mod.rs
+++ b/ta_codegen/generator/src/backends/mod.rs
@@ -181,7 +181,9 @@ impl LanguageBackend for RustBackend {
         ("", ".rs")
     }
     fn clean_keep(&self) -> &'static [&'static str] {
-        &["types.rs", "mod.rs"]
+        // Hand-written modules copied from `templates/rust/` (see
+        // `RUST_TEMPLATE_MODULES`) plus the generated `mod.rs`.
+        &["types.rs", "scratch_election.rs", "mod.rs"]
     }
     fn generate_server(
         &self,
--- a/ta_codegen/generator/src/backends/rust_lang.rs
+++ b/ta_codegen/generator/src/backends/rust_lang.rs
@@ -198,6 +198,15 @@ fn gen_impl_block(func: &FuncDef, enums: &HashMap<String, EnumDef>, registry: &R
     let mut out = String::new();
     let snake = func.name.to_lowercase();
 
+    // C's pointer-based scratch-buffer election becomes a rename here, so the
+    // batch bodies below run the calculation directly in the caller's output
+    // slices instead of allocating and copying (issue #146). Rust-only: the C,
+    // Java and .NET backends assign the pointer/reference and need no rewrite,
+    // and the stream tier keeps the untransformed `func` (it composes its own
+    // scratch buffers). See [`ScratchElection`].
+    let elected = elect_output_scratch(func);
+    let func = &elected;
+
     out.push_str(
         "// Allow non-snake-case names to maintain TA-Lib API compatibility\n\
          #[allow(non_snake_case)]\n\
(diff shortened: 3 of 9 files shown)


kevinlincg:
Rebased onto `dev` @ `9c0b76979`. **The diff is now one commit** (`5aa8b5f72`) — the two #148 commits this branch was stacked on are in `dev` via `aedf4839a`, so they drop out and what remains is only the #146 scratch election.

### Why I rebased even though GitHub said CLEAN

`CLEAN` was accurate about git and misleading about intent. `d135662a4` renamed the sentinels in 15 generated Rust files, `bbands.rs` among them. My base predated it, so my diff carried no `-REAL_DEFAULT` line for git to flag as a conflict — merging as it stood would have quietly put `-4e37` back into `bbands.rs` and broken regen-check, because the generator would no longer agree with its own output. Worth flagging as a general hazard for anything else stacked on #149: a mergeable diff and a non-reverting diff are different properties, and only the second one is what you want.

No conflict actually arose. Our two touch points in `rust_lang.rs` (the hook at `gen_impl_block`, the pass at the end of the file) are nowhere near your emitter at `gen_opt_param_validation_with`, and `format_yaml_num` is a different function from `generate_rust_crate_scaffolding`. I verified that as a result rather than trusting it: `REAL_DEFAULT` and `real_bound_literal` are present in the emitter, `TA_REAL_MIN`/`TA_REAL_MAX` in `format_yaml_num`, and your trimmed known-issue #2 wording in `generator/CLAUDE.md`.

### The generated output carries both

`ta_codegen/output/rust/library/src/ta_func/bbands.rs` after `generate`:

- 8x `optIn... == REAL_DEFAULT`, and the `REAL_MIN`/`REAL_MAX` bounds `2f4393c2b` restored for BBANDS' two unbounded reals — your rename, untouched.
- `to_vec()` count **0**. The SMA fast path runs in `outRealMiddleBand`/`outRealUpperBand` directly.
- `git diff dev HEAD -- bbands.rs | grep -E '^-.*(REAL_DEFAULT|REAL_MIN|REAL_MAX)'` returns nothing — not one line of the rename is removed — and no `+` line anywhere in the PR reintroduces a bare `4e37`.

The cleanest evidence that the two compose rather than collide: **switch the pass off and `bbands.rs` becomes byte-identical to `dev`** (16x `to_vec()` back, `REAL_DEFAULT` still 8x). So this commit's delta on that file is exactly the election and nothing else.

The bit-identity gate needed no change — `templates/rust/scratch_election.rs` never names a sentinel. It drives BBANDS through explicit in-range values, all inside the now-enforced `[REAL_MIN, REAL_MAX]`, so `d135662a4` does not reach it. (Had a leg wanted the symbol, it is `ta_lib::REAL_DEFAULT` and the gate's `use super::*;` already has it in scope.)

### Re-verified on the new base

- `generate` then `git status` — clean.
- Only `bbands.rs` changes among the 168 generated indicators. `mavp.rs`, `stoch.rs`, `stochf.rs`: 0 changed lines, elections and allocating arms intact. `rust_scratch_election_declines_arms_that_allocate` sweeps all of `input/` and asserts the pass fires for `bbands` alone.
- C, Java, .NET, `include/`, `src/ta_abstract/`, `src/ta_common/`, `src/tools/`, website, `ta_func_api.xml`, `CMakeLists.txt`, `Makefile.am`: 0 changed files each.
- generator `cargo test` 621 (620 + your `ta_defs_sentinels_match`); crate `cargo test --lib` 26, `--doc` 340; `build.py test` all succeeded; `clippy`, `format-check`, `check-source-lists` clean.
- `ta_regtest --codegen-only --language=c,rust` — 161/161 per language, `All 2 language(s) passed`.
- `ta_regtest --xlang-hash --language=rust` over BBANDS/MAVP/STOCH/STOCHF and the MA family — 87146 cases, **0 mismatches**, bit-identical to the in-process C library. The trimmed #148 contract leg runs inside it on the same functions: 86 reject batch + 86 lookback, 3483 sentinel batch + 43 lookback.

Discrimination, since a green suite on a rebase proves little on its own: with the pass disabled the gate still passes all 5752 comparisons (it pins values, not the absence of an allocation), and collapsing the election mapping so the standard deviation lands in `outRealMiddleBand` fails it immediately — `middle[0] differs (matype=0 period=2 dev=(2,2) n=2): 1.5856344715222406 vs 102.0570759092143`.


Two numbers in the commit message's Verified block were stale on the new base and are corrected in it: generator `cargo test` 620 -> 621 (your `ta_defs_sentinels_match`), and the `--xlang-hash` sweep 93 288 -> 87 146 cases (`ec71327a4` puts rejection vectors on one dataset instead of 81). Still 0 mismatches.

mario4tier:
Merged to `dev` as `cf55bdfe6`, your commit and authorship intact. Thank you — the
rule is stated where it belongs, the four red lines hold under a mechanical check of
the pass with doc comments stripped, and the gate is the part that will still be
earning its keep in a year.

Verified here before merging, on Linux/glibc, x86_64:

| check | result |
|---|---|
| `generate` over 168 functions → `git status --porcelain` | empty |
| `git diff` over C / Java / .NET / include / website | empty |
| `ta_regtest --codegen-only --language=c,rust` vs the frozen oracle | C 161/0, Rust 161/0 |
| `ta_regtest --xlang-hash --language=rust` | 30 886 cases, 0 mismatches |
| C reference tests, generator suite, crate `--lib` / `--doc` | pass / 622 / 26 / 340 |
| clippy both crates, `format-check`, `check-source-lists` | clean |

I also sabotaged the gate rather than trusting it: rewriting the standard deviation
into the middle band fails 2 of its 3 legs with a bit-level diff. It is not a gate
that cannot fail.

## One follow-up commit on top: `648afed83`

The pass scopes an election to the remainder of the electing block, and the doc
comment calls that *"exactly C's scope for it."* That is the one claim that does not
hold. C's scratch is a **function**-scoped pointer, so one elected inside a nested
block still points at that output after the block ends.

Where control falls out of the electing block and the local is read afterwards, the
rename leaves that read pointing at a `Vec` nothing ever assigns:

```rust
let tempBuffer1: Vec<f64> = Vec::new();          // election deleted, never assigned
…
outRealMiddleBand[i] = ((inReal[i]) as f64);     // renamed inside the block
…
outRealLowerBand[i] = ((tempBuffer1[i]) as f64); // reads the empty Vec
```

Well-typed, generates and compiles cleanly, panics on the first index where C copies
the elected output — the same failure signature as trap 2, which is what made me go
looking for it.

`BBANDS` is unaffected, and for a reason worth naming: its SMA fast path `return`s,
so nothing after the block can observe the local. The pass has been correct in
practice because of a property of `BBANDS` that the rule never asked for.

The fix is one clause beside your two existing guards — decline unless control never
falls out of the block, or no enclosing scope mentions the local again — plus an
`outer` tail threaded through `block`/`descend`. Falls back to the `.to_vec()` copy,
exactly as your other declines do. `BBANDS` still elects, every generated file stays
byte-identical, and the fixture that used to slip through is now in
`backend_suite.rs`. Your `MAVP` instinct was right twice over: narrowing to "every
arm" caught the greedy case, and this is the same lesson one level up.

## Performance

Reproduced the direction, not the magnitude. Same method — both versions in one
process, bit-identity checked before timing (2400 comparisons), `taskset`-pinned:

| bars | gain |
|---|---|
| 100 | +15–20% |
| 250–5000 | +6–14% |
| 20 000 | **+220%** |

500 bars of data into preallocated output slices: **303×** at 50k capacity, **3747×**
at 500k. Flat in the period throughout.

Your +71.3% at 100 bars does not reproduce here — we read +15–20%. The 20 000-bar row
is where glibc takes the scratch to `mmap`, which macOS's libmalloc has no analogue
for, so the platforms disagree about *which* size is dramatic while agreeing on the
shape. The capacity result is the one that transfers cleanly, and it is the one worth
quoting.

## Three notes, none of which changed the merge

- `references_var`'s inner match has a `_ => {}` arm — a fourth walker beside the
  three you made exhaustive. Its failure mode is the safe one (miss a reference →
  decline the election), so this is only a request for a comment saying so, since the
  exhaustiveness argument reads as covering everything.
- `visit_stmts` calls `f(init)`/`f(update)` on `ForC` without recursing into them.
  Harmless while both are leaves.
- `scratch_election.rs` ships in the published `.crate` (307 test-only lines, confirmed
  via `cargo package --list`). `types.rs` is the precedent, so this is fine — but the
  whole `RUST_TEMPLATE_MODULES` / `RUST_TEST_ONLY_MODULES` / `clean_keep` plumbing
  exists only because CI runs `cargo test --lib` and not `--test`. The gate could live
  in `output/rust/library/tests/` — the clean step never touches that directory — with
  one CI line changed to `cargo test -p ta-lib`, dropping the plumbing and keeping test
  source out of the crate. Not worth churning now; worth knowing before a third
  template appears.

Widening the matcher to tolerate one allocating arm, and the Rust enum-optional-param
residual you flagged, are both yours if you want them — separate changes, as you said.
#146 closed.

PR #152, opened 2026-07-31
docs(midpoint,midprice): correct the amortized O(1) claim (#147)

kevinlincg:
Comment-only fix for the part of #147 you accepted. **No algorithm change** — that
discussion stays in #147.

## What the comment claimed

Both functions documented the cached-extremum rolling min/max as:

> a rescan of the window is needed only when the cached extremum drops out of the
> window (amortized O(1) per bar instead of O(period))

## What the code does

The rescan is gated on `highestIdx < trailingIdx` / `lowestIdx < trailingIdx`, and
`trailingIdx` advances one bar per iteration. So an extremum sitting on the **oldest
in-window bar** drops out on the very next bar, the rescan fires again, and per-bar
cost stays O(period) for as long as that condition holds. That is not amortized
O(1) — it is O(1) in the common case with an O(period) worst case that can persist.

These two track *both* extrema, which is what makes the condition self-sustaining
rather than incidental: through a trend leg the high is refreshed by each new bar
while the low stays pinned at the oldest bar for the entire leg (and the reverse on
the way down). A flat stretch pins both. Random-walk input is the favourable case,
where rescans are rare.

Counting the inner-loop comparisons actually executed by the `midpoint` loop
(20k bars, comparisons per bar):

| input | period 14 | period 30 | period 200 |
|---|---|---|---|
| random walk | 3.99 | 5.99 | 13.42 |
| alternating trend/chop (200-bar legs) | 8.40 | 17.17 | 103.50 |
| monotone up or down | 13.00 | 29.00 | 199.01 |

The monotone rows are `period - 1` per bar, i.e. every bar rescans. They are the
clean limit, not the interesting case — the middle row is the ordinary one, and it
tracks period rather than staying flat.

## What this PR changes

The comments now describe that behaviour instead of asserting a bound the code does
not reach: O(1) in the common case, O(period) per bar while an extremum sits at the
trailing edge, and the trending input class where that keeps happening. Wording
follows the surrounding comment style; `midprice`'s existing `<= 20` naive-rescan
bullet was already accurate and is left alone.

Edited only the two source-of-truth inputs:

- `ta_codegen/input/midpoint/midpoint.c`
- `ta_codegen/input/midprice/midprice.c`

then regenerated (`cd ta_codegen/generator && cargo run -- generate`, plus
`generate-servers` for the tracked Java server source, which embeds `Core` and
therefore carries the comment too). The remaining 8 files in the diff are that
generated output — C, Rust and Java all pick up the new wording. The .NET backend
has no per-function generated source, so nothing there changes.

**Zero numerics change.** Every changed line is inside a comment:

```
git show -U0 HEAD | grep -E '^[+-]' | grep -vE '^(\+\+\+|---)' \
  | grep -vE '^[+-][[:space:]]*(\*|//|/\*)'      # -> no output
```

## Verification (all run locally, macOS arm64, Apple clang 21)

| check | result |
|---|---|
| `cargo run -- generate` then `git status --porcelain` | clean (also clean after re-running `generate-servers`) |
| `scripts/build.py format-check` | all 171 input files already formatted |
| `scripts/build.py test` | `* All tests succeeded. Enjoy the library. *` |
| `scripts/build.py clippy` | Clippy clean (generator + generated crate) |
| `ta_codegen/generator` `cargo test` | 70 passed, 0 failed |
| generated crate `cargo test` | 366 passed, 0 failed |
| `scripts/regtest.py --language=c,rust --codegen-only` | all functions PASS vs the reference oracle, incl. MIDPOINT / MIDPRICE |
| `ta_regtest --xlang-hash --language=rust` | `Rust: 211162 cases, 0 mismatch(es)` — PASS, bit-identical |

`--xlang-hash` for Java/.NET was not run here: this box has JDK 11 (the build wants
`--release 17`) and .NET SDK 7 (the server targets net10), so those servers do not
build locally. Pre-existing environment limitation, unrelated to this change — and
since the diff is comments only, CI covers it.

## Notes for review

- I did **not** add a `Change history` entry or contributor initials to either file
  header, to keep the diff to the two comment blocks you pointed at. Happy to add
  them in the repo's format if you'd prefer the correction logged there.
- The comments reference `issue #147`, matching the inline-issue-reference style
  used elsewhere in `ta_codegen/input` (e.g. `cmf.c`, `imi.c`).
- Framing deliberately leads with the trending case and treats monotone/flat as the
  limit, per your note about not optimizing for — or advertising — synthetic input.

Algorithm evaluation (deque vs Van Herk/Gil-Werman vs attacking the rescan), the
benchmark corpus, and the index-returning tie-break question are all out of scope
here and continue in #147.

### DIFF
--- a/src/ta_func/ta_MIDPOINT.c
+++ b/src/ta_func/ta_MIDPOINT.c
@@ -134,9 +134,20 @@ TA_LIB_API TA_RetCode TA_MIDPOINT( int    startIdx,
     * output to be the same buffer.
     *
     * The highest/lowest of the window is cached with its
-    * index; a rescan of the window is needed only when the
-    * cached extremum drops out of the window (amortized O(1)
-    * per bar instead of O(period)).
+    * index; the window is rescanned only when the cached
+    * extremum drops out of the window. That is O(1) per bar
+    * while the extremum sits away from the trailing edge, but
+    * it is not amortized O(1): an extremum on the oldest
+    * in-window bar drops out on the very next bar, so the
+    * rescan repeats and the cost stays O(period) per bar for
+    * as long as that persists.
+    *
+    * Tracking both extrema keeps that state going through a
+    * trend: while the high is refreshed by each new bar, the
+    * low stays pinned at the oldest bar for the whole leg
+    * (and the reverse on the way down). A flat stretch pins
+    * both. Random-walk input is the favourable case, where
+    * rescans are rare. See issue #147.
     */
    outIdx = 0;
    today = startIdx;
@@ -642,9 +653,20 @@ TA_RetCode TA_MIDPOINT_OpenInternal( struct TA_MIDPOINT_Stream **stream, const d
        * output to be the same buffer.
        *
        * The highest/lowest of the window is cached with its
-       * index; a rescan of the window is needed only when the
-       * cached extremum drops out of the window (amortized O(1)
-       * per bar instead of O(period)).
+       * index; the window is rescanned only when the cached
+       * extremum drops out of the window. That is O(1) per bar
+       * while the extremum sits away from the trailing edge, but
+       * it is not amortized O(1): an extremum on the oldest
+       * in-window bar drops out on the very next bar, so the
+       * rescan repeats and the cost stays O(period) per bar for
+       * as long as that persists.
+       *
+       * Tracking both extrema keeps that state going through a
+       * trend: while the high is refreshed by each new bar, the
+       * low stays pinned at the oldest bar for the whole leg
+       * (and the reverse on the way down). A flat stretch pins
+       * both. Random-walk input is the favourable case, where
+       * rescans are rare. See issue #147.
        */
       outIdx = 0;
       today = startIdx;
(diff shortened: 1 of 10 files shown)


mario4tier:
Merged as b4fdc69c8 — thank you. This is exactly the standalone comment fix I was
after, and keeping it to the two inputs plus regenerated output made it quick to
verify: every changed line lands inside a comment, the two shipped C files are
token-identical once comments are stripped, and regenerating from your inputs here
leaves the tree clean, so nothing in the generated output was hand-touched.

The framing is right too — leading with the trend leg and treating monotone as the
limit rather than the headline is the correct way round.

One thing that fell out while I was counting rescans on my side, which you may want
for #147: **a flat stretch is the actual worst case, at `2*(period - 1)`** — worse
than monotone, not equal to it. Counting the inner-loop comparisons the `midpoint`
loop executes, 20k bars, per bar:

| input | period 14 | period 30 | period 200 |
|---|---|---|---|
| random walk | 3.99 | 5.98 | 15.86 |
| monotone up *or* down | 13.00 | 29.00 | 199.01 |
| flat | 26.00 | 58.00 | 398.00 |

The reason is a little counter-intuitive, and it is worth knowing before anyone
reasons about this loop from the tie-breaks: the `else if` arms are `tmpHigh >=
highest` and `tmpLow <= lowest`, which look like they should park both extrema on
the *newest* bar on flat input and make rescans disappear entirely. They never run.
The rescan's inner loop uses strict `>` / `<`, so on ties it leaves `highestIdx` and
`lowestIdx` sitting at `trailingIdx`, and the `< trailingIdx` guard refires on the
next bar — for *both* extrema, indefinitely. Monotone only pins one of the two,
which is why it costs half as much. So your "A flat stretch pins both" is right, for
a slightly different reason than the sentence before it.

The algorithm question stays open in #147. Looking forward to what you turn up
there — and no rush on it.

Issue #130, opened 2026-07-20
In-place aliasing (input==output) silently returns wrong values in 8 functions; 5 are an unreleased regression from #103


mario4tier:
TA-Lib lets a caller pass the same buffer as an input and an output (`website/src/api/README.md:158,168`). **Rarely used**, but a documented guarantee — and silently broken in eight functions: they return `TA_SUCCESS` with correct `outBegIdx`/`outNBElement` and wrong values.

Five are an unreleased regression from 47d444d7 (#103).

## Measured

200 bars, `startIdx=0`, period 14 (STOCH/STOCHF 5/3/3), aliased run vs a separate-buffer reference, bitwise.

| function | aliased pair | wrong | |
|---|---|---|---|
| `LINEARREG` | out0 ← inReal | 186/187 | #103 |
| `LINEARREG_ANGLE` | out0 ← inReal | 186/187 | #103 |
| `LINEARREG_SLOPE` | out0 ← inReal | 186/187 | #103 |
| `LINEARREG_INTERCEPT` | out0 ← inReal | 186/187 | #103 |
| `TSF` | out0 ← inReal | 186/187 | #103 |
| `STOCH` | out1 ← inHigh | **out0** 190/192 | pre-existing |
| `STOCHF` | out1 ← inHigh | **out0** 192/194 | pre-existing |
| `MAVP` | out0 ← inReal | 160/171 | pre-existing |

`SMA` control: 0/187. At `47d444d7^`, `LINEARREG` and `TSF` are 0/187; STOCH and MAVP already fail. For STOCH/STOCHF, aliasing **D** onto a price input leaves D correct and corrupts **K**.

Clean: ACCBANDS, MINMAX, AROON, ULTOSC, CORREL, MAMA, TRIMA, KAMA, VAR/STDDEV, BBANDS (both paths), all 9 `ma()` types.

## Root causes

- **LINEARREG family + TSF** — the O(1) sliding recurrence reads the departing value (`linearreg.c:97`) after the priming write at `:88` overwrote it. The old O(period) rescan read the window before writing.
- **STOCH / STOCHF** — the scratch election (`stoch.c:147-153`, `stochf.c:145-150`) picks an output as scratch; when it aliases a price input, scratch writes destroy input the other output needs.
- **MAVP** — no defence when `outReal == inReal`, though `inPeriods` gets it at `mavp.c:85-92`.

## For whoever gates this

[...]

PR #154, opened 2026-07-31
perf(ultosc): evaluate each bar's true range once, not four times

dexhunter:
## What this changes

`ultosc.c`'s steady-state loop computes the per-bar true range and
close-minus-true-low pair **four times for every output bar**: once for the
entering bar, and once each for the three trailing bars leaving the 7/14/28
windows.

```c
while( today <= endIdx )
{
   /* bar `today`        */ trueLow = min(tempLT,tempCY); ... a1Total += ...; b1Total += ...;
   /* bar trailingIdx1   */ trueLow = min(tempLT,tempCY); ... a1Total -= ...; b1Total -= ...;
   /* bar trailingIdx2   */ trueLow = min(tempLT,tempCY); ... a2Total -= ...; b2Total -= ...;
   /* bar trailingIdx3   */ trueLow = min(tempLT,tempCY); ... a3Total -= ...; b3Total -= ...;
}
```

Both quantities depend only on `(inHigh[i], inLow[i], inClose[i],
inClose[i-1])`, so three of those four evaluations reproduce a value the loop
already produced a few hundred iterations earlier. The warm-up has the same
shape: three overlapping priming loops re-evaluate the bars inside the shorter
windows two and three times over.

This patch computes each bar once on entry, keeps the pair in a `CIRCBUF` sized
to the longest period, and has the three trailing subtractions read it back. The
warm-up becomes a single pass over the longest window that adds each bar to the
shorter windows' totals as it reaches them.

## Measurements, absolute and wall-clock

`ta_bench_direct --function=ULTOSC --points=100000 --iters=200`, GCC 13.3
x86-64. I rebuilt each arm from source and measured both in the same session.

| | ns per call (100,000 bars) | ns per bar |
|---|--:|--:|
| `dev` @ `cae5c3d` | 962,141 | 9.62 |
| this branch | 586,417 | 5.86 |
| | **−375,724 ns (−39.1%)** | **−3.76** |

A second harness agrees. It sweeps six `.text` layouts (nop padding ahead of the
archive plus varying `-falign-functions`), draws a fresh random-walk fixture for
every repetition, and times both arms in the same window. Its layout-swept
median ratio is **0.6166 and 0.6134** across two independent measurement blocks,
so both methods land near −38%.

This is a wall-clock number, not an instruction count. Your note on #147 says
Callgrind Ir moved a flat −8% on the last change in this family while wall time
ranged from −27% to +2%, so I measured accordingly. It is also one toolchain on
one machine, GCC x86-64. I have not measured MSVC or Apple clang and would not
want the number read as transferable.

`--points` sweep on the same box, rebuilding each arm and alternating, ns per
call (`dev` → branch): 20,000 → 191,594 → 119,872 (−37.4%); 100,000 → 962,141 →
586,417 (−39.1%); 200,000 → 1,920,054 → 1,199,955 (−37.5%). A repeat of the
branch arm gave 120,678 / 601,996 / 1,195,507. The ratio holds flat in `n`,
which is what a per-bar constant-factor change should do.

## Correctness

**Bit-identical by construction, not by tolerance.** The same expression
produces every value from the same operands, and every total accumulates the
same bars in the same ascending order. I reassociated nothing, carried no
running total across the series, and left the output expression alone. The
change is *when* a value gets computed, never *how*.

What I checked:

* `ta_regtest` passes in full.
* A differential harness compares raw IEEE bytes against a `dev` build over 10
  period vectors (including `1/1/1`, `28/14/7` reversed, `14/14/14`, `5/100/20`)
  × 7 price regimes (random walk, monotone up, monotone down, flat, alternating
  trend/chop, near-epsilon magnitudes, large magnitudes) × the guarded,
  unguarded and `TA_S_` float entry points, full range, two subranges, the
  single-bar `startIdx==endIdx` shape, the first computable bar, the empty
  range, in-place aliasing with `outReal == inClose` (#130), the streaming
  `Open`/`Peek`/`Update`/`Close` trajectory and `OpenAndFill`. Byte-identical
  throughout.
* Every timed repetition also compares a per-call output digest at the measured
  size (100,000 bars) against a `dev` build, so the timed path gets checked on
  exactly the data it is timed on.

Two scope notes, since both are easy to get wrong here:

* The ring is `CIRCBUF_PROLOG_CLASS(term, UltOscTerm, 32)`, so at the default
  7/14/28 it lives entirely on the stack and **adds no allocation**. It reaches
  the heap only when the longest period exceeds 32, `CIRCBUF_INIT_CLASS` already
  returns `TA_ALLOC_ERR` on failure, and `CIRCBUF_DESTROY` runs on the one path
  that can reach it.
* `ULTOSC` is `flags: [stream]` and the streaming tier derives from this same
  body. Generation still derives it, and the stream trajectory is part of the
  bit-exactness check above.

I left three further changes out on purpose. Rewriting the true range as
`max(high, prevClose) - min(low, prevClose)`, hoisting `prevClose` across the
loop, and inferring the two longer windows' zero-checks from the shortest one
each measured faster, and together they reach about −45% rather than −38%. But
each one rests on an argument I could not discharge for every input the API
accepts: that `high >= low` always holds, that the in-place aliasing read order
does not matter, and that a running sum carrying subtractions cannot round to
zero while a shorter one has not. This diff is the part that needs no such
argument. Happy to send any of them separately if you want them measured on
their own.

## Generated files

`ta_codegen/input/ultosc/ultosc.c` is the only file I edited by hand; the rest
is `cd ta_codegen/generator && cargo run -- generate`. The generated surfaces
lose 1,602 lines net, because each backend had unrolled the four true-range
blocks into every variant.

## Provenance

The `CIRCBUF` formulation came out of an automated optimization search over this
one file, run against a harness that gated on the bit-exactness matrix above.
The trajectory is public as a record of that search:
https://dashboard.weco.ai/share/x-g7CdO52PGLAE_gLtyWrCxmRE1w-p9H

The diff here is not the search's best-scoring step. It is a deliberately
narrower version of it, for the reason given above. I have read and measured the
change myself and take responsibility for it.

### DIFF
--- a/src/ta_func/ta_ULTOSC.c
+++ b/src/ta_func/ta_ULTOSC.c
@@ -55,6 +55,9 @@
  *  -------------------------------------------------------------------
  *  281206 DM   Initial Implementation
  *  010606 MF   Abstract local arrays. Detect divide by zero.
+ *  073126 DX   Evaluate each bar's true range and close-minus-true-low once
+ *              instead of four times, keeping them in a CIRCBUF sized to the
+ *              longest period. Same values, same summation order, bit-exact.
  */
 
 TA_LIB_API int TA_ULTOSC_Lookback( int optInTimePeriod1, int optInTimePeriod2, int optInTimePeriod3 )
@@ -112,12 +115,17 @@ TA_LIB_API TA_RetCode TA_ULTOSC( int    startIdx,
    int j;
    int today;
    int outIdx;
-   int trailingIdx1;
-   int trailingIdx2;
-   int trailingIdx3;
+   int trailingPos1;
+   int trailingPos2;
    int usedFlag[3];
    int periods[3];
    int sortedPeriods[3];
+   double local_term_closeMinusTrueLow[32];
+   double *term_closeMinusTrueLow;
+   double local_term_trueRange[32];
+   double *term_trueRange;
+   int term_Idx;
+   int maxIdx_term;
 
    if( startIdx < 0 )
       return TA_OUT_OF_RANGE_START_INDEX;
@@ -145,6 +153,14 @@ TA_LIB_API TA_RetCode TA_ULTOSC( int    startIdx,
    if( !outReal )
       return TA_BAD_PARAM;
 
+   /* The two per-bar terms the three moving sums are built from. Both are a
+    * pure function of the bar, so each bar is evaluated once on entry and read
+    * back when it leaves each of the three windows.
+    */
+   /* One entry per bar of the longest window. Stays on the stack for every
+    * period up to 32, which covers the 7/14/28 default.
+    */
+   /* Id, Type, Static Size */
    *outBegIdx= 0;
    *outNBElement= 0;
    /* Ensure that the time periods are ordered from shortest to longest.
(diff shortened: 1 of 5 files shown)


mario4tier:
Merged — thank you. Your commit is a merge parent (`0053e4af7`, plus `3744b97a9` picking up your amended head) and the CHANGELOG credits you.

**Your numbers reproduced before merging.** Independent interleaved medians on a second GCC x86-64 box: −39.9% / −41.1% / −40.4% at 20k/100k/200k points, flat in n as you reported. Bit-exactness confirmed with a 1052-case matrix hashing raw IEEE bytes — 7 price regimes × 10 period vectors (including >32 heap rings, `1/1/1`, equal periods) × full/sub/single/empty ranges × double, `TA_S_` float, and `outReal==inClose` aliased entry points, plus streaming trajectories — byte-identical throughout. Post-merge, the full suite passes 161 functions × 4 language servers.

**Two things fixed in the merge, for the record:**
1. The branch's generated Rust didn't compile: `trailingPos1 -= optInTimePeriod3;` renders as `usize -= i32` (8 sites). Not a defect in your change — your input C is correct; the generator's compound-assign cast gate consulted only a name heuristic, and `trailingPos*` is the first variable ever typed `usize` purely by subscript inference. Fixed generator-side in `fc7ab50dd` (+ follow-up hardening in `58dca3b6c`) with regression tests; the merge regenerated with the casts. For next time, `scripts/build.py servers && bin/ta_regtest --codegen --function=ULTOSC` (steps 3–4 in CLAUDE.md) builds the Rust/Java surfaces and would have flagged it immediately.
2. `ta_codegen/output/java/tools/TaCodegenServe.java` was stale — plain `generate` doesn't render the servers; same pipeline step as above.

Also folded in: house-style one-line change-history entry, and you're added to the contributors block in `ultosc.c`.

Review spin-off: #155 — your ring measured *faster* than the old allocation-free code even at 64 bars, so the Rust `Vec` allocations are a property question, not a performance one; that issue tracks making the Rust CIRCBUF rendering match C's stack-first hybrid.

On the three deferred optimizations: your call to keep them out was the right one. If you want to pursue any as separate PRs with the same measurement rigor, they're welcome — the in-place aliasing read-order one deserves the hardest look, since the `outReal==inClose` contract (#130) is exactly where it can go wrong.

PR #174, opened 2026-08-07
fix(ta_codegen): make the C bench tools build under MSVC

kevinlincg:
`ta_bench_cg` and `ta_bench_stream` have never compiled on MSVC. I hit this while producing the MSVC column for #147 and fixed it in the generator; the numbers in that issue exist because of this patch.

## What breaks

Three POSIX-only dependencies, none present in the MSVC CRT:

| | symptom |
|---|---|
| `clock_gettime` / `CLOCK_MONOTONIC` | `error C2065` — `get_nanotime()` has `__APPLE__` and POSIX branches, no `_WIN32` |
| `strtok_r` | `LNK2019` |
| `strcasestr` | `LNK2019` |

The first one is a hard parse error, so nothing downstream even gets a chance.

This leaves MSVC with no way to run the in-tree benchmarks — the toolchain where a plain-C codegen change is least likely to be caught by anything else.

## The fix

All three in `bench_gen.rs`, so both generated tools inherit it.

**Timing.** A `_WIN32` branch on `QueryPerformanceCounter`. The frequency is fixed for the process lifetime, so it is queried once. Ticks are converted by splitting into whole seconds plus a remainder:

```c
return (long long)((now.QuadPart / freq.QuadPart) * 1000000000LL
     + ((now.QuadPart % freq.QuadPart) * 1000000000LL) / freq.QuadPart);
```

`(t / f) * 1e9` on its own truncates to whole seconds; `t * 1e9` overflows a signed 64-bit at ~9.2e9 ticks, which is about an hour of uptime at a 10 MHz QPC. The split avoids both.

**`strtok_r`** → `strtok_s` behind a `bench_strtok_r` macro. Same reentrancy contract, same argument order.

**`strcasestr`** → a local `bench_strcasestr`. It is a short loop and every alternative is another platform extension.

## Verification

- **MSVC Build Tools 2022 x64, pristine `dev` clone at `755800275`**: patch applies clean, both binaries compile, both run. `ta_bench_cg --points=2000 --iters=5 --function=MAX` returns sane timings.
- **macOS clang regression**: both still compile with no new warnings and produce unchanged output.
- Generated sources are committed alongside the generator change, so `generate-bench --backend=c` is a no-op on this branch.

## One measurement note

Worth knowing before anyone quotes Windows numbers: QPC is typically 10 MHz, so `ta_bench_stream`'s `batch_last_ns` at `--iters=50` quantizes to 2 ns steps — 3–10 ticks per measurement. In my first MSVC run that read as one function *improving* 0.70x and another regressing 2.50x, both pure artefacts; at `--iters=5000` the same comparison moved to 0.33x and the baseline itself went from 20.000 to 45.000 ns. A low-iteration Windows run measures the timer, not the code.

### DIFF
--- a/ta_codegen/generator/src/bench_gen.rs
+++ b/ta_codegen/generator/src/bench_gen.rs
@@ -22,8 +22,8 @@ pub fn generate_c_bench(funcs: &[FuncDef]) -> String {
     s.push_str(" * Output: FUNCNAME timing_ns (one per line)\n");
     s.push_str(" */\n");
     s.push_str("#include <stdio.h>\n#include <stdlib.h>\n#include <string.h>\n");
-    s.push_str("#include <math.h>\n#include <time.h>\n");
-    s.push_str("#ifdef __APPLE__\n#include <mach/mach_time.h>\n#endif\n\n");
+    s.push_str("#include <math.h>\n#include <time.h>\n#include <ctype.h>\n");
+    s.push_str("#ifdef _WIN32\n#include <windows.h>\n#endif\n#ifdef __APPLE__\n#include <mach/mach_time.h>\n#endif\n\n");
     // The shared benchmark input corpus (src/tools/ta_bench is on the include
     // path — see the ta_bench_cg / ta_bench_stream gcc invocations in main.rs).
     s.push_str("#include \"bench_corpus.h\"\n\n");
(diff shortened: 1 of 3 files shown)


mario4tier:
Merged. Thanks for the contribution — and for the QPC quantization note; that saves the next person a misread Windows run.

PR #175, opened 2026-08-07
fix(ta_codegen): stop emitting the duplicated NULL-check blocks

### LINKED ISSUE #169: ta_codegen: stream bodies emit duplicated NULL-check blocks (22 sites)
The generated stream bodies contain 22 pairs of byte-identical NULL-check blocks:

```c
adx = malloc((endIdx - startIdx + optInTimePeriod) * sizeof(double));
if( !adx )
{
   TA_ADX_Close( sub0 ); TA_Free( sc_outReal );
   return TA_ALLOC_ERR;
}
if( !adx )                      /* <- identical, unreachable */
{
   TA_ADX_Close( sub0 ); TA_Free( sc_outReal );
   return TA_ALLOC_ERR;
}
```
`src/ta_func/ta_ADXR.c:340`, and 21 more across the stream bodies of the
functions with allocating intermediates.

Dead code, no runtime effect. Long-standing — present on `dev` well before #166;
`git show origin/dev:src/ta_func/ta_ADXR.c` has it.

**Cause.** Two mechanisms emit the same check.
`malloc_null_check_block` (`ta_codegen/generator/src/backends/c_stream.rs:1263`)
is injected unconditionally by the `Statement::Assign` / `Statement::VarDecl`
arms for every allocating intermediate; the explicit check in the input `.c`
(e.g. `ta_codegen/input/stoch/stoch.c`, `ta_codegen/input/adxr/adxr.c`) is then
transcribed as a second one, its `Return` remapped to the same cleanup form.

**Why it is not a one-liner.** Two obvious fixes were tried and neither works:

1. *Skip the injection when the source already checks.* Breaks BBANDS. The
   injected block carries a cascading `free()` of every intermediate allocated
   before it — `free( tempBuffer1 )` in `ta_BBANDS.c:797` — which the source
   check does not. The injected block is the STRONGER of the two, so suppressing
   it loses OOM cleanup. Caught by
   `test_c_bbands_open_frees_prior_intermediate_on_oom`.
2. *Drop the transcribed source check instead.* Compiles, all tests pass, clippy
   clean — and changes nothing: 22 duplicates before, 22 after. The statement
   shapes that reach the transform are not the ones the drop-arm matches.

So the fix has to reconcile the two blocks rather than pick one: keep the
injected cleanup semantics while suppressing the duplicate emission, most likely
by having the transform recognise the source check and *rewrite* it (adding the
cascading `free()`s) rather than emit alongside it.

Worth doing because the injection is not idempotent: the next NULL check added
to an input `.c` for a stream intermediate doubles again.

kevinlincg:
Closes #169.

## Why the two obvious fixes both fail

You recorded both in the issue, and they hold up:

1. **Skip the injection when the source already checks** — loses the cascading `free()`. The injected block frees every intermediate allocated before it; the source check does not. `test_c_bbands_open_frees_prior_intermediate_on_oom` catches it.
2. **Drop the transcribed source check** — changes nothing, 22 before and 22 after.

The reason (2) does nothing is structural rather than a matter of matching the right shape: `fs` is a **1:1 statement map**. It is handed one `Statement` at a time and returns one back. When it is holding the source's `if( !tempBuffer )`, it has no way to know whether the statement *before* it was an injection for that same variable — that context does not exist at that point in the fold.

## The fix

Reconcile one level up instead. After `rewrite_stmts` has produced the statement list, a peephole walks it: an `if( !x )` whose **immediately preceding sibling** is an injected block ending in `if( !x )` is dropped. What survives is the injected block, with the stronger cleanup — so the OOM semantics that (1) would have destroyed are preserved by construction, not by care.

It recurses through nested bodies (`If` / `While` / `DoWhile` / `For` / `ForC` / `Switch` / `Block`) since the composed region can be arbitrarily nested.

`ta_BBANDS.c`, before and after — the block that survives is the one that was already there:

```c
  if( !tempBuffer2 )
  {
     free( tempBuffer1 ); TA_MA_Close( sub0 ); TA_STDDEV_Close( sub1 );
     TA_Free( sc_outRealUpperBand ); TA_Free( sc_outRealMiddleBand ); TA_Free( sc_outRealLowerBand );
     return TA_ALLOC_ERR;
  }
- if( !tempBuffer2 )          /* the weaker copy, unreachable */
- {
-    free(tempBuffer1);
-    return TA_ALLOC_ERR;
- }
```

## Effect

22 duplicate pairs across 9 functions → 0: `ADXR`, `APO`, `BBANDS`, `MACDEXT`, `PPO`, `PVO`, `STOCH`, `STOCHF`, `STOCHRSI`.

Your idempotence point holds after this too — the peephole keys off the pairing, not off a fixed list, so a NULL check added to a new input `.c` for a stream intermediate collapses the same way instead of doubling.

## Verification

```
test_c_bbands_open_frees_prior_intermediate_on_oom   ok    (the gate that caught attempt 1)
generator suite      322 + 255 + 14 + 11 + 10 + 2 passed, 0 failed
ta_regtest           all green
variant parity       168 functions x 37,248 vectors, bit-identical
C library            0 errors
```

Duplicate count measured directly on the generated sources with a regex over consecutive identical `if( !x ){...}` pairs: 22 on `dev`, 0 here.

### DIFF
--- a/src/ta_func/ta_ADXR.c
+++ b/src/ta_func/ta_ADXR.c
@@ -342,11 +342,6 @@ TA_RetCode TA_ADXR_OpenInternal( struct TA_ADXR_Stream **stream, const double in
          TA_ADX_Close( sub0 ); TA_Free( sc_outReal );
          return TA_ALLOC_ERR;
       }
-      if( !adx )
-      {
-         TA_ADX_Close( sub0 ); TA_Free( sc_outReal );
-         return TA_ALLOC_ERR;
-      }
       /* Compute ADX over a range that starts (period-1) bars earlier, so each
        * ADXR bar can pair the current ADX with the ADX from (period-1) bars ago.
        */
@@ -480,11 +475,6 @@ TA_LIB_API TA_RetCode TA_ADXR_OpenAndFill( TA_ADXR_Stream **stream, const double
          return TA_BAD_PARAM;
       }
       adx = malloc((endIdx - startIdx + optInTimePeriod) * sizeof(double));
-      if( !adx )
-      {
-         TA_ADX_Close( sub0 ); TA_Free( sc_outReal );
-         return TA_ALLOC_ERR;
-      }
       if( !adx )
       {
          TA_ADX_Close( sub0 ); TA_Free( sc_outReal );
--- a/src/ta_func/ta_APO.c
+++ b/src/ta_func/ta_APO.c
@@ -320,11 +320,6 @@ TA_RetCode TA_APO_OpenInternal( struct TA_APO_Stream **stream, const double inRe
          TA_MA_Close( sub0 ); TA_MA_Close( sub1 ); TA_Free( sc_outReal );
          return TA_ALLOC_ERR;
       }
-      if( !tempBuffer )
-      {
-         TA_MA_Close( sub0 ); TA_MA_Close( sub1 ); TA_Free( sc_outReal );
-         return TA_ALLOC_ERR;
-      }
       /* Make sure slow is really slower than
        * the fast period! if not, swap...
        */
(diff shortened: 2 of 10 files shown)


mario4tier:
Thank you for this — and especially for the diagnosis behind it, which is the part that mattered most. Establishing that the *injected* block is the stronger of the two (it frees every intermediate allocated before it, which the source check does not) is what determined the shape of the answer, and `test_c_bbands_open_frees_prior_intermediate_on_oom` is now in the tree guarding exactly that. Both carried forward unchanged.

We ended up landing a different implementation on `dev` (d1deb92f2), so I'm closing this one. The short version: the peephole keys on the two blocks being adjacent siblings with identical conditions, and two ordinary edits defeat that — a comment on the source check (`Statement::Comment` lands between the injected block and the check), and the `x == NULL` spelling that `mavp.c` already uses. Both bring the duplicates back silently. The version on `dev` keys instead on a proof recorded by the injection itself, which holds in both cases.

More detail in #169. Your analysis is credited in the commit — genuinely appreciated, and please do keep them coming.

Issue #112, opened 2026-07-13
IMI: successful call emits NaN on an all-flat window (0/0) — guard the divide, return 50.0


mario4tier:
## Bug

`IMI` computes `100 * upsum/(upsum + downsum)`. For an all-flat window — every bar
`close == open` (a halted/illiquid instrument, a constant series, FUZZ_CONSTANT) —
`upsum == downsum == 0` → `0/0` → **NaN**. A *successful* TA call must never emit
NaN (NaN as input is separately undefined behavior). Present in **v0.6.4 and the
current shipped code** — a latent invariant violation, independent of any perf work.

## Fix

Guard the divide: when `upsum + downsum` is zero, return **50.0** — IMI's neutral
center. IMI is a 0–100 oscillator; equal up/down → 50, and *no* movement → no
up/down bias → 50 is the least-surprising value. (Contrast CCI #7, which returns
`0.0` for its degenerate case because CCI is centered at 0 — the right degenerate
value is each indicator's own neutral point.)

Numerics-changing vs 0.6.4 (NaN → 50.0), so it needs a latest→0.6.4 tolerance-
manifest entry. **Manifest wrinkle:** the divergence is `NaN(0.6.4)` vs `50.0`, not
a small numeric delta, so the fuzz-064 classifier needs an explicit "old==NaN &&
new finite-defined" tolerated case for IMI (today it would score NaN-vs-finite as a
real failure). CCI #7 is the precedent for the manifest mechanism; this adds the
NaN-specific comparison.

## Follow-up (broader invariant)

Audit the library for other unguarded `x/0` → NaN paths under the same "a
successful call never emits NaN" invariant — candidate ratio families: RSI, STOCH,
MFI, ULTOSC. Likely a separate sweep once IMI sets the pattern.

Separate from #110 (pure speed).

mario4tier:
This is now fixed and will be part of release 0.8.1

Issue #193, opened 2026-08-12
perf(streaming): reduce integer division per bar in Update/Peek


mario4tier:
This issue is a up to ~40% speed optimization, not a fix or numerical change.
-----
The generator has two ring-buffer lowerings for the streaming tier. 85 functions get the cheap increment-and-wrap `ringPos_` form. 14 get the absolute-index form — needed because a rolling-extremum rescan indexes arbitrary positions inside the window, not just the trailing element — and that one maps **every** access through `%`:

```c
static void TA_WILLR_StepInternal( struct TA_WILLR_Stream *sp, ... )
{
   sp->x_inHigh[sp->today % sp->xCap] = inHigh;
   ...
   while( ++sp->i <= sp->today )
   {
      tmp = sp->x_inLow[sp->i % sp->xCap];   /* per element of the rescan */
```

`xCap` is a runtime field, so these are real divisions. `TA_WILLR_StepInternal` is 159 instructions containing **6 `idiv`**, on a body that otherwise does a few compares and a subtract.

Affected: `AROON AROONOSC MAX MAXINDEX MIDPOINT MIDPRICE MIN MININDEX MINMAX MINMAXINDEX STOCH STOCHF VAR WILLR`. Batch is unaffected — it indexes the caller's arrays linearly and contains no `%` at all.

Prototype: branch [`bench-modulo-ring`](https://github.com/TA-Lib/ta-lib/tree/bench-modulo-ring) (`586c9e5c2`).

### Fix

Round the extrema ring to the next power of two in `Open`, and map slots with a mask. `idx & xMask` equals `idx % xPhys` and stays injective over any `xCap` consecutive bars, so the automaton's semantics are untouched.

Add a slot-mapping hook to the `NameMap` trait, defaulting to today's behaviour so other backends are unaffected:

```rust
/// Map an absolute bar index onto its extrema-ring slot. The default is the
/// modulo of the logical capacity; a backend whose Open rounds the ring to a
/// power of two overrides this with the equivalent mask.
fn extrema_slot(&self, idx: Expr) -> Expr {
    Expr::BinOp(Box::new(idx), BinOp::Mod, Box::new(Expr::Var(self.extrema_cap())))
}
```

override it in the C backend with `BinOp::BitwiseAnd` over a new `sp->xMask`, and in `Open`:

```c
sp->xPhys = 1;
while( sp->xPhys < sp->xCap ) sp->xPhys <<= 1;
[...]

Issue #192, opened 2026-08-11
Composed streaming Open recomputes every sub-call, and STDDEV's sqrt map cannot vectorize


mario4tier:
> [!IMPORTANT]
> **This issue was reframed.** It opened as a `PRAGMA TA_ALT` mechanism cleanup, and
> the investigation it asked for found something larger underneath: **every composed
> function's `Open` was recomputing its sub-calls**, and half of STDDEV's batch cost
> was a `sqrt` loop that could not vectorize. That is what the issue now tracks —
> see [this comment](https://github.com/TA-Lib/ta-lib/issues/192#issuecomment-5273846095)
> for the measurements, the two fixes, and what is left.
>
> **The original request below is preserved for context, not as pending work.** It is
> untouched, and by its own description it is a mechanism cleanup whose stream output
> is byte-identical — no user-visible effect either way. It is not what keeps this
> issue open; the unfinished work is the Rust and Java backends.

---

## The remaining batch/stream split that is not a `PRAGMA TA_ALT`

#190 gave the input tree a first-class way to say "the batch and streaming tiers run
different algorithms": declare `<name>_ALT<n>` under a `PRAGMA TA_ALT={<api>,<lang>}`
decoration. The six rolling-extremum functions migrated to it and
`analyze_fastpath_skip` — the `if( <param> <= <literal> )` idiom it replaced — was
deleted.

One mechanism doing the same job survives, in different syntax and under a
confusingly similar name: **`is_fastpath_block`** (`generator/src/streaming.rs`,
used from `analyze_composed`). It strips a leading parameter-guarded, early-returning
block out of the Composed tier's region, which is how two functions keep a batch-only
specialization out of their stream:

- **BBANDS** — the fused SMA+STDDEV path, guarded by `optInMAType == TA_MAType_SMA`
  (`ta_codegen/input/bbands/bbands.c`).
- **MACDEXT** — the all-EMA shortcut that delegates to `TA_MACD`
  (`ta_codegen/input/macdext/macdext.c`).

So the tree has two ways to express one concept. Folding these into `PRAGMA TA_ALT`
would leave one, and would let `is_fastpath_block` and its strip filter be deleted the
way `analyze_fastpath_skip` was.

### What the migration would look like

The construct absorbs the enum guard without extension, because an alternate is a
whole body rather than an arm of a recognized branch:

```
base   bbands          the general MA path — what Composed already streams
[...]

Issue #184, opened 2026-08-10
Low priority: the period-1 identity arm is hand-copied into ten input files the generator already parses


mario4tier:
Low priority — a refactor to revisit, not a defect.

The "a period of 1 performs no smoothing, so the output is a copy of the input"
arm is now **hand-copied into ten `ta_codegen/input/` files**, in three different
shapes, and the generator already knows how to parse it.

## Where it lives today

| file | shape |
|---|---|
| `sma`, `wma`, `trima` | *(no arm — the window math is naturally exact at 1)* |
| `wma.c` | element loop, `*outNBElement` set up front |
| `rsi.c`, `cmo.c` | counted `for`, `*outNBElement` up front |
| `t3.c`, `kama.c`, `vwma.c`, `hma.c`, `ema.c`, `dema.c` | `today`-cursor `while` |
| `tema.c` | mutates `startIdx` directly |
| `ma.c` | pre-switch dispatch guard (`period == 1 \|\| maType == DISABLED`) |
| `macd.c`, `macdfix.c` | *per-step* arm inside the loop, not an early return (the signal EMA only) |

`streaming::detect_identity_path` (`ta_codegen/generator/src/streaming.rs`)
already parses this exact shape to drive every stream surface — `Open`,
`OpenAndFill`, the transition step — so the generator **recognises** the arm but
never **emits or requires** one.

## Why it is worth revisiting

The cost has materialised repeatedly. Each function has to remember to hand-copy
the block, and several did not:

- `HMA` rejected period 1 outright (fixed, `012585d00`).
- `VWMA` computed `(P*V)/V` (fixed, `9325679e3`).
- `EMA` and `DEMA` ran `(x-prev)+prev` — **while `ema.md`, `dema.md` and the
  0.6.5 CHANGELOG all promised the copy** (fixed, `0773ecb98`).
- `MACD`/`MACDFIX` ran the unguarded signal recursion at `signalPeriod == 1`
  (fixed, `fb54b1235`).

Four omissions of the same contract, each found only by going looking.

## Sketch of the options

1. **Declarative**: a YAML flag (e.g. `period1: identity`) that makes the
   generator emit the arm into every surface — batch, `TA_S_`, and the three
   stream tiers. Removes the hand-copies entirely and makes the contract
   impossible to forget.
2. **Gate only**: keep the hand-written arms but have `generate` FAIL when a
   function is in the MAType enum (or otherwise declares the contract) and no
[...]

PR #194, opened 2026-08-12
feat(marketfi): Market Facilitation Index

kevinlincg:
Implements `MARKETFI` per proposal TA-Lib/ta-lib#230.

    MARKETFI = ( High - Low ) / Volume

Stateless and per-bar — no seeding, smoothing, accumulator or unstable period,
lookback 0, no `start_dependent`.

### Notes on the two decisions the proposal left open

**Zero volume reports 0 rather than dividing.** Both references divide anyway —
Tulip emits ±Inf, and NaN when the range is zero too — but #112 settled that a
successful call never emits either, and an untraded bar facilitated no movement.
The test is an exact `!= 0.0` rather than `TA_IS_ZERO`, whose 1e-14 band is
absolute and meaningless against an unbounded volume scale; same reasoning as the
`prevClose` guard in `input/nvi/nvi.c`.

**Named MARKETFI, not MFI.** `TA_MFI` is already the Money Flow Index. MARKETFI
is the name Tulip and pandas-ta-classic use.

### New price bundle: HLV

MARKETFI is the first function wanting a range against volume with no close.
Adding it surfaced that `ta_def_ui.h`'s extern list was a second hand-kept copy
of `PRICE_INPUT_CONSTANTS`: the definition appeared, the declaration did not, and
the build failed in `table_m.c` with an undeclared identifier pointing nowhere
near the cause. `gen_def_ui_h()` now derives that list from the same table, so
the two cannot fall out of step again.

### Tests

`src/tools/ta_regtest/ta_test_func/test_marketfi.c`, four legs — pinned values
(6 samples recomputed at full precision, plus an assertion that the full range
comes back), zero volume (real range and 0/0), aliasing (`outReal` over each of
the three inputs), range independence (bitwise).

**The pinned leg is not an external oracle and is no longer labelled as one.**
Those six values are this definition evaluated by hand against `test_data.c`,
so they constrain the implementation, not the formula — both sides are the same
definition. No usable external vector exists for MARKETFI: Tulip ships one at
`tests/untest.txt:248`, but its expected column is `{0.000, 0.000, …}` for all
15 bars, because the values are ~1e-7 and that file prints three decimals. That
vector passes for any implementation returning approximately nothing. What does
constrain the formula here is the mutation set below, which a wrong formula
cannot survive.

Deliberately **not** `CHECK_EXPECTED_VALUE`: it compares against an absolute 0.01
band and takes no tolerance argument, and these values are ~1e-7. A constant 0.0,
a sign flip, a dropped divisor, an off-by-one bar and the wrong component pulled
from the new HLV bundle all pass that band. Each of the five was run against the
relative check and each is rejected.

### Verified

`scripts/regtest.py` green on all four servers · variant parity 169 functions ×
42,912 vectors bit-identical · stream verify 169 functions, 19,252 legs.
MARKETFI joins the post-cutover set with no frozen-reference baseline, still
bitwise-gated by VARIANT. Guard removal and a sign flip were both sabotage-checked
against the new tests.

### Not included

The `ta_tulip_serve` / `ta_pandas_serve` oracle arms the proposal asks for —
those servers are not in this tree. Flagged when claiming the proposal.


kevinlincg:

kevinlincg:

- **`camel_case` removed from `marketfi.yaml`.** Same dead key, retired by #179. Worth noting it would now be a hard failure rather than a silent discard: `47aa52ae3` turns on `deny_unknown_fields`, so this branch as it stood an hour ago would fail to generate on current `dev`. Removing it and regenerating on `71c88e046` is the check that it is actually gone — `cargo run -- generate` succeeds and leaves zero drift.

kevinlincg:

kevinlincg:

kevinlincg:
Audited this branch against the defect you fixed in `0edfbe874`, and it had the same shape in a milder form. Fixed and force-pushed.

**The leg called `ORACLE` here was never an external oracle.** Its six values are this definition evaluated by hand against `test_data.c` — honestly described in the file as "recomputed from test_data.c", but sitting under a heading that implied third-party corroboration it never had. Renamed to `PINNED VALUES` throughout, with the limitation stated where a reader meets it: they constrain the implementation, not the formula, because both sides are the same definition.

**There is no usable external vector for MARKETFI, and that is worth recording rather than papering over.** Tulip ships one — `tests/untest.txt:248`, 15 bars — and its expected column is `{0.000, 0.000, …}` for every bar. MARKETFI's values are ~1e-7 and that file prints three decimals, so the published vector passes for any implementation that returns approximately nothing. I would have had to either invent precision it does not have or cite it as corroboration it cannot give.

What actually constrains the formula in this PR is the mutation set: constant 0.0, sign flip, dropped divisor, off-by-one bar, wrong HLV component — each run and each rejected.

Also checked the citations, since `efi.md` cited a TradingView `ta.fi(length)` that does not exist: `ti_marketfi` is in Tulip and `volume/marketfi.py` is in pandas-ta-classic, both verified by fetching them rather than by trusting the proposal.

The PR description is updated too, not just this comment — leaving a body that states something withdrawn is the other half of what went wrong on #197.

Full `scripts/regtest.py` green on all four servers after the change: `Stream verify: 170 functions, 17084 legs bit-exact vs batch`.

kevinlincg:
Rebased onto `92b3a9ce7`.

**That commit is a defect I shipped, and it exposes a gap in how I was verifying these.** EFI landed with a fusing body and no entry in the FMA drift-detector inventory, so `rust_fma_dispatch_fires_for_exactly_the_fusing_functions` has been red on `dev` since the merge. My routine was `scripts/regtest.py --no-perftest` across the four servers plus a targeted `cargo test --test period1_suite` when #184 landed — and `regtest.py` does not run the generator's own Rust suites, so a stale inventory literal there was invisible to everything I ran.

Fixed going forward: the full `cargo test --release` in `ta_codegen/generator` is now part of what I run before pushing, not a suite I reach for when a specific issue prompts it.

**Verified on this branch rather than assumed**: all eleven generator suites pass — 744 tests, 0 failed, including `backend_suite` where that inventory test lives. MARKETFI needs no entry: `(High - Low) / Volume` is a subtract and a divide, with no multiply-add for the dispatcher to fuse. That is the test agreeing, not me reasoning about it.

Also still green: full `scripts/regtest.py` on all four servers, `Stream verify: 172 functions, 17272 legs bit-exact vs batch`, and zero drift under both `generate` and `generate-servers`.

kevinlincg:
Rebased onto `14c6810f7`, and applied the half of that commit that also lands here.

Your QSTICK leg (4) fix — bitwise `+0.0` instead of `!= 0.0` — applies verbatim to MARKETFI's zero-volume leg, which had the same `out[1] != 0.0` shape. Changed to a `memcmp` against `+0.0`, with the reason at the call site: `!= 0.0` is true for `-0.0` too, so it would accept a guard that wrote the wrong zero. Issue #147's class, which I should have carried across from my own work on it.

**Getting the sabotage right took two attempts, and the first one would have been a false report.** Editing `input/marketfi/marketfi.c` to write `-0.0` and regenerating produced a binary the test still passed — which reads as "the check has no discriminating power". It is not: `grep '= -0.0;' src/ta_func/ta_MARKETFI.c` returns 0, because the generator normalises the literal back to `0.0` on the way through. The mutation never existed. Patching the *generated* C directly does reach it, and the leg then fails loudly:

```
Fail: TA_MARKETFI zero-volume out[1]=-0 out[3]=-0, expected exactly +0.0
      (a -0.0 here would pass a != 0.0 check)
```

Green again on restore. Worth recording that `-0.0` is not expressible through the input tier at all, since it makes that the wrong layer to mutate when testing a signed-zero contract.

On leg (3)'s guard, the other half of your commit: MARKETFI has no equivalent. Its pinned leg asserts the output length rather than a property of the input's floating-point regime, so there is no analogue of the `(double)(float)x != x` claim to correct here.

Verified on the new base: 744 generator tests pass, `scripts/regtest.py` green on all four servers, `Stream verify: 172 functions, 17272 legs bit-exact vs batch`, zero drift under `generate` and `generate-servers`.

kevinlincg:
Closing: this PR's content was merged in `ceb5747ab`. I force-pushed a rebase afterwards without noticing the merge had happened, which left it open with a head that is not an ancestor of `dev` — the confusion is mine.

The one change on that head which is genuinely missing from `dev` — pinning the zero-volume guard to `+0.0` bitwise instead of `!= 0.0`, the issue #147 class you fixed in QSTICK — is now #209 on its own, rebased onto your `717453336`.

I also owe a correction here rather than only in the new PR: my claim that no usable external vector exists for MARKETFI was wrong. Tulip's published test file really is vacuous at three decimals, but I turned "I cannot run the oracle servers on this box" into "no external anchor exists", having named those servers in my own first description of this PR. Your live capture from both, bitwise-agreeing across the full corpus, is what I foreclosed. The same shape applies to the zero-volume guard: I verified it through the C batch entry point and called the contract covered, when streaming and the language servers had never seen an exact zero.

Issue #195, opened 2026-08-12
ta_codegen: a new function's yaml hint/suggested is silently discarded when the parameter folds into a pre-defined constant


kevinlincg:
Adding two functions with an `optInTimePeriod` cost me two full `scripts/regtest.py` cycles each to a failure that reads as though the language servers are wrong, when the yaml was.

## What happens

`match_predefined_opt_input()` (`ta_codegen/generator/src/backends/ta_abstract_c.rs:1277`) keys on **name, range and default only**:

```rust
if opt.name == "optInTimePeriod" && max_i == 100_000 {
    if min_i == 1 {
        return match def_i {
            30 => Some("TA_DEF_UI_TimePeriod_30".into()),
            ...
            10 => Some("TA_DEF_UI_TimePeriod_10".into()),
```

When it matches, the C table references the shared constant, whose `hint` is hardcoded `"Time period"` (`:505`). The yaml's `hint` is discarded without a word.

The `suggested` triple is discarded on **both** paths, matched or not: a bespoke constant still points at the shared range struct, and `TA_DEF_TimePeriod_Positive` hardcodes `1, 200, 1` (`:444`).

The Rust / Java / C# metadata registries are generated from the yaml, so they echo what the yaml says. The two then disagree.

## What it looks like

`cargo run -- generate` succeeds. The C library builds. All four servers build. `ta_regtest` passes every function test. The failure arrives at the very end of the pipeline:

```
Testing Abstract metadata parity (Rust server vs C)
  ABSTRACT ERROR [QSTICK]: TA_GetOptInputParameterInfo[0] hint C="Time period" server="Number of period"
  Abstract metadata parity: 169 functions checked, 1 failed
Failed: Abstract interface Tests (error number = 1202)
```

and on the next run, after fixing the hint:

```
  ABSTRACT ERROR [EFI]: TA_GetOptInputParameterInfo[0] suggested C=[1,200,1] server=[2,200,1]
```

The gate is doing its job — it caught both. The problem is the distance: the message names the server, the cause is one line of yaml, and finding out costs a full four-server rebuild each time. It is also two separate lessons, since `hint` and `suggested` fail on different code paths (`EFI` defaults to 13, matches no pre-defined constant, and *still* lost its `suggested`).

## Suggested fix

[...]

Issue #179, opened 2026-08-08
Rust: remaining loose ends before the first crates.io publish


mario4tier:
The `ta-lib` crate (`ta_codegen/output/rust/library/`, v0.8.1) builds, packages and passes
every gate it has. This is the single checklist of what is still open before publishing —
the Rust analogue of #172.

## What was verified first (so the list below is scoped correctly)

**Does anything test the code we would publish? Yes — unlike Java.** The Rust JSON-RPC
server links the shipped crate rather than embedding a copy of it
(`tools/Cargo.toml:14 ta-lib = { path = "../library" }`; `tools/src/bin/ta_codegen_serve.rs:7
use ta_lib::{Core, ...}`; zero inlined indicator bodies against 1514 `core.*` calls). So
`--codegen`, `--xlang-hash`, `server_verify` and `stream_verify` all exercise the crate
source itself. Contrast `project_java_server_embeds_core`. **The untested thing is
packaging, not code.**

**Can it be published today?** Mechanically, yes. `cargo package` succeeds for both crates;
the `ta-lib` tarball is ~858 KB gzipped (limit 10 MiB). Both names are already owned
(`mario4tier` + team `github:ta-lib:rust-crate-io-owners`), and `ta-lib-dispatch` 0.1.1 is
already on crates.io (2026-08-03) byte-identical to `dispatch/src/lib.rs`, so the `=0.1.1`
pin resolves.

**Is the declared MSRV honest?** Yes, and it is not arbitrary: `cargo +1.86 check -p ta-lib`
passes, `+1.85` is rejected by cargo's own gate, and 1.86 is exactly where safe
`#[target_feature]` fns (RFC 2396) stabilized — which is what the 26 FMA clones are built
on. It is true but unguarded.

**Is aarch64 covered?** Yes — dev-nightly `arm64-linux` (`dev-nightly-tests.yml:517`) builds
and value-tests the Rust server on `ubuntu-24.04-arm`, exercising the
`#[cfg(not(target_arch = "x86_64"))]` arm in all 26 dispatch files.

So the code is covered. What is not: **packaging, the publish procedure, and ~12 public-API
shapes that freeze the moment 0.8.1 lands.**

---

## Status audit — 2026-08-31

Re-checked every item against `dev` `bbe63f18e`, after the two-day run of PRs #288–#319.
**Twelve items ticked, five split, one added.** What is left is small and almost entirely
yours.

[...]

Issue #24, opened 2024-01-20
Renko for Candlesticks


mrjbq7:
A request was made for "Renko for Candlesticks"

https://github.com/TA-Lib/ta-lib-python/issues/362

mario4tier:
This is now tracked on the [New TA Functions Board]( https://github.com/orgs/TA-Lib/projects/1 )

PR #196, opened 2026-08-12
feat(qstick): Qstick

kevinlincg:
Implements `QSTICK` per proposal TA-Lib/ta-lib#226.

    body[i]   = close[i] - open[i]
    QSTICK[i] = SMA( body, optInTimePeriod )

`input/qstick/qstick.c` is `input/sma/sma.c` with `inReal[x]` replaced by `(inClose[x] - inOpen[x])` and deliberately nothing else — same running-sum order, same read-before-write of the trailing term, same divide. That is what makes the differential a `memcmp` rather than a tolerance.

### The two decisions the proposal left open

**Default period 10**, matching Tulip and pandas-ta-classic. TraderEvolution documents 1; AmiBroker community code commonly cites 8 as "what Chande suggests", which I could not verify against a primary source. Both alternatives are recorded in `qstick.md`.

**SMA only, no `optInMAType`.** Chande's definition is an SMA and only the SMA path has any oracle. APO/PPO/PVO cut the other way, so this is worth an explicit ruling — adding the parameter later is an API break.

**Divide, not reciprocal multiply.** Tulip's `qstick.c` finishes with `sum * (1.0/period)`, up to one ULP away from `TA_SMA`. Dividing buys the bit-exact differential. The Tulip vector passes either way; a switch to the multiply would keep passing it and break the differential — the two legs constrain different things on purpose.

### Tests

Five legs in `test_composite.c`, as the second member after PVO:

- **Differential**, bit-exact against `TA_SUB` → `TA_SMA`, over periods {1,2,3,5,8,10,14,30,100} × startIdx {0,1,40,200}. The non-zero startIdx values are what make a seeding-loop off-by-one visible; at startIdx 0 the seed and output windows begin at the same place.
- **External oracle**, two published vectors. Achelis p.280 (`tests/atoz.txt:187`) is pinned **exactly**, not at a tolerance: every price on that page is a sixteenth and the divisor 4 is a power of two, so the vector is computed without a single rounding, and the book's printed four decimals are what those exact binary values round to. Tulip `untest.txt:327` is pinned at 1e-12 relative.
- **Period 1**, bit-exact against the raw body, with an assertion that the corpus still holds bodies that are not exactly representable — so the leg cannot quietly go vacuous.
- **Flat input**, exactly `0.0`. Issue #112 is structural here; the only divisor is the period.
- **In-place aliasing**, `outReal` over `inOpen` and over `inClose` separately, at periods 1, 10 and 30.

Sabotage-checked, all four rejected and green after restore: open/close swapped, the divide turned into a reciprocal multiply (a 1-ULP change), the seeding loop off by one, the lookback short by one.

### One thing worth knowing

The yaml's `hint` and `suggested` are pinned to `"Time period"` and `[1,200,1]` because the C abstract table folds this parameter into the shared `TA_DEF_UI_TimePeriod_10` constant while the language servers echo the yaml. Anything else generates and builds cleanly and fails only at the metadata-parity gate, with a message that points at the servers. Filed separately as #195 with a suggested generate-time error.

### Verified


kevinlincg:
Rebased onto `df817f151` and **reverted the yaml workaround this PR used to need**, now that #195 is fixed.

The PR previously pinned `suggested: [1, 200, 1]` because `match_predefined_opt_input` folded QSTICK onto `TA_DEF_UI_TimePeriod_10` and replaced the yaml's triple with the constant's. With full-equality folding, the proposal's `[4, 200, 1]` survives: QSTICK now gets its own `TA_DEF_UI_D_QSTICK_TimePeriod`, and all four backends' metadata agree on it. `hint` stays `"Time period"` — that one genuinely matches the shared constant, so it still folds, which is the dedup working as intended.

So this doubles as a live exercise of the bespoke-descriptor branch your commit message notes was going live for the first time. Full `scripts/regtest.py` green on all four servers afterwards, including the metadata-parity gate that caught the original mismatch.

Also verified on the current `dev`: `cargo run -- generate` **and** `generate-servers` both produce zero drift against what is committed here.

kevinlincg:

- **`camel_case` removed from `qstick.yaml`.** Same dead key, retired by #179. Worth noting it would now be a hard failure rather than a silent discard: `47aa52ae3` turns on `deny_unknown_fields`, so this branch as it stood an hour ago would fail to generate on current `dev`. Removing it and regenerating on `71c88e046` is the check that it is actually gone — `cargo run -- generate` succeeds and leaves zero drift.

kevinlincg:

kevinlincg:

PR #197, opened 2026-08-12
feat(efi): Elder's Force Index

kevinlincg:
Implements `EFI` per proposal TA-Lib/ta-lib#206.

    force[t] = ( close[t] - close[t-1] ) * volume[t]
    EFI      = EMA( force, optInTimePeriod )

Default 13 (Elder's intermediate-term reading; his short-term one is 2 — that is the parameter, not a second formula). Lookback is `optInTimePeriod + TA_GetUnstablePeriod(TA_FUNC_UNST_EMA)`. No new `FuncUnstId` row and no `unstable_period` yaml flag: EFI keys off `TA_FUNC_UNST_EMA` the way DEMA/TEMA/TRIX/MACD do, and is added to `UNSTABLE_MAP[]` in `test_codegen.c` alongside them.

`efi.c` is `ema.c`'s arithmetic with `inReal[t]` replaced by `force[t]`, including the `(x - prevMA)*k + prevMA` ordering. That ordering is the bit-exactness contract against the composed reference, and its 1-ULP nature is not theoretical — the sabotage run below rewrites it as `k*x + (1-k)*prev` and the differential catches it.

### The decision with no gate coverage

**METASTOCK seeding.** `ema.c` has a live `TA_COMPATIBILITY_METASTOCK` branch that every EMA composite inherits through `ema_private`; a fused EFI does not, and the mapping is ill-posed for a derived series — "the first element of the array handed to EMA" is `force[1]`, there being no `force[0]`. This seeds from `force[1]`, so `TA_EFI` stays equal to `EMA(MOM(close,1) * volume)` in that mode too.

Nothing else in the tree would notice if that were wrong: the cross-language sweep diffs C/Rust/Java/.NET against each other and all four are generated from this same input file, so it agrees vacuously, and the external arm ignores the compatibility setting. So the differential leg is run a **second time** under METASTOCK. A seed taken one bar late is caught only there.

### Tests

Five legs in `test_composite.c`:

- **Differential**, bit-exact against `TA_MOM(close,1)` → `TA_MULT(mom, &volume[1])` → `TA_EMA`, over periods {1,2,13,30} × unstable {0,5} × startIdx {0,1,40,200}. Ranges are compared as well as values — the force series is one shorter than the bar series and `force[k]` belongs to bar `k+1`, so a reference silently one bar out would otherwise agree on a shifted series that is nearly as smooth.
- **Differential under METASTOCK**, as above.
- **External oracle**, six values from pandas-ta-classic 0.6.52 at 1e-12 relative. Single-arm and not fully independent: Tulip ships no force index, and pandas-ta-classic's EMA was written to reproduce TA-Lib's seeding, so agreement on the warm-up anchor is by construction. What it *does* corroborate is the algebra over a 239-value trajectory — precisely what the differential cannot see.
- **Degenerate inputs**, flat close and zero volume, both exactly `0.0`. Structural: nothing on the data path divides by anything derived from the data.
- **In-place aliasing**, `outReal` over `inClose` and over `inVolume`, periods 1/2/13, startIdx 0 and 1. `prevClose` is carried in a scalar for this reason — at startIdx 1 the slot holding `close[t-1]` already holds an output.

Sabotage-checked, all five rejected and green after restore: smoothing factor changed to `2/period`; the recurrence rewritten as `k*x + (1-k)*prev`; force paired with the previous bar's volume; the lookback short by one; the METASTOCK seed one bar late.

### Verified


kevinlincg:

kevinlincg:

- **`camel_case` removed from `efi.yaml`.** Same dead key, retired by #179. Worth noting it would now be a hard failure rather than a silent discard: `47aa52ae3` turns on `deny_unknown_fields`, so this branch as it stood an hour ago would fail to generate on current `dev`. Removing it and regenerating on `71c88e046` is the check that it is actually gone — `cargo run -- generate` succeeds and leaves zero drift.

kevinlincg:

mario4tier:
Please remove support for TA_COMPATIBILITY_METASTOCK for *new* functions.

We are deprecating the capability to set the TA_COMPATIBILITY_METASTOCK (to verify if there is any opposition from the community). If there is no major revolt against that decision, then all TA_COMPATIBILITY_METASTOCK variant will be removed in a future release.

( We are not even supporting TA_SetCompatibility from Rust/Java/C# API ) 

In short, we are preserving the support to existing function for TA_COMPATIBILITY_METASTOCK in C... but dropping it from new functions and Haave/Rust/C# backends.

kevinlincg:
Done — `TA_COMPATIBILITY_METASTOCK` is gone from EFI. Pushed.

The seeding arm is removed and `efi.c` no longer consults `TA_GetCompatibility()`; the comment there states the reason in your terms, that honouring it would make the C output diverge from three backends which cannot read the setting at all.

**I turned the removal into a pinned contract rather than just deleting the leg.** The test that was the METASTOCK differential is now "`TA_SetCompatibility` is inert for EFI", swept over the same period grid.

That test needed a control, because the obvious form of it passes for two different reasons and only one is the contract: *EFI under METASTOCK equals EFI under DEFAULT* also passes if the setting never reached the library at all. So the leg additionally requires the **same force series through `TA_EMA`** — which does still honour the setting — **to change**. If that control ever stops moving, the leg reports that it has stopped testing anything instead of quietly passing. Period 1 is exempt from the control since EMA's identity copy is seeding-independent by construction.

Both directions sabotage-proved:

- METASTOCK arm added back → `EFI compatibility Fail [period 2] at out[0]: METASTOCK 4075514.83… != DEFAULT 7163838.25 — TA_SetCompatibility must be inert for this function`
- control disabled → `the control did not move — TA_SetCompatibility no longer changes TA_EMA either, so this test would pass whether or not EFI honoured it`

Full `scripts/regtest.py` green on all four servers afterwards: `Stream verify: 170 functions, 17156 legs bit-exact vs batch`, 161 passed / 0 failed per language.

Checked the other two in flight: #194 (MARKETFI) and #196 (QSTICK) never consulted `TA_GetCompatibility()` — MARKETFI is stateless and QSTICK is a plain SMA window — so there is nothing to remove there.

kevinlincg:
Correction to my previous comment: the commit I pushed a minute ago was missing one regenerated file, `ta_codegen/output/java/tools/TaCodegenServe.java` — the Java server's copy of the doc comment still described the old METASTOCK seeding. Fixed and force-pushed as `024d4ed88`; the branch now regenerates to zero drift under both `generate` and `generate-servers`.

Same trap as before and worth naming again since it has now caught me twice: `cargo run -- generate` does not emit that file, only `generate-servers` does. A branch can therefore pass "regenerate leaves the tree clean" while carrying a stale server artifact, because the check itself never writes it. `scripts/regtest.py` does regenerate it — which is how it appeared as an uncommitted change *after* the gate, i.e. the gate tested the correct bytes and the commit did not contain them.

The gate result stands: the tested tree is byte-identical to what is now pushed.

kevinlincg:

PR #198, opened 2026-08-12
feat(wad): Williams' Accumulation/Distribution

kevinlincg:
Implements `WAD` per proposal TA-Lib/ta-lib#200.

    TRH = max( prevClose, high )      TRL = min( prevClose, low )
    AD  = close - TRL   if close > prevClose
        = close - TRH   if close < prevClose
        = 0             if close == prevClose
    WAD = running sum of AD

### Two rulings this bakes in

**It consumes no volume.** Larry Williams' original multiplies each move by that bar's volume; Achelis published the modification that drops it (*A to Z*, 2nd ed., p.368), the industry kept Williams' name on it, and the no-volume form is what Tulip, pandas-ta-classic, cTrader, TC2000, WealthCharts and MultiCharts all ship. Board policy says the original author wins; here the author's own name has been attached by the industry to the modification, and shipping the volume-weighted series under this name would surprise every user. The hint, `wad.md` and the source comment all say so, and describe the volume form as a different series rather than leaving it implicit.

**Lookback 0**, the open question in the proposal. The first bar is measured against itself, contributes exactly `0.0`, and is emitted — following all four cumulative lines already in the tree (OBV, AD, NVI, PVI). Tulip's `ti_wad_start()` returns 1, so its series is this one without the leading zero, and the book vector is 12 bars in / 11 values out. Seeding is an API convention rather than a formula question and the author says nothing about it, so in-tree precedent decides. **Cheap to reverse if you rule the other way**: one line in `wad_lookback` plus a one-index shift in the pinned values.

Group is `Volume Indicators` for discoverability next to AD/ADOSC/OBV, not because it reads volume — `Momentum` would be the literally accurate label.

### Tests

In `test_per_hlc.c`, so no new file and no dual registration to keep in sync.

- **Ten pinned values** on the reference corpus, recomputed independently from `test_data.c` rather than read back out of the implementation. `CHECK_EXPECTED_VALUE` is adequate here, unlike for MARKETFI: its band is an absolute 0.01 and WAD spans [-20.73, 45.77] on this corpus. The discriminating question is *where* to pin, and the answer is the flat arm, which random fuzz doubles never produce. It has two occurrences: **the seed bar**, at every startIdx, since it is measured against itself — so the `out[0] == 0.0` golden is the cheapest guard in the table and not decoration — and **bar 101**, the corpus's only pair of equal consecutive closes (`close[101] == close[100] == 116.0`), with goldens on both sides of it.
- **Two published vectors**, checked by hand: Achelis p.368 (`atoz.txt:270`) and `untest.txt:471`. Both reproduce exactly. They are compared with an explicit one-element offset for the leading zero rather than fed through a generic harness, where a silent misalignment would be indistinguishable from agreement.
- **Start dependence**, pinned at startIdx 100 and at a single-bar range. Flagged `path_dependent`, so the range sweep runs with `TA_DO_NOT_COMPARE` — AD and OBV's treatment.

Sabotage-checked, all five rejected and green after restore: `>` turned into `>=` (caught by the `out[0] == 0.0` golden, since the seed bar is itself a flat-arm case); high and low swapped; the true-range clamp removed; lookback changed to 1; the accumulator reset each bar.

### Drive-by

`WAD` and `ACCBANDS` are added to this file's `DO_TEST` filter string. Without it `--function=WAD` selects nothing and reports success — a green that means "nothing ran", which is how I first "passed". ACCBANDS had the same gap; its rows still ran in a full pass, they were just not selectable by name.

### Verified


kevinlincg:
The proposal moved from `ta-lib-proposal-drafts#25` into this repo as #200 while this was open, so I force-pushed one line: the CHANGELOG entry now cites `(#200)` instead of the drafts number, which no longer resolves. Nothing else changed — same tree otherwise, same gates.

kevinlincg:

mario4tier:
Merged to `dev` as 52534ff3c. Thanks — this was a clean one to review.

**Security review first**, before any build or test, since the diff is 4275 lines from a fork. The decisive check was that a full `cargo run -- generate` reproduces every generated file byte-identically, which collapses the diff to the ~150 hand-written lines in `wad.c` / `wad.yaml` / `wad.md` / `test_per_hlc.c` and makes them the only thing that needs reading. Nothing outside the expected paths, no build hooks, no CI edits.

**We did add the two oracle arms**, on our own box. `ta_tulip_serve` already had a `TA_WAD` arm; `ta_pandas_serve` now has one too. Both compute the identical formula — I read their sources rather than trusting the name. Then all four generated backends, both oracles, and a from-scratch Python reference were driven over the same inputs and compared as raw IEEE-754 bits across seven shapes (reference corpus, tie-heavy, constant, monotone up/down, gapped, random walk): **bit-identical everywhere, tolerance 0.0**, exactly as #200 predicted.

Your alignment caveat was the right thing to insist on, and it went further than the proposal knew: **pandas is index-for-index with us**, not offset. Bar 0's `prev_close` is NaN so both masks are False and `ad[0]` is exactly `0.0` — pandas emits a leading zero at full length, which *is* the lookback-0 seed bar. Tulip is the one at +1. Two oracles, two different alignments, which is precisely the shape where a generic harness would have agreed with itself.

Everything else held up. The ten pinned goldens reproduce to full precision from an independent recomputation, `close[101] == close[100] == 116.0` is the corpus's only equal-close pair, both published vectors reproduce, `--xlang-hash` sweeps WAD non-vacuously (244 cases per language, 0 mismatches), stream verify is 32 legs bit-exact against batch, and `OpenAndFill == batch` bitwise. All five of your sabotage mutations are caught by shipped gates, and your correction to the proposal was right: the seed-bar `out[0] == 0.0` golden catches the `>` → `>=` flip at every `startIdx`, earlier and more cheaply than bar 101 does. Benchmarks show nothing — 1.00x batch at 5% spread, 0.92x streaming, mid-pack among AD/OBV/NVI/PVI.

We also ran an adversarial multi-agent review across numerics, metadata, tests, streaming and backend parity, and nothing substantive came back. The one documentation improvement we made on top: `wad.md` asserted the Williams/Achelis history without a `## References` section carrying the sources, so we added one (plus `See Also`) and trimmed the vendor list to the two implementations verifiable in our oracle tree.

**Two small things were on us, not on you**, and both are already fixed on top in 4281bd2a1:

- `camel_case: Wad` — that key was retired from the schema by #179, which landed on `dev` only three days before your PR. Easy to be working from a base that predates it.
- The comment saying the function is "flagged `start_dependent`" — that name went away when `TA_FUNC_FLG_START_DEP` became `TA_FUNC_FLG_PATH_DEP`. Your `wad.yaml` declares `path_dependent` correctly; only the prose lagged.

Both were harmless — one silently discarded, one a comment — and we checked our own docs before assuming anything: the schema table, the `new-ta-func` skill and all three `CLAUDE.md` were updated in the same commits as each rename, so there were no stale instructions pointing you at either name.

The real lesson was ours. A retired key survived a *byte-identical* regeneration, which means nothing downstream could ever have noticed it — the input YAML parser had no `deny_unknown_fields`, so serde silently skipped anything it did not recognise, and a typo like `hints:` for `hint:` read as "no hint" rather than as an error. Fixed in 47aa52ae3: the run now stops and names the file, the key, its line and the expected fields. All 169 existing definitions parse unchanged. Your PR is what surfaced it.

Also picked up your two ops findings — the `--function=WAD` filter gap (a green meaning "nothing ran" is the worst kind) and the stale `test_codegen.c` function-count literal. Catching the `ACCBANDS` instance of the same gap while you were in there was a nice touch.

Thank you — genuinely good work. The verification plan was thorough, the two rulings were argued from evidence rather than preference, you documented what you had *not* verified instead of implying full coverage, and you corrected the proposal where it was wrong. That is exactly how we would want this done.

PR #199, opened 2026-08-12
perf(codegen): mask the streaming extrema ring in the Rust backend too

kevinlincg:
Stacked on `bench-modulo-ring` (#193), not on `dev` — it only makes sense on top of your C change, and it is trivially rebaseable if you reshape that one.

Your C commit left the Rust stream emitter on the default `extrema_slot`, so the same 14 indicators still index their extrema ring through `% sp.xCap` there. This overrides it the same way.

The Rust side is slightly smaller than the C side: `Peek` clones the state rather than memcpy'ing into a mirror, so there is no `xPhys` field to carry — only `xMask` outlives `Open`. `xMask` joins the i32-forced sentinel set so the slot renders as `(sp.today & sp.xMask) as usize`, the same shape as the modulo it replaces, with no extra cast.

### Measured

Intel i7-10700K, rustc release with `lto = true`, `codegen-units = 1`. Update ns/bar, best-of-3 passes, **8 interleaved A/B runs per arm**, medians. Harness drives `Stream::update` directly over a fixed random-walk corpus with the same `it & 4095` bar feed `ta_bench_stream` uses.

| func | modulo ns | mask ns | chg |
|---|---:|---:|---:|
| MININDEX | 11.94 | 4.91 | −58.9% |
| MAXINDEX | 11.62 | 5.04 | −56.7% |
| MIN | 10.67 | 4.82 | −54.8% |
| MAX | 10.61 | 4.90 | −53.8% |
| MINMAX | 18.82 | 8.93 | −52.5% |
| MINMAXINDEX | 19.62 | 9.37 | −52.2% |
| WILLR | 20.44 | 10.43 | −49.0% |
| MIDPRICE | 17.54 | 9.35 | −46.7% |
| VAR | 8.23 | 4.45 | −45.9% |
| AROONOSC | 18.06 | 9.87 | −45.4% |
| MIDPOINT | 16.11 | 8.82 | −45.2% |
| AROON | 18.42 | 10.14 | −44.9% |
| STOCHF | 23.92 | 15.79 | −34.0% |
| STOCH | 29.55 | 21.02 | −28.9% |

Median **−47.8%** over the 14. Mechanism verified rather than inferred: `idivl` in the bench binary drops **110 → 34**, and `WILLR_Stream::update` goes from **6 to 0**.

**Where this measurement is weak, stated plainly:** my Rust harness has only 5 control functions, and one of them (WMA, untouched, on the `ringPos_` lowering) moves −39.6% *stably* across all 8 runs in each arm — 2.91 ns vs 1.76 ns, not noise. It is fully inlined into `main` along with everything else in a `codegen-units = 1` LTO binary, so changing 14 bodies moves the layout of the rest. That is the same effect your C table flags for IMI/MFI/PLUS_DM, but with 5 controls I cannot bound it the way your 154 can. So: the direction and the rough size are solid — the `idiv` counts and the C run's 154-function control carry that — but I would not defend the second digit of any single Rust row.

### Correctness

Full `scripts/regtest.py` on all four servers: 161 passed / 0 failed per language, and `Stream verify: 168 functions, 16678 legs bit-exact vs batch` on the Rust server.

### DIFF
--- a/ta_codegen/generator/src/backends/rust_stream.rs
+++ b/ta_codegen/generator/src/backends/rust_stream.rs
@@ -170,6 +170,13 @@ impl streaming::NameMap for RustStreamNames {
     fn extrema_cap(&self) -> String {
         "sp.xCap".to_string()
     }
+    fn extrema_slot(&self, idx: Expr) -> Expr {
+        Expr::BinOp(
+            Box::new(idx),
+            crate::ir::BinOp::BitwiseAnd,
+            Box::new(Expr::Var("sp.xMask".to_string())),
+        )
+    }
 }
 
 // ---------------------------------------------------------------------------
@@ -419,6 +426,7 @@ fn state_fields_from(
     }
     if let Some(ex) = model.extrema() {
         fields.push(("xCap".into(), "i32".into(), "1_i32".into()));
+        fields.push(("xMask".into(), "i32".into(), "0_i32".into()));
         for arr in &ex.arrays {
             fields.push((
                 format!("x_{arr}"),
@@ -715,6 +723,7 @@ fn build_step_ctx(func: &FuncDef, models: &[&StreamModel], typing: &Typing) -> R
         }
         if let Some(ex) = model.extrema() {
             ctx.sentinel_vars.insert("xCap".to_string());
+            ctx.sentinel_vars.insert("xMask".to_string());
             for arr in &ex.arrays {
                 ctx.vec_vars.insert(format!("x_{arr}"));
                 ctx.real_array_vars.insert(format!("x_{arr}"));
@@ -897,7 +906,7 @@ fn emit_extrema_rebase(o: &mut String, model: &StreamModel, indent: usize) {
         let _ = writeln!(o, "{pad}if sp.{} >= 1073741824 {{", model.cursor);
         let _ = writeln!(
             o,
-            "{inner}let rebaseShift: i32 = (sp.{} / sp.xCap) * sp.xCap;",
+            "{inner}let rebaseShift: i32 = sp.{} & !sp.xMask;",
             ex.trailing
         );
         for v in &vars {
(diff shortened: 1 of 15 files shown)


mario4tier:
Merged — thank you, this was the right shape and the right level of scepticism about your own numbers.

Merged with `--no-ff` off the fetched PR head so your commit keeps its authorship in `bench-modulo-ring`, which is now merged into `dev` (`091bb50ed`).

I re-measured rather than take the table as read: one timed block per function, generated from the emitted `*_Open`/`update` signatures, each arm built against its own copy of the library, 6 interleaved rounds with **alternating arm order** (a fixed order cost me a spurious −1% control median in an earlier pass), and all 168 stream functions so the 154 untouched ones bound the layout noise.

| | affected median | control median (154 funcs) |
|---|---|---|
| Rust update | **−42.9%** (−3.01 ns) | +0.4% |
| Java update | **−6.1%** (−2.37 ns) | −0.2% |

Direction and size hold up. Your caveat about the second digit was the right call — and your WMA anomaly has a name on this box: `IMI`, untouched, reads **+185% stably across every round** in the mask arm, the same layout effect the C run flagged. With 154 controls the aggregate stays clean, which is the only thing 5 controls could not give you.

Two things came out of the review, both on top of your commit rather than in it:

- With all three streaming backends masking, the `% cap` default had no user left. `extrema_cap` became `extrema_mask`, the mask is now the shared default, and the three identical `extrema_slot` overrides (including yours) are gone. `xCap` is write-only in the Rust and Java handles once the slot map stops reading it, so it is dropped there; C's Open still reads it.
- `peek` deep-copies the handle in Rust and Java, so it copies the *rounded* ring. Rust peek at default periods reads worse — but the 154-function control median moves +5% in that same run, so the measurement is allocator-bound and confounded, and at a power-of-two period (where the rounding is a no-op) it flips to −4.5%. C's peek, which memcpy's a preallocated mirror, gains. Not a blocker, but it is the one number that is not better.

Java followed the same shape in `354af6ee3`. Gates on the merge: stream verify bit-exact vs batch on all three streaming servers, `--xlang-hash` bit-identical across four languages, C and C# output unchanged.

PR #208, opened 2026-08-13
fix(codegen): emit the declared sweep triple in the XML, and gate the two surfaces (#207)

### LINKED ISSUE #207: ta_abstract: generated XML discards the suggested sweep triple for every integer optional parameter — emits the range maximum instead (71 of 171 functions)
The C abstract table and the generated XML disagree about
`suggested_start` / `suggested_end` / `suggested_increment` for every integer
optional parameter. The table is right; the XML emits the range **maximum** for
all three.

QSTICK, both surfaces, same parameter:

| surface | start | end | increment |
|---|---|---|---|
| `TA_GetOptInputParameterInfo()` — `src/ta_abstract/tables/table_q.c:42` | 4 | 200 | 1 |
| `ta_func_api.xml` + the XML embedded in `ta_func_api.c` | 100000 | 100000 | 100000 |

4/200/1 is what `qstick.yaml` declares (`suggested: [4, 200, 1]`).

## Cause

`ta_codegen/generator/src/backends/func_api_xml.rs:258`, in `write_integer_opt`:

```rust
// gen_code uses max for all integer suggested values.
let _ = writeln!(out, "\t\t\t\t\t<SuggestedStart>{max_i}</SuggestedStart>");
let _ = writeln!(out, "\t\t\t\t\t<SuggestedEnd>{max_i}</SuggestedEnd>");
let _ = writeln!(out, "\t\t\t\t\t<SuggestedIncrement>{max_i}</SuggestedIncrement>");
```

`opt.suggested` is in scope and carries the right triple — `write_real_opt`
(same file, line 220) uses it correctly. Only the integer path discards it.

The stated reason is bug-compatibility with `gen_code`, which was removed in the
v0.7.1 cutover. Nothing verifies the XML against it any more: `ta_ref_serve` is
not a metadata oracle, so no gate covers this either way.

## Scope

All 5 `TA_IntegerRange` constants in the tree are affected — shared across
**71 of 171** functions in the XML:

| range constant | min | max | table (start,end,incr) | XML emits |
|---|---|---|---|---|
| `TA_DEF_TimePeriod_Positive` | 1 | 100000 | (1,200,1) | (100000,100000,100000) |
| `TA_DEF_TimePeriod_Positive_Minimum2` | 2 | 100000 | (4,200,1) | (100000,100000,100000) |
| `TA_DEF_TimePeriod_Positive_Minimum5` | 5 | 100000 | (5,200,1) | (100000,100000,100000) |
| `TA_DEF_QSTICK_TimePeriod` | 1 | 100000 | (4,200,1) | (100000,100000,100000) |
| `TA_DEF_HorizontalShiftPeriod` | -200 | 200 | (0,8,1) | (200,200,200) |

Real-typed optional parameters are unaffected.

`TA_DEF_HorizontalShiftPeriod` is the clearest illustration of why the emitted
values are not usable: the parameter's domain is [-200, 200] and the XML
suggests sweeping from 200 to 200 in steps of 200.

## Impact

These fields exist for parameter-optimisation sweeps. A consumer reading the
shipped `ta_func_api.xml` gets a degenerate one-point sweep at the top of the
range; the same consumer reading `TA_GetOptInputParameterInfo()` gets the
intended one. `ta_func_api.xml` ships at the repo root and is embedded in the
Rust crate.

## Fix

Use `opt.suggested` in `write_integer_opt`, as `write_real_opt` already does,
and drop the `gen_code` reference. Regenerating moves `ta_func_api.xml`,
`src/ta_abstract/ta_func_api.c` and `ta_codegen/output/rust/library/src/ta_func_api.xml`.

Worth deciding at the same time whether a metadata-parity gate should compare
the two surfaces, since nothing does today.

Found while reviewing #196.

kevinlincg:
Fixes #207, and adds the gate the issue asks about.

### The fix

`write_integer_opt` now uses `opt.suggested`, as `write_real_opt` already did. The `gen_code` bug-compatibility comment goes with it.

Measured by diffing the regenerated artifact per function rather than asserted: **71 of 171 functions change, and the only tags that move are `SuggestedStart`, `SuggestedEnd`, `SuggestedIncrement`** — nothing else in the XML shifts. QSTICK's `optInTimePeriod` goes from `(100000, 100000, 100000)` to the `(4, 200, 1)` its yaml declares and `TA_GetOptInputParameterInfo()` already reported.

### One correction to the issue

`TA_DEF_HorizontalShiftPeriod` — the clearest illustration in your table, a `[-200, 200]` parameter told to sweep 200 to 200 by 200 — **never reaches the shipped XML**. `TA_DEF_UI_HorizontalShift` is defined in `ta_def_ui.c` and referenced by no function table, so that constant is dead; only the four live ones were ever emitted. The count of affected functions is unchanged.

### The gate

`ta_codegen/generator/tests/xml_metadata_suite.rs`.

It reads the **emitted XML** and compares it against the yaml, rather than rendering both surfaces from the IR and comparing those. That distinction is the point: a test built the second way agrees with itself no matter what either emitter does. Parsing the artifact means it fails if the XML writer *or* the yaml parser drifts, in either direction.

**Sabotage-proved in the only order that means anything** — written first, run against the unfixed tree, where it fails naming the functions and both values:

```
ta_func_api.xml disagrees with the YAML for 87 parameter(s):
  MIN.optInTimePeriod: YAML declares (4 200 1), XML emits (100000 100000 100000)
  EFI.optInTimePeriod: YAML declares (1 200 1), XML emits (100000 100000 100000)
  ...
```

It turns green only after the emitter is fixed.

It also refuses to pass vacuously: the run asserts it reached more than 50 integer parameters, because the integer path is the one that was wrong and a refactor that stopped emitting `<Range>` for integers would otherwise leave it green with nothing compared. Real parameters are sampled too, compared numerically rather than textually — `double_to_str`'s formatting is not the contract.

### Verified

745 generator tests pass; `scripts/regtest.py` green on all four servers, `Stream verify: 171 functions, 17244 legs bit-exact vs batch`.


kevinlincg:
Superseded by the gate-only branch — closing since `2c4c33ed5` landed the emitter fix and this PR carried both.

You reached the same fix independently, and the same measured numbers (87 slots, 71 of 171 functions). What is not in `dev` is the gate, and your own commit note is the argument for it: *the parity gate cannot catch this class on its own, because C and the servers regenerate from the same IR and a wrong derivation moves them together.* A check that reads the emitted XML against the yaml does catch it.

Reopened as a single new file, no emitter changes: #210.

mario4tier:
Thanks for this, Kevin.

The core fix (`write_integer_opt` taking the triple from `opt.suggested` instead of emitting the range max three times) landed independently in dev at `2c4c33ed5` a bit earlier the same day, also closing #207. I diffed the regenerated `ta_func_api.xml` and `ta_func_api.c` between dev and this branch and they come out byte-identical — the only delta is MARKETFI, which merged to dev after this branch was cut, so that's just staleness, not a real difference.

What dev is still missing is the gate: `ta_codegen/generator/tests/xml_metadata_suite.rs`. That's exactly what #207 asked for and it's genuinely new. Could you rebase onto dev and reduce this PR to just that test file? Appreciate you building it either way.

PR #210, opened 2026-08-14
test(codegen): gate ta_func_api.xml against the declaration (#207)

kevinlincg:
One new file, no emitter changes — `2c4c33ed5` already landed the fix, and we reached it independently with the same numbers (87 slots, 71 of 171 functions).

What is not in `dev` is a gate, and **your own commit note is the argument for one**:

> the parity gate cannot catch this class on its own — C and the servers regenerate from the same IR, so a wrong derivation moves them together.

This check can, because it reads the **emitted `ta_func_api.xml`** and compares it against the yaml. A derivation that is wrong in the XML writer fails here regardless of what the other surfaces do.

It is deliberately not written the other way. A test that renders both surfaces from the IR and compares *those* agrees with itself no matter what either emitter does — the same shape as the MARKETFI leg I had to disown, and the reason #207 survived long enough to be found by inspection rather than by a gate.

### Discriminating power, shown your way

Reverting the `write_integer_opt` hunk and regenerating: QSTICK's `SuggestedStart` returns to `100000` and this fails naming **all 87 parameter slots** — the same count your commit message reports for the cross-surface comparison. Restored, it passes on `dev` as it stands.

It also refuses to pass vacuously: the run asserts it reached more than 50 integer parameters, the path that was wrong, so a refactor that stopped emitting `<Range>` for integers cannot leave it green with nothing compared. Reals are sampled too, compared numerically rather than textually — `double_to_str`'s formatting is not the contract.

### Verified

745 generator tests pass.

### DIFF
--- a/ta_codegen/generator/tests/xml_metadata_suite.rs
+++ b/ta_codegen/generator/tests/xml_metadata_suite.rs
@@ -1,0 +1,190 @@
+//! `ta_func_api.xml` must agree with what the YAML declares (issue #207).
+//!
+//! Two metadata surfaces ship for every optional parameter: the C abstract
+//! table, read at run time through `TA_GetOptInputParameterInfo()`, and the
+//! generated `ta_func_api.xml`, which sits at the repo root and is embedded in
+//! `src/ta_abstract/ta_func_api.c` and in the Rust crate. Both are generated
+//! from the same YAML. Nothing compared them.
+//!
+//! They disagreed for every integer optional parameter across 71 of 171
+//! functions: `write_integer_opt` emitted the range *maximum* three times in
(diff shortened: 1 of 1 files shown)


mario4tier:
Merged as 1d1bbd0b3. Exactly the reduction asked for on #208 — thanks for the quick turnaround.

**sec-check:** CLEAR — single new test file, no generator source/build/CI surface touched.

**Verified:**
- `cargo test --test xml_metadata_suite`: passes standalone.
- Full generator suite: 745 passed, 0 failed — matches your count exactly.
- Local `dev` fast-forwarded to `1d1bbd0b3`, working tree clean.

#207 now has both the fix and the gate on dev.

Issue #216, opened 2026-08-14
testing: candle helpers have no cross-language value coverage and no candlestick is ever run at non-default candle settings


mario4tier:
## Summary

The candle helpers (`ta_candlerange`, `ta_candleaverage`) have no cross-language
value coverage, and no candlestick is ever run at non-default candle settings
anywhere in the test suite. A live four-language divergence (#217) sat undetected
because of it.

This issue tracks building that coverage. It depends on #215 for the settings
transport.

## Why the existing gates cannot catch this class

Worth reading before designing, because the obvious fix (more seeds / more shapes /
more settings on the existing gates) does not work.

`--xlang-hash` already iterates every fuzz shape cross-language, and already covers
`CDLLONGLINE` at default settings — where `TA_ShadowShort` **is** `Shadows`-typed, so
the divergent arm is genuinely executed. It still passes.

Measured: a faithful `CDLLONGLINE` evaluated under both spellings of the `Shadows`
arm over all 9 fuzz shapes x 20,000 seeds x 256 bars = **44M bars produced zero
output differences** — despite `FUZZ_EXTREME` yielding bit-differing intermediates on
10,041 of 51,200 bars and `FUZZ_WITH_ZEROS` on 4,925 of 51,200.

The reason is structural: a candlestick output is 3-valued (`-100/0/+100`), so a
1-ULP shift in a threshold essentially never straddles a comparison boundary. **Any
gate that hashes candlestick outputs is blind to reassociation divergence in the
helpers.** The intermediate has to be made observable.

## Gaps

1. **The synth gate has no candle coverage.** No `synth<n>/` calls `ta_candlerange`
   or `ta_candleaverage`; none takes OHLC inputs.
2. **No candlestick is ever run at non-default settings.** The `TA_CDLGlobals` struct
   (`test_candlestick.c:111`) — `rangeType`/`avgPeriod`/`factor` for all 11 settings —
   is declared and **never used**. The only real `TA_SetCandleSettings` call outside
   parameter validation is `test_variants.c:365`, which pins to the defaults. Either
   wire `TA_CDLGlobals` up or delete it; do not leave it dormant.
3. **No cross-language settings variation is possible** — see #215. `sv_candle_avg`
   (`ta_codegen_serve.c:417`, 4 modes including all-`Shadows`) exists but is reachable
   only from `stream_verify`, which compares batch against stream inside one server.
[...]

Issue #215, opened 2026-08-14
servers: no JSON-RPC method can set candle settings, so every cross-language gate runs only at defaults


mario4tier:
## Summary

No JSON-RPC server exposes a way to set candle settings, so every cross-language
gate can only ever run at the default settings. This blocks any test that wants to
vary `rangeType` / `avgPeriod` / `factor` across languages.

The **library** APIs already exist and are at parity in all four backends:

| Backend | API | Shape |
|---|---|---|
| C | `TA_SetCandleSettings` / `TA_RestoreCandleDefaultSettings` | global (`ta_global.c`) |
| Rust | `Core::candle_setting(...)` | builder, immutable `Core` (`ta_func/types.rs`) |
| Java | `CoreBuilder.candleSetting(...)` | builder |
| C# | `CoreBuilder.CandleSetting(...)` / `RestoreCandleDefault(...)` | builder |

What is missing is the transport. `grep -c set_candle_settings` returns **0** in all
four servers. The servers already carry the two other global-state setters
(`set_unstable_period`, `set_compatibility`), so the pattern to follow is established.

## Current state per server

| Server | candle mentions | notes |
|---|---|---|
| `ta_codegen_serve.c` | 313 | has `sv_candle_avg()` (4 modes: default / `avgPeriod+3` / `avgPeriod=0` / all-`Shadows`), but it is reachable **only** from `stream_verify` via the `candleLegs` flag — which compares batch against stream *inside one server*, never across languages |
| `ta_codegen_serve.rs` | 217 | no settings entry point |
| `TaCodegenServe.java` | 2710 | no settings entry point |
| `TaCodegenServe.cs` | **0** | no candle handling of any kind; holds `static Core core = new Core()` and mutates `core.unstablePeriod[...]` directly |

C# needs the most work, but not because the library is missing anything — the server
simply never touches candle settings. Note that in Rust, Java and C# the `Core` is
immutable and settings arrive through a builder, so the server has to **rebuild** the
`Core` when settings change rather than mutate in place; the C server sets a global.

## Proposed

Add a `set_candle_settings` method (and a restore-to-defaults counterpart) to all four
servers, mirroring how `set_unstable_period` is already handled. Suggested params:
`settingType`, `rangeType`, `avgPeriod`, `factor`, matching `TA_SetCandleSettings`.

Wire-format reminders that have bitten before: nested `params`, no space after the
colon, non-standard FNV basis. Flat fields panic the Rust server.

## Why now

[...]

PR #218, opened 2026-08-15
fix(codegen): spell the candle Shadows range as upper + lower in all three generated backends

### LINKED ISSUE #217: ta_candlerange: the Shadows arm is spelled two different ways — C disagrees with Rust/Java/C# (unreleased backends only)
> **Not a released bug.** Nothing that has ever shipped is affected — see
> [Release impact](#release-impact) below. This exists only on `dev` and `main`, in
> backends that have never been published.

## Summary

`ta_candlerange`'s `Shadows` arm (`TA_RangeType_Shadows`) has **two source texts**, and
they are not bitwise equal:

| Backend | Location | Expression |
|---|---|---|
| C | `src/ta_func/ta_utility.h:219` (macro `TA_CANDLERANGE`; stream mirror at `:245`) | `(high - max(open,close)) + (min(open,close) - low)` |
| Rust / Java / C# | `ta_codegen/input/helpers/candlestick.c:41` (case 2) | `high - low - fabs(close - open)` |

They never had to agree because `backends/c.rs` maps `ta_candlerange` onto the macro
(`C_CANDLE_MACRO_FNS`, `c.rs:21`) instead of emitting the helper — `grep -rn
ta_candlerange src/` returns nothing. The two are algebraically identical and
differ in floating point by reassociation.

## Reproduction

At **default** candle settings, no `TA_SetCandleSettings` call. `TA_ShadowShort`
defaults to `{Shadows, 10, 1.0}`, and `CDLLONGLINE` reads it:

```
o = [0]*9 + [0.0,  0.0]
c = [0]*9 + [2.0, -1.0]
h = [0]*9 + [9007199254740994.0, 450359962737049.6]
l = [0]*9 + [-1.0, -1.0]

CDLLONGLINE(0, 10, o, h, l, c)
  C            ->    0
  Rust         -> -100
  Java         -> -100
  C#           -> -100
```

The bar-9 upper shadow is exactly 2^53. The C form evaluates to `fl(2^53 + 1)` = 2^53;
the generated form to `fl(2^53 + 3) - 2` = 2^53 + 2. That difference crosses the
`upper < avg` comparison.

## Reachability

The boundary is Sterbenz' lemma: while a bar's low stays above half its high, every
subtraction in both forms is exact and the orderings cannot disagree.

| Input class | Samples | Bitwise differences |
|---|---|---|
| Real daily bars, 15 symbols | 123,538 | 0 |
| Random prices, range <= 50% of high | 40,000,000 | 0 |
| Random prices, low < high/2 | 4,000,000 | ~37–40% |

So it is invisible on equity data and reachable on anything that halves within a bar —
an unadjusted split, a negative-price series, a synthetic spread.

Flipping an **output** additionally needs a near-exact tie: a random search over ten
million bars with lows as low as a thousandth of the high produced ULP differences in
the range value and **zero** output differences. This is a cross-language parity
defect, not a correctness bug anyone will hit on market data.

## Blast radius

Three shipped functions read the only `Shadows`-typed default (`TA_ShadowShort`):
`CDLADVANCEBLOCK`, `CDLLONGLINE`, `CDLSHORTLINE` — plus anything a user points at
`Shadows` via `TA_SetCandleSettings`.

`ta_candleaverage`'s two spellings are **operation-order identical** (0 differences
over 20M random inputs), so the blast radius is `ta_candlerange` case 2 alone.

## Release impact

Nothing released is affected.

- **Every GitHub release ships C only** — src tarball, Windows `.msi`/`.zip`, `.deb`.
  C uses the macro form, which is the reference behaviour, unchanged since the 2005
  candlestick commit.
- **`ta-lib-python` links the C library**, so it inherits the macro form.
- **The generated Rust/Java/C# backends are not published anywhere.** crates.io
  `ta-lib` belongs to an unrelated project (`virtualritz/ta-lib-rs`); Maven Central has
  only the legacy `com.tictactec:ta-lib 0.4.0` and `io.github.ta-lib` returns zero
  artifacts; the NuGet `Talib 0.1.0` is unrelated and `TALib.csproj` still notes that
  NuGet packaging has not landed. The only crate this repo publishes is
  `ta-lib-dispatch`, 26 lines with no candle code. No workflow contains a
  `cargo publish`, `mvn deploy` or `nuget push` step.
- **The one non-C artifact ever released is also clean.** `ta-lib-0.4.0.jar`
  (`com.tictactec`, classes dated 2007) returns `0` on the input above — verified by
  compiling against the released jar and running it, not by reading bytecode.

Both `dev` and `main` carry the divergence.

## Fix

Change `ta_codegen/input/helpers/candlestick.c` case 2 to the upper-plus-lower form
and regenerate. C is the reference: both artifacts that have ever shipped — the C
library and the 2007 `com.tictactec` Java port — agree on that spelling, from two
independent hand-written implementations twenty years apart. The generated helper is
the sole outlier.

`git diff --stat src/` must be **empty** after regenerating: C emits the macro and
never reads the helper body, so the fix is provably C-neutral. If any C file changes,
stop and investigate.

## Sequencing

Land together with the coverage from #216, as two commits in one PR that are not
squashed: the gate first (checking out that commit reproduces the failure), the fix
second (turning it green). CI is never red, and the gate is proven non-vacuous on a
real defect rather than on manufactured sabotage.

## Adjacent nit

`src/ta_common/ta_global.c:157` claims a negative `factor` "simply never matches".
In a `>` comparison it always matches — e.g.
`TA_SetCandleSettings(TA_BodyLong, RealBody, 10, -1.0)` makes `realbody > factor*avg`
unconditionally true.

kevinlincg:
Closes #217. Implements #216 **Layer 1 only** — see *Not in this PR* below.

Two commits, deliberately not squashed, in the shape the issue's Sequencing asks for: checking out the first reproduces the failure, the second turns it green.

## The fix is three places, not one

The issue lists two source texts for the Shadows arm. There are five, and two of them are copies the helper does not feed:

| Where | Spelling | Reaches |
|---|---|---|
| `ta_utility.h:219` `TA_CANDLERANGE` | upper + lower | C batch — reference |
| `ta_utility.h:245` `TA_STREAM_CANDLERANGE` | upper + lower | C stream — reference |
| `input/helpers/candlestick.c` case 2 | `h - l - \|c-o\|` | Rust |
| **`backends/java.rs:2128`** | `h - l - \|c-o\|` | **Java** |
| **`backends/csharp.rs:1770`** | `h - l - \|c-o\|` | **C#** |

`try_render_candle_ternary` in those two backends builds the expression with a `format!` string and never reads the helper body, so changing the helper alone moves Rust and leaves Java and C# untouched. The failure mode is quiet: `generate` rewrites the Java and C# outputs and they come out byte-identical, so the diff simply looks like they had nothing to change. Each backend now carries a comment naming the other.

C is unaffected by construction — `backends/c.rs` maps `ta_candlerange` onto the macro and never emits the helper body — so `git diff --stat src/` is empty after regenerating. Confirmed across three regenerations, including the sabotage run.

## Verification

The emitted expression was checked by running it, not by reading it. The Shadows arm was extracted from the generated Rust, Java and C# and evaluated on the issue's repro bar (`o=0, c=2, h=9007199254740994, l=-1`):

```
C   (macro form)  9007199254740992
Rust              9007199254740992   was 9007199254740994
Java              9007199254740992   was 9007199254740994
C#                9007199254740992   was 9007199254740994
```

Gate, on the fixtures from the first commit:

```
gate commit only   --xlang-hash FAIL, 48 mismatches across SYNTH7 and SYNTH8
+ fix commit       --xlang-hash PASS, 0 mismatches, 8 functions swept
```

Both runs sweep the same 11190 golden cases (97% with non-empty output), so the delta is the fix and not a change of scope.

**Sabotage beyond the known bug:** perturbing the unrelated `HighLow` arm by one ULP turns the gate red with 216 mismatches, and only on SYNTH7 — SYNTH8 never reads that arm. The perturbation was confirmed present in 57 generated Rust files before the run was believed; editing `input/` does not prove the generated code moved.

Also green: `regen-check` (output matches committed source, no probe residue), `check-source-lists` (41 files agree), and the generator's own suite — 747 tests, 0 failures.

## The fixtures

`SYNTH7` exposes `ta_candlerange` with one output per rangeType arm; `SYNTH8` exposes `ta_candleaverage` for both branches and both divisors, including the `avgPeriod == 0` path that reads the current bar (gap 5). Named settings at their defaults span every arm, so neither depends on #215.

Two things they turned up that are worth knowing before the next one is written:

- **These are the first OHLC consumers with `double` outputs, so the first the in-place alias gate can bite.** It failed both on their first run (10 hits on SYNTH7, 18 on SYNTH8). Every shipped candlestick escapes it because an `int` output is never aliased onto a `double` input. Two properties fix it: every read of a bar precedes any write for that bar (the C backend's macros read `inHigh[i]` directly, so storing output 1 before computing output 2 hands output 2 a clobbered bar), and the sliding window keeps per-bar values in a circular buffer instead of re-reading the trailing bar — the reason `cmf.c` carries its volume there.
- **The accumulation shape decides whether the gate can see anything.** SYNTH8's slide was first written `sum -= old; sum += new`. That went *non-vacuously* green — 243 golden cases, 100% with output, 0 mismatches — because it rounds twice against the running total instead of once against the difference, re-rounding the per-bar difference away. Writing it as the shipped candlesticks do (`sum += new - old`, one expression) restored detection.

Both fixtures land in the nightly `synth-gate`, which is where the synth family already runs; no per-PR wall-clock is added.

## Not in this PR

- **#215, and #216 Layer 2.** As scoped when I claimed the issue. Gaps 2, 3 and 6 stay open — the `TA_CDLGlobals` matrix and cross-language settings variation still need the transport.
- **Java and C# are not verified end to end here.** This box has JDK 11 and .NET 6/7 and cannot build those servers, so my gate runs are `--language=c,rust`; the per-language expression check above is real but it is not the cross-language server gate. CI is the first place all four run together on this change. Worth stating plainly because it is also how the Java/C# copies nearly went unfixed: the C+Rust legs went green on a fix that had corrected one of the three generated backends, and what caught it was noticing that a regeneration which should have touched Java produced no diff at all.
- **The `ta_global.c:157` nit.** Your reading is right — `realbody` and `avg` are both non-negative, so a negative `factor` makes the comparison very nearly always true, not never. Left out on purpose: it is a hand-written file under `src/`, and touching it would make `git diff --stat src/` non-empty and muddy the one piece of evidence that says this change cannot reach the C library. Happy to send it separately.


mario4tier:
Ran the leg you flagged as unverified. This box has **JDK 21 and .NET 10**, so all four servers build and the synth gate runs end to end. Both halves of your paired-commit sequence, `scripts/synth_gate.py` (which is the nightly `synth-gate` job, byte for byte):

## Your fix is correct in all four languages

**Gate commit `ccde2708c` — RED, and red in Java and C# too:**

```
XLANG MISMATCH TA_SYNTH7  C(golden) vs Rust  shape=5 seed=1 n=40  hash b02f74c26b4610eb/f0937a91dbc47524
XLANG MISMATCH TA_SYNTH7  C(golden) vs Java  shape=5 seed=1 n=40  hash b02f74c26b4610eb/f0937a91dbc47524
XLANG MISMATCH TA_SYNTH7  C(golden) vs C#    shape=5 seed=1 n=40  hash b02f74c26b4610eb/f0937a91dbc47524
XLANG MISMATCH TA_SYNTH8  C(golden) vs Rust  shape=6 seed=3 n=120 hash e3c3ff183b6b0d7f/af6d3a64e35da1e7
XLANG MISMATCH TA_SYNTH8  C(golden) vs Java  shape=6 seed=3 n=120 hash e3c3ff183b6b0d7f/af6d3a64e35da1e7
XLANG MISMATCH TA_SYNTH8  C(golden) vs C#    shape=6 seed=3 n=120 hash e3c3ff183b6b0d7f/af6d3a64e35da1e7

golden cases: 11190   (10846 with non-empty output = 97% non-vacuous)
Rust: 11252 cases, 48 mismatch(es)
Java: 11252 cases, 48 mismatch(es)
C#  : 11252 cases, 48 mismatch(es)
FAIL — 144 output mismatch(es) across 2 function(s).
```

48 per server, matching your `--language=c,rust` number, and the three generated
backends produce the **identical wrong hash** — which is the direct confirmation
that java.rs and csharp.rs really were carrying the same third spelling, not that
they happened to be right for another reason.

**Fix commit `91723b268` — GREEN, four languages:**

```
golden cases: 11190   (10846 with non-empty output = 97% non-vacuous)
Rust: 11252 cases, 0 mismatch(es)
Java: 11252 cases, 0 mismatch(es)
C#  : 11252 cases, 0 mismatch(es)
PASS — 8 function(s) swept: every server matches the in-process C library:
BIT-IDENTICAL (zero tolerance)

synth_gate: PASS — 8 synthetic function(s), 4 languages, stream + batch bitwise legs all exercised
```

Same 11190 golden cases in both runs, so the delta is the fix.

## One thing to fix before merge: the nightly will fail on an unrelated assertion

Neither leg ran until I patched this, and it is not in your diff — but your PR is
what triggers it, so it lands with you.

`ta_codegen/output/java/library/src/test/java/io/github/talib/test/StreamSmokeTest.java:303`

```java
check(valueTypes == 14, "14 multi-output handles carry a Value (found " + valueTypes + ")");
```

A hardcoded corpus count. `synth_gate.py` copies every `input_synth/synth<n>/`
into `ta_codegen/input/` before regenerating, so under the gate the corpus is not
the shipped one. SYNTH7 (3 real outputs) and SYNTH8 (2) are the **first synth
fixtures with more than one real output** — synth1–6 are all single-`integer`-output
— so they are the first to generate a `_Stream.Value` record, and the count goes
14 → 16:

```
FAIL: 14 multi-output handles carry a Value (found 16)
StreamSmokeTest: 1 FAILURES
  Running Java StreamSmokeTest... FAILED (exit 1)
Error: 1 server build step(s) FAILED
synth_gate: FAIL — worktree build exited 1
```

It aborts during the **build**, before either gate leg runs, so the failure names
MACD's `Value` and says nothing about candle ranges. Normal CI is unaffected
(shipped corpus → 14 → green); only the nightly `synth-gate` job breaks.

Deriving the expectation from the registry fixes it and is strictly stronger than
the literal — it pins the component count **per function** against
`FunctionInfo.outputs()`, where today only MACD's is checked. `12bbf901d` already
did this for the other counts in that file; this one was left behind. Sketch:

```java
int expectedValueTypes = 0;
List<String> wrongValue = new ArrayList<>();
for (FunctionInfo vf : Functions.all()) {
    if (vf.outputs().size() <= 1) continue;
    Class<?> handle = null;
    for (Class<?> nested : Core.class.getDeclaredClasses())
        if (nested.getSimpleName().equals(vf.name() + "_Stream")) { handle = nested; break; }
    if (handle == null) continue;               // not stream-capable
    expectedValueTypes++;
    // ...find the nested "Value", require isRecord() and
    //    getRecordComponents().length == vf.outputs().size()
}
check(expectedValueTypes > 0, "...non-vacuity...");
check(valueTypes == expectedValueTypes, ...);
```

That is what I ran on top of both commits to produce the results above; happy to
send it as its own commit if you would rather not touch it here.

## Also checked

`/sec-check` — **CLEAR**. No workflow, build-system, script or dependency changes;
`src/` and `ta_codegen/output/c/` byte-identical, so the shipped C artifact
provably cannot move; the 182-file diff is regenerated output only (every changed
file a `CDL*`, plus `Core.java` and the Java server that embeds it); the emitted
Shadows expression matches `TA_CANDLERANGE` term for term in all three backends.

## Not blocking, for the record

I picked up #215 and #216 Layer 2 in parallel, exactly as you scoped them, so gaps
2, 3 and 6 are being closed separately and your PR does not need to grow.

Issue #219, opened 2026-08-15
testing: MC/DC coverage for the remaining 55 candlesticks


mario4tier:
## Summary

49 of the 61 candlesticks have no MC/DC coverage. 12 do (10 true MC/DC, 2 Hikkake
predicate gates) — **unit 1 landed in #220**: DOJI, BELTHOLD, CLOSINGMARUBOZU,
LONGLEGGEDDOJI, DRAGONFLYDOJI, GRAVESTONEDOJI. This is #216's gap 7, split out because it is a different fault
class, a different technique, and roughly 6,000 lines of work.

**The harness is built, adversarially reviewed, and merged** (`31cb13913`). This
issue is the remaining per-pattern work, which is genuinely parallel and
delegable. Read the harness first — `src/tools/ta_regtest/ta_test_func/test_candlestick.c`,
the `pb_*` section — and copy the builder whose threshold class matches yours.

## Why MC/DC and not more of what we already have

Every other candlestick gate is **differential**: the value tables, the v0.6.4
freeze, `--xlang-hash`, `--fuzz-064`, the settings matrix. All four backends
regenerate from one IR, so an inverted or dead predicate is inverted identically
everywhere and every differential gate agrees. #217 was the same shape of
blindness from the other direction.

MC/DC is the only gate here with an **independent oracle**. Its power is in the
FLIP cases: "I broke a required condition, therefore the pattern must not fire"
is derived from the pattern's definition, not observable from the code. A
recording of current behaviour cannot produce that expectation.

Note this does **not** subsume the differential gates and is not meant to: it
cannot see a floating-point divergence (a 3-valued output essentially never
straddles a threshold on a 1-ULP shift — 44M bars, 0 flips, see #216).

## What the harness gives you

Scenario vocabulary — `pb_conditions(n)`, `pb_signs(n)`, `pb_detect`,
`pb_flip(i, cond, s)`, `pb_control(i, v, cond, s)`, `pb_waive(cond, why)` — plus a
checker that enforces:

- an **independent condition model** per pattern (transcribed by hand from
  `ta_codegen/input/<name>/<name>.c`): zero conditions false at every detect and
  control, **exactly one — the declared one — false at every flip**;
- every flip has a paired control, and the control must fire;
- every output class the pattern can emit is fired by a detect or a control
  (`pb_signs`, see rule 5);
- every declared condition is flipped or waived, totals must agree;
- `scripts/build.py check-mcdc` parses the indicator and requires
[...]

PR #220, opened 2026-08-15
test(candles): MC/DC for the six smallest single-bar patterns (#219 unit 1)

kevinlincg:
Unit 1 of the mechanical tier, as claimed on #219. Six single-bar patterns: **CDLDOJI, CDLBELTHOLD, CDLCLOSINGMARUBOZU, CDLLONGLEGGEDDOJI, CDLDRAGONFLYDOJI, CDLGRAVESTONEDOJI**.

```
MC/DC: 4 -> 10 function(s), 54 -> 67 condition(s), 48 -> 61 flipped, 6 waived
```

13 conditions, 13 flips, 13 controls, **no waivers** — a single-bar pattern has no cross-bar entailment, so every condition is independently falsifiable. If a waiver had been needed I would have left the condition unflipped and said so rather than assert a derivation I could not defend.

## Every flip sits on the comparison boundary

`pb_primer(12,100,2,4)` gives trailing bars with RealBody 2 and HighLow 10, which makes the thresholds exact:

```
avg(BodyLong)        = 1.0 * (20/10) / 1  = 2.0   exactly
avg(BodyDoji)        = 0.1 * (100/10) / 1 = 1.0   exactly
avg(ShadowVeryShort) = 0.1 * (100/10) / 1 = 1.0   exactly
avg(ShadowLong)      = the CURRENT bar's real body (avgPeriod 0)
```

`0.1*10.0` is exactly `1.0` in IEEE-754 and the library reaches it by the same multiplication — I ran both paths rather than assuming it. So where the library tests `>` or `<` the flip is placed **at equality**, and where it tests `<=` the paired **control** is. A relaxed comparison then fails a case instead of passing one, which is the mutation class your point 4 says the whole-unit-off flips could not catch.

## Proven, not assumed

Three mutations. Each was confirmed present in the **generated C** before its result was believed — editing `input/` does not prove the generated code moved, and I have had a literal silently normalised away twice this week with the gate staying green:

| mutation | artifact evidence | gate |
|---|---|---|
| `cdldoji` `<=` → `<` | `ta_CDLDOJI.c:141` becomes `<` | `CDLDOJI MC/DC FAIL bar=60 expected=100 got=0 (restore c0: body 1 == 1, inclusive)` |
| `cdlbelthold` `>` → `>=` | `ta_CDLBELTHOLD.c:156` becomes `>=` | `CDLBELTHOLD MC/DC FAIL bar=39 expected=0 got=100 (break c0: body 2 == avg 2)` |
| `cdldragonflydoji` shadows exchanged | guard flips `&& (inHigh[i]` → `&& (inLow[i]` | 4 failures: the detect and all three controls |

The middle one is worth noting for a second reason: my first attempt at it was a bad string replace that changed nothing, the gate stayed green, and only the artifact check caught that the sabotage had never applied. A green run there would have read as "the flip is solid".

The third is why DRAGONFLY and GRAVESTONE are in the same unit, and BELTHOLD next to CLOSINGMARUBOZU: each pair differs only in which shadow a given colour reads, so a model copy-pasted between them stays self-consistent while disagreeing with the library. The same trap exists inside DRAGONFLY, whose c1 and c2 read one average in opposite directions.

## One case worth its comment

`CDLLONGLEGGEDDOJI`'s second condition compares the shadows against `ShadowLong`, whose `avgPeriod` is 0 — so the threshold is the **current bar's real body**, not a windowed mean. At body 0 the threshold is 0, and the only bar that falsifies that condition alone is the degenerate `O=H=L=C` one. It is the single place in this unit where the threshold moves with the bar under test.

## Verification

`ta_regtest` passes end to end; `scripts/build.py check-mcdc` agrees for all ten builders; `check-source-lists` unchanged (no new files).

## On the tier count

My sweep finds 13 uncovered single-bar patterns, which matches your "13 are single-bar" but leaves my mechanical set one short of 14 — our conjunct definitions differ on the 14th somewhere. Now that the template is calibrated I can chase it down when I pick up the next unit, unless you already know which one you had in mind.

Not claiming the other seven mechanical functions yet — everything appends to `test_marquee_predicate_coverage()`, so I will take the next unit once this merges.

### DIFF
--- a/src/tools/ta_regtest/ta_test_func/test_candlestick.c
+++ b/src/tools/ta_regtest/ta_test_func/test_candlestick.c
@@ -1993,6 +1993,322 @@ static void build_3starsinsouth( void )
   pb_flat(8);
 }
 
+/* ---- Mechanical tier, unit 1: the six smallest single-bar patterns -------- *
+ *
+ * Every one of these decides on the CURRENT bar alone, so the primer is the
+ * whole of the averaging context and the scenario is a single bar appended to
+ * it. That makes the thresholds exact, which is the point:
+ *
+ *   pb_primer(12,100,2,4) lays down bars with RealBody 2 and HighLow 10, so
+ *
+ *     avg(BodyLong)         = 1.0 * (20/10) / 1  = 2.0   exactly
+ *     avg(BodyDoji)         = 0.1 * (100/10) / 1 = 1.0   exactly
+ *     avg(ShadowVeryShort)  = 0.1 * (100/10) / 1 = 1.0   exactly
+ *     avg(ShadowLong)       = the CURRENT bar's real body (avgPeriod 0)
+ *
+ *   0.1*10.0 is exactly 1.0 in IEEE-754 and the library reaches it by the same
+ *   multiplication, so a flip can sit ON the comparison boundary instead of a
+ *   safe distance from it. Every flip below does: where the library tests `>`
+ *   or `<` the flip is placed at equality, and where it tests `<=` the paired
+ *   control is placed at equality. A relaxed comparison therefore fails a case
+ *   rather than passing one, which is what the whole-unit-off flips in #219's
+ *   point 4 could not do.
+ *
+ * A single-bar pattern has no bar-to-bar coupling, so nothing here needs a
+ * waiver: all 13 conditions across the six are independently falsifiable and
+ * all 13 get a flip.
+ */
+
+/* CDLDOJI -- one condition: the body is no larger than the doji threshold.
+ *
+ *   c0  realbody(i) <= avg(BodyDoji)
+ *
+ * The control sits exactly at equality, which is what pins the comparison as
+ * inclusive: were it `<`, the control would stop firing.
+ */
+static void cond_doji( int i, int *c )
+{
+   c[0] = pb_body(i) <= pb_avg(TA_BodyDoji, i);
+}
+
+static void build_doji( void )
+{
+  pb_conditions(1);
+
+  pb_flat(6);
+  pb_primer(12,100,2,4);
+  int d=pb_bar(100,105,95,100);            /* body 0 */
+  pb_detect(d,100,"detect: body 0 <= 1");
+  pb_flat(8);
+
+  pb_primer(12,100,2,4);
+  int f0=pb_bar(100,105,95,102);           /* body 2 */
+  pb_flip(f0,0,"break c0: body 2 > 1");
+  pb_flat(8);
+  pb_primer(12,100,2,4);
+  int k0=pb_bar(100,105,95,101);           /* body 1 == avg */
+  pb_control(k0,100,0,"restore c0: body 1 == 1, inclusive");
+  pb_flat(8);
+}
+
+/* CDLBELTHOLD -- a long body whose OPENING side has no shadow to speak of.
+ *
+ *   c0  realbody(i) > avg(BodyLong)
+ *   c1  ( white(i) && lowershadow(i) < avg(ShadowVeryShort) )
+ *       || ( black(i) && uppershadow(i) < avg(ShadowVeryShort) )
+ *
+ * c1 is one condition, not two: the colour selects which shadow is read, so
+ * the disjunction is the atomic thing the decision tests. A white scenario
+ * falsifies the black arm by construction, so raising the lower shadow to the
+ * threshold falsifies c1 as a whole.
+ */
+static void cond_belthold( int i, int *c )
+{
+   double vs = pb_avg(TA_ShadowVeryShort, i);
+   c[0] = pb_body(i) > pb_avg(TA_BodyLong, i);
+   c[1] = (  pb_white(i) && pb_losh(i) < vs )
+       || ( !pb_white(i) && pb_upsh(i) < vs );
+}
+
+static void build_belthold( void )
+{
+  pb_conditions(2);
+
+  pb_flat(6);
+  pb_primer(12,100,2,4);
+  int d=pb_bar(100,108,99.5,105);          /* white, body 5, lower shadow 0.5 */
+  pb_detect(d,100,"detect: body 5 > 2, lower shadow 0.5 < 1");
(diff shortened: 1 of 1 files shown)


mario4tier:
Merged as e5a47812a, `--no-ff` and unsquashed so your commit stays checkoutable.

Everything this PR claims, I re-derived rather than read: all six condition models transcribe their `ta_codegen/input/<name>/<name>.c` faithfully (including `pb_white` ≡ `ta_candlecolor(c,o)==1` — both are `close >= open`), `check-mcdc` agrees on every conjunct count, and the totals land exactly where you said. Your boundary claim I checked with a mutation you *didn't* list, since the ones an author picks are the ones they had in mind: relaxing `CDLDRAGONFLYDOJI`'s c2 from `>` to `>=` in all three generated bodies fails at precisely the flip written for it —

```
CDLDRAGONFLYDOJI MC/DC FAIL bar=123 expected=0 got=100
  (break c2: lower shadow 1 == avg 1, the test is strict)
```

That is the class of mutation the whole unit exists to catch, and it catches it. The artifact-level discipline you describe — confirming the sabotage actually reached the generated C before believing a red or a green — is the right habit and it is what caught your own no-op string replace. Keep doing that.

## One gap, and it was the harness's, not yours

`CDLBELTHOLD` and `CDLCLOSINGMARUBOZU` are the first **bi-signed** patterns to enter the MC/DC set — their firing arm is `ta_candlecolor(...)*100`, so they emit both +100 and −100. Every scenario in the PR is white. Two mutations therefore survived the entire MC/DC tier green:

- `ta_candlecolor(...)*100` → `100` (the output sign never checked)
- the black arm of c1 deleted outright (half the decision never executed)

Both were caught, but only by the v0.6.4 freeze, which reaches just the 35 of 61 patterns that fire on the 252-bar series. For a bi-signed pattern among the other 26 there would have been no gate at all.

**Your modelling was correct.** `pb_conditions()` counts *top-level conjuncts* — that is what `check-mcdc` pins — so `(white && x) || (black && y)` is genuinely one condition, and declaring it as one was right. The consequence is that a disjunction's interior cannot be reached by flips *however many you write*: it needs a different axis. That is the harness's gap, and it was mine to fix, not yours.

Fixed in 1cf74d3ac, on top of your merge:

1. **`pb_signs(n)`** declares the pattern's output-class count (default 1). `pb_check_mcdc` now requires every declared class to be fired by a detect or a control, and `check_mcdc_conditions.py` derives the count from the pattern's firing arms and pins the declaration — exactly as it already pins `pb_conditions()`. So omitting `pb_signs(2)` on a bi-signed pattern fails the *static* check rather than silently covering half the decision.
2. **Black-arm detect + flip + control** for both patterns, on the boundary like yours (shadow exactly `1 == avg` at the flip, `0.5` at the control). BELTHOLD's black arm reads the **upper** shadow, CLOSINGMARUBOZU's the **lower** — the same exchange your white arms already exploit.
3. The `pb_*` header said *"for single-sign patterns"*. True when written, and the source of the assumption. Corrected.

Red then green: with (1) alone the gate fails `CDLBELTHOLD 1 of the pattern's 2 output class(es) exercised`; with (2) it is green at 12 detects / 63 flips / 63 controls. All three mutations now fail in the MC/DC tier itself instead of falling through to the freeze.

## And the same bug, one level down, in my own fix

Worth telling you because it lands directly in your path. Reviewing my own patch before pushing, `count_signs` turned out to return on the **first** firing arm. `CDLENGULFING`, `CDLHARAMI` and `CDLHARAMICROSS` each have **two** — `ta_candlecolor(...)*100` *and* `*80` — so they emit **four** classes. A future HARAMI builder would have declared `pb_signs(2)`, satisfied the check, and left the ±80 arm uncovered. I had written the gate against the one instance of the shape in front of me and assumed it covered the shape.

Enumerating all 61 settled it: **34 single-class, 22 bi-signed, 3 four-class**, plus the two Hikkakes, which emit a computed `patternResult` and fail closed if anyone ever writes an MC/DC builder for them.

One asymmetry is deliberate and commented at both sites: `count_signs`' regex is looser than `count_conditions`'. `CDLTRISTAR` writes `outInteger[outIdx]` with no `++`. Loosening it for *classes* is safe; loosening `count_conditions` is not, because TRISTAR's ±100 come from decisions **nested inside** the outer one, which the top-level-conjunct model does not describe — it returns 3 only by an accident of whitespace. Please don't "fix" that inconsistency.

## Your 13-vs-14

There is no missing 14th single-bar pattern — your sweep is right, and I confirmed it independently by parsing every `cdl*` detection expression for `[i-N]` references. **13 is the count.**

The table in #219 reads `mechanical | 14 | 56 | 13 are single-bar`. That last column is a *subset* note, not an equality: the 14th mechanical function is multi-bar by construction. You read it as an equality, which is a fair reading of a badly-worded cell, and I've reworded it.

While there, the `56` does not survive a parse either. Your 13 total **35 top-level conjuncts** (43 atomic) — neither convention reaches 56, and a 14th mechanical function would need 21 conditions to close the gap, which is hard-tier territory. That table is flagged in the issue as agent estimates rather than a parse, and this is that caveat cashing out. Plan against 35: unit 1 covered 13, leaving **22 across the remaining 7**.

## Green light for unit 2

Go ahead — the 7 remaining single-bar patterns:

| pattern | conjuncts | classes |
|---|---|---|
| CDLHIGHWAVE | 3 | **2** |
| CDLLONGLINE | 3 | **2** |
| CDLMARUBOZU | 3 | **2** |
| CDLSHORTLINE | 3 | **2** |
| CDLSPINNINGTOP | 3 | **2** |
| CDLRICKSHAWMAN | 4 | 1 |
| CDLTAKURI | 3 | 1 |

**Five of the seven are bi-signed** — that is why this was worth fixing before you started rather than after. Each needs `pb_signs(2)` and a firing scenario on both colours; `check-mcdc` will now tell you so up front if you forget. `git show 1cf74d3ac` has the two worked examples.

Pull `dev` before you branch.

PR #221, opened 2026-08-15
test(candles): MC/DC for the remaining seven single-bar patterns (#219 unit 2)

kevinlincg:
Unit 2 of the mechanical tier, on your green light. **CDLHIGHWAVE, CDLLONGLINE, CDLMARUBOZU, CDLSHORTLINE, CDLSPINNINGTOP, CDLTAKURI, CDLRICKSHAWMAN** — that completes the single-bar patterns.

```
MC/DC: 10 -> 17 function(s), 67 -> 89 condition(s), 12 -> 24 detect(s),
       63 -> 85 flipped, 85 control(s), 6 waived
```

22 conditions, 22 flips, 22 controls, no waivers. Five of the seven are bi-signed and carry `pb_signs(2)` with a firing scenario on each colour — `check-mcdc` confirms conditions **and** signs for all seventeen builders.

## Three thresholds unit 1 never touched

All exact under the same primer, checked by running the arithmetic rather than assuming it:

```
avg(BodyShort)      = 1.0 * (20/10) / 1 = 2.0
avg(ShadowShort)    = 1.0 * (80/10) / 2 = 4.0   <- Shadows-typed: divisor 2
avg(Near)           = 0.2 * (50/5)  / 1 = 2.0   <- avgPeriod 5, not 10
avg(ShadowVeryLong) = 2.0 * body(i)             <- avgPeriod 0
avg(ShadowLong)     = 1.0 * body(i)             <- avgPeriod 0
```

The `ShadowShort` divisor is worth a flag: a model that forgot it would place the threshold at 8 and agree with the library on every scenario whose shadows are under 4. LONGLINE's and SHORTLINE's flips sit exactly at 4 for that reason.

The two `avgPeriod == 0` settings move the threshold *with the bar under test*, so several flips must grow the shadows as they grow the body. HIGHWAVE's c0 flip raises the body to 2, which raises its own shadow threshold from 2 to 4; a bar that kept the detect's shadows would break three conditions instead of one.

## Mutations

Each confirmed present in the generated C before its result was believed.

| mutation | gate |
|---|---|
| `cdlhighwave` `ta_candlecolor(...)*100` → `100` | `CDLHIGHWAVE MC/DC FAIL bar=165 expected=-100 got=100 (detect black)` |
| `cdllongline` c1 `<` → `<=` | `CDLLONGLINE MC/DC FAIL bar=81 expected=0 got=100 (break c1: upper shadow 4 == avg 4, the test is strict)` |
| `cdlrickshawman` c3's `min(open,close)` → `max(open,close)` | `CDLRICKSHAWMAN MC/DC FAIL bar=186 expected=100 got=0 (restore c3: min 105 == band top, inclusive; max 106 differs)` |

The first is the mutation that survived unit 1's white-only scenarios. It now fails on the black detect.

## The third one failed on my first attempt, and that is the useful part

Every RICKSHAWMAN scenario I first wrote was a doji. c3 reads `min(open,close)` against the band's top and `max(open,close)` against its bottom — and with `open == close` those are the **same number**, so swapping one for the other was invisible. The mutation fell through to the v0.6.4 freeze, which reaches only 35 of the 61 patterns; for a pattern outside that 35 there would have been no gate at all.

The c3 pair now carries a real body of 1, with `min` exactly on the band's top (pinning the comparison as inclusive) and `max` above it, so the two terms disagree and reading the wrong one stops the control firing.

I mention it at length because it is the same shape as the gap you found in unit 1: a scenario that is *degenerate along the axis the condition actually discriminates*. White-only hid a colour disjunction; open == close hid a min/max pair. Worth a line in the harness notes for whoever writes the moderate tier — I'd suggest it as "make the scenario non-degenerate along whatever the condition compares", but it is your file and I have not touched the header.

## Verification

`ta_regtest` end to end; `check-mcdc` for conditions and signs on all seventeen; `check-source-lists` unchanged (no new files). Nothing outside `test_candlestick.c` is touched.

Not claiming the moderate tier yet.

### DIFF
--- a/src/tools/ta_regtest/ta_test_func/test_candlestick.c
+++ b/src/tools/ta_regtest/ta_test_func/test_candlestick.c
@@ -2415,6 +2415,500 @@ static void build_gravestonedoji( void )
   pb_flat(8);
 }
 
+/* ---- Mechanical tier, unit 2: the remaining seven single-bar patterns ----- *
+ *
+ * Same primer as unit 1, pb_primer(12,100,2,4) -- RealBody 2, HighLow 10 -- and
+ * the same discipline: every flip on the comparison boundary. Three more
+ * thresholds come into play here, and all three are exact under that primer
+ * (checked by running the arithmetic, not by assuming it):
+ *
+ *   avg(BodyShort)       = 1.0 * (20/10) / 1   = 2.0
+ *   avg(ShadowShort)     = 1.0 * (80/10) / 2   = 4.0   <- Shadows-typed, so the
+ *                                                         divisor is 2, not 1
+ *   avg(Near)            = 0.2 * (50/5)  / 1   = 2.0   <- avgPeriod 5, not 10
+ *   avg(ShadowVeryLong)  = 2.0 * body(i)       (avgPeriod 0)
+ *   avg(ShadowLong)      = 1.0 * body(i)       (avgPeriod 0)
+ *
+ * The two avgPeriod-0 settings are the reason several flips below cannot be
+ * built by moving a shadow while holding the body: the threshold moves with
+ * the bar. Where that bites, the comment says so.
+ *
+ * Five of the seven are BI-SIGNED and carry pb_signs(2) with a firing scenario
+ * on each colour. Unit 1 shipped two bi-signed patterns with white-only
+ * scenarios, which left the -100 class and half of a disjunction unexercised;
+ * that gap is what pb_signs() now refuses to let happen silently.
+ *
+ * No waivers: single-bar patterns have no cross-bar entailment, so all 22
+ * conditions are independently falsifiable.
+ */
+
+/* CDLHIGHWAVE -- a small body with two very long shadows.
+ *
+ *   c0  realbody(i)    <  avg(BodyShort)
+ *   c1  uppershadow(i) >  avg(ShadowVeryLong)      = 2 * realbody(i)
+ *   c2  lowershadow(i) >  avg(ShadowVeryLong)
+ *
+ * c1 and c2 share a threshold that is a multiple of the CURRENT body, so the
+ * c0 flip has to grow the shadows as it grows the body: at body 2 the shadow
+ * threshold is 4, and a bar that merely kept the detect's shadows would break
+ * three conditions at once instead of one.
+ */
+static void cond_highwave( int i, int *c )
+{
+   double vl = pb_avg(TA_ShadowVeryLong, i);
+   c[0] = pb_body(i) < pb_avg(TA_BodyShort, i);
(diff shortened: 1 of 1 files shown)


mario4tier:
Merged as 90600d833. Everything the PR claims checks out: the seven condition models transcribe their `input/<name>/<name>.c` conjunct by conjunct, the five threshold values are exact against `ta_common/ta_global.c` (including the `Shadows` divisor of 2 and `Near`'s avgPeriod 5), and all three of your mutations reproduce with the failure text you quoted, verbatim.

I ran four more you didn't list. Three are caught:

```
MARUBOZU c2    <  -> <=    break c2: lower shadow 1 == avg 1
SPINNINGTOP c0 >  -> >=    break c0: upper shadow 1 == body 1
TAKURI c0      <= -> <     restore c0: body 1 == 1, inclusive
```

The TAKURI one is the shape worth naming: an inclusive boundary can't be pinned by a flip at all — there's no minimal value above a `<=` threshold, so the flip always sits a whole unit clear and says nothing. The control does it, sitting exactly on the equality. You put one there in both patterns that need it.

**The fourth escaped, and it was our bug, not yours.** Relaxing CDLRICKSHAWMAN's band from `max(open,close) >= mid - Near` to `>` left the whole tier green — 89 conditions, 85 flips, every one passing. Your builder was correct and consistent with the harness as documented; `check-mcdc` was not. It counted `&&` only at the detection expression's own depth, so the parenthesised `(min... && max...)` group read as one condition, and pinning the upper edge declared the whole thing covered.

That's right for a disjunction — BELTHOLD's `(white && x) || (black && y)`, where a disjunct can't be falsified while its sibling holds, which is why the output class became its own axis. It's simply wrong for a conjunction: `A && (B && C)` is `A && B && C` and the parens carry no meaning. So the fix is in the counter rather than in your builder (7ecabee9b), and the corpus says it's cheap — of all 61 patterns, RICKSHAWMAN is the **only** one whose count changes, 4 to 5. BELTHOLD correctly stays at 2. Totals are now 90 conditions / 86 flips / 86 controls, and a builder that groups two comparisons from here on fails the static check instead of banking a silent hole.

Your suggested wording is in the harness header as the rule, unedited: *a scenario must be non-degenerate along whatever the condition compares*. Two instances of one shape now — white-only bars leaving a colour disjunction unreached, and `open == close` making `min` and `max` the same number. Neither moves the totals, which is exactly why it needed naming there rather than in whichever builder hits it next.

**The moderate tier is yours if you want it** — you've earned the call on that, and the harness is in better shape for it than when you started unit 2. Equally fine if you'd rather not: we'll start working through them on Sunday, New York time, so there's no gap either way. Just say which so we don't both write the same builders.

PR #231, opened 2026-08-17
test(candles): MC/DC for the seven moderate-tier candlesticks (#219)

kevinlincg:
The moderate tier is done: CDLSEPARATINGLINES, CDLKICKING, CDLKICKINGBYLENGTH, CDL3INSIDE, CDLSTICKSANDWICH, CDLXSIDEGAP3METHODS and CDLCONCEALBABYSWALL. Sixty builders now, 380 conditions, 437 flips, 376 controls, 20 waivers; `check-mcdc` agrees with the pattern source on all sixty.

Per my note on the issue, this is reported per pattern, and every claim below is from a `ta_regtest` run rather than from my own scripts. Each condition got a mutation probe applied to **all four tiers** of the generated source, and every probe is reported with the site count so an anchor that matched nothing is visible as such rather than passing for green.

## Three things worth reading, none of them about coverage counts

**CDLXSIDEGAP3METHODS has a coupled pair, and the gate refused my first attempt at it.** c0 (1st and 2nd same colour) and c1 (3rd opposite) are the same predicate once the other five hold: c3, c6 and c4 chain into `open(i) > bodylo(i-1) > bodyhi(i-2) > close(i)`, so the third candle is black by construction, and the black arm mirrors that and forces it white. Either way `color(i) == -color(i-2)`, which turns c1 into `color(i-1) == color(i-2)` — c0 again. I gave c0 a flip anyway and the gate said it broke two conditions. It was right; a waiver on c1 is not a licence to damage it in passing.

So both are waived, and that is a description of the source rather than a hole: deleting **either** conjunct preserves the function exactly, and the suite confirms it — deleting c0 alone leaves all of `ta_regtest` green including the 911 frozen v0.6.4 values, and so does deleting c1 alone. Deleting **both** does change the function, so there is one zero-valued `pb_expect` case (2nd candle black, everything else intact) that exists only to stop the pair vanishing unnoticed. It makes no independence claim. With both conjuncts removed it reports 100 where it must report 0.

**CDLCONCEALBABYSWALL caught me with a green gate that was blind.** Thirteen conditions, four bars, and the settings average read at three lags off the rolling per-bar array. My first draft gave the 1st and 2nd candles the primers' own HighLow range, which put all three averages at exactly 1.0. Every flip pinned, every condition covered, gate green — and then I rewired c6/c7 to read the `[3]` window and c9 to read the `[2]` one, eight and four call sites, and **the whole suite stayed green**.

Equal thresholds make the wiring unobservable no matter how exactly each flip is pinned. Giving those two candles a range of 35 against the primers' 10 separates them to 1.0 / 1.25 / 1.5, all dyadic — which matters, because a shadow is a difference of two prices and has to land on the same double the library computes, so a threshold that merely prints as a short decimal is not enough. The c6 and c7 controls now sit at 1.0, deliberately the neighbouring window's value. Both mis-wirings are red.

There is a small rule in that: for a `<` condition the **control** catches the mix-up, since the flip lands false either way; for a `>` condition it is the **flip**, since it fires against the smaller neighbour. That is why `check_candle_windows.py` earns its keep — this is only observable dynamically when the windows carry different values, and nothing forces a builder to arrange that.

**CDLKICKINGBYLENGTH's decision is byte-identical to CDLKICKING's**, all eight conjuncts, same source text. No condition scenario can separate them; the separation is the emitted value and lives on the `pb_signs` axis. Every scenario keeps the first body at 12 against the second's 5, so the longer candle is always the first and the two report opposite signs on the same bars. Reusing KICKING's builder with every expectation negated and staying green is itself the evidence that the difference is only the sign.

## Per-pattern notes

- **CDL3INSIDE** — first pattern whose disjuncts carry three terms. The third candle's colour and its close-out are not independent by construction, so those two flips sit at different price levels rather than being one bar edited two ways. c3's flip shortens the first body, which is inside c4's window, so c4's threshold moves under it; the second body is kept far below either value so the move cannot break c4 as collateral.
- **CDLSTICKSANDWICH** — `avg(Equal)` is read at i-2 over five bars, so its window holds only primers and none of the three pattern candles. That is what lets the c0 flip turn the first candle white while leaving its close at 100. The three colour flips are dojis: the library's test is `close >= open`, so `open == close` is the minimal violation of "black". Four of its six probes land on a control rather than a flip.
- **CDLSEPARATINGLINES / CDLKICKING** — as in the earlier units; the colour that selects the arm also sets the output sign, so the second arm and the second class are reached together.

## Not claimed

- Java and C# are not verified end to end by me — this box has JDK 11 and .NET 6/7, so my runs are `--language=c,rust`. This change is test-only and touches no generated artifact (`git diff --stat` against dev is one file), so the exposure is that the new cases are C-only until CI runs them.
- One flip in CDLXSIDEGAP3METHODS is **not** pinned to an edge, and the comment says so. The colour conditions there have no boundary to sit on: the minimal white candle is a doji, but c2 and c3 need the 2nd real body wide enough to hold `open(i)` strictly inside it, so a degenerate body breaks those first.

### DIFF
--- a/src/tools/ta_regtest/ta_test_func/test_candlestick.c
+++ b/src/tools/ta_regtest/ta_test_func/test_candlestick.c
@@ -9970,6 +9970,167 @@ static void build_kicking( void )
   pb_control(k7b,-100,7,"restore c7 alt1: high 92.5 < prior low 93.5");
   pb_flat(8);
 }
+static void cond_kickingbylength( int i, int *c )
+{
+   double vs1 = pb_avg(TA_ShadowVeryShort, i-1);
+   double vs0 = pb_avg(TA_ShadowVeryShort, i);
+   c[0] = pb_white(i-1) != pb_white(i);
+   c[1] = pb_body(i-1) >  pb_avg(TA_BodyLong, i-1);
+   c[2] = pb_upsh(i-1) <  vs1;
+   c[3] = pb_losh(i-1) <  vs1;
+   c[4] = pb_body(i)   >  pb_avg(TA_BodyLong, i);
+   c[5] = pb_upsh(i)   <  vs0;
+   c[6] = pb_losh(i)   <  vs0;
+   c[7] = ( !pb_white(i-1) && pbL[i] > pbH[i-1] )
+       || (  pb_white(i-1) && pbH[i] < pbL[i-1] );
+}
+
+/* The scenario the two KICKINGs share. CDLKICKING reports this candle's
+ * colour, CDLKICKINGBYLENGTH the colour of whichever body is longer, so a
+ * builder whose longer body is always the current one cannot tell them apart.
+ * Every scenario below keeps the FIRST body longer (12 against 5), which is
+ * what makes the two functions disagree and gives each its own expected value.
+ */
+/* c7's alternatives are a colour selector plus a gap test. The selector is
+ * waived as in BELTHOLD; the gap test is the term the boundary flip attacks. */
+static void arm_kickingbylength( int i, int cond, int arm, int *a )
+{
+   if( cond != 7 ) return;
+   if( arm == 0 ) { a[0] = !pb_white(i-1); a[1] = pbL[i] > pbH[i-1]; }
+   else           { a[0] =  pb_white(i-1); a[1] = pbH[i] < pbL[i-1]; }
+}
+
+static void build_kickingbylength( void )
+{
+  pb_conditions(8);
+  pb_signs(2);
+  pb_arm(7,0,2); pb_arm(7,1,2);
+  pb_arm_model(arm_kickingbylength);
+  pb_waive_arm(7,0,0,"the arm's own colour selector -- it chooses the arm, and the class it chooses is fired by pb_signs(2)");
+  pb_waive_arm(7,1,0,"the arm's own colour selector -- it chooses the arm, and the class it chooses is fired by pb_signs(2)");
+
+  pb_flat(6);
+  pb_primer(12,100,2,4);
+  pb_bar(106,106.5,93.5,94);               /* 1st: black marubozu, body 12, shadows 0.5 */
+  int d=pb_bar(107,112.5,107,112);       /* 2nd: white marubozu, body 5, gaps up */
+  pb_detect(d,-100,"detect: opposite marubozus, 2nd gaps up over 106.5");
+  pb_flat(8);
+
+  pb_primer(12,100,2,4);
+  pb_bar(106,106.5,93.5,94);
+  int f0=pb_bar(112,112.5,107,107.5);      /* BLACK like the 1st, low 107 keeps the gap */
+  pb_flip(f0,0,"break c0: both candles are black");
+  pb_flat(8);
+  pb_primer(12,100,2,4);
+  pb_bar(106,106.5,93.5,94);
+  int k0=pb_bar(107,112.5,107,112);
+  pb_control(k0,-100,0,"restore c0: opposite colours");
+  pb_flat(8);
+
+  pb_primer(12,100,2,4);
+  pb_bar(96,96.5,93.5,94);                 /* 1st body 2 == avg at i-1 */
+  int f1=pb_bar(107,112.5,107,112);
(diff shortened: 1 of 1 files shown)


mario4tier:
Merged to dev as 715beda72, with follow-up fixes in d12e3a610.

Three things were adjusted on top rather than sent back:

**c5 and c6 in both KICKINGs were not on the boundary they documented.** The threshold is `avg(ShadowVeryShort, i)` = 1.03, but the library compares a subtraction of two prices, not the literal — and `113.03 - 112` is 1.0300000000000011, five ULP above, because 112 + 1.03 is not representable. Both `<` and `<=` were false, so the flip reported 0 either way; relaxing those two tests to `<=` passed the entire suite. The first bar's range moves 13 -> 13.125, which puts the average on 1.03125 and makes both flips bitwise exact. The body stays 12, so `avg(BodyLong, i)` stays 3.0 and c4 is untouched.

The header comment argued no dyadic threshold existed for this pattern. One does — that is where 13.125 comes from — and the comment now says so, since the derivation is what the next unit reads.

**CDLKICKINGBYLENGTH's length comparison was never reached on its true branch.** Every scenario kept the first body longer, so the selector always resolved to `i-1`; deleting the comparison outright and reporting `i-1`'s colour unconditionally also passed the whole suite. One detect with the second body longer covers it.

**c7's second arm** now flips and controls on the same bars, one price apart, instead of the control reverting to the detect bars.

Everything else held up under mutation: all 52 declared conditions go red when deleted except the two XSIDEGAP3METHODS waivers, and deleting *both* of those goes red exactly as your `pb_expect` case claims; the CONCEALBABYSWALL window rewiring is caught on 6 and 3 call sites; the five inclusive sites are each pinned by a control sitting on the edge. The coupled-pair derivation and the window-separation argument both check out.

Java and C# are now verified end to end — all four languages 161 passed / 0 failed.

CDL3OUTSIDE landed in the same follow-up, which completes the tier at 61 of 61.

kevinlincg:
Reproduced all three before replying. Two of them are mine and the first one is a repeat.

**c5/c6 — I verified one side of the comparison.** I had checked that the threshold lands on the literal: `0.1*(90+13)/10 == 1.03` is true, so I called the flip pinned. What the library compares against it is `113.03 - 112`, and that is `1.0300000000000011`, five ULP up, because 112 + 1.03 is not representable. Strict `<` was already false, so the case reported 0 for a reason that had nothing to do with the boundary — the tell being exactly what you found, that `<=` passes too.

The rule I had written for myself was "a flip is pinned iff the literal I write and the double the library computes are bitwise equal." That is right and I applied it to the threshold alone. The comparison has two sides and the other one is a subtraction of two prices, which is where the representability goes. The check that would have caught it is one line:

```
shadow  = 113.03 - 112 = 1.0300000000000011
avg     = 0.1*((90+13)/10) = 1.03
equal   = False          <- the flip is not on the boundary
```

And the header comment claiming no dyadic threshold exists was wrong on top of that. `13.125` gives `1.03125` on both sides exactly. I have now claimed "no solution exists" three times in this issue and been wrong every time; in each case the search space I had explored was narrower than the one I described. I will state the bound I actually searched rather than the conclusion I drew from it.

**CDLKICKINGBYLENGTH — I wrote the fact down and did not draw the consequence.** My own commit message says every scenario keeps the first body at 12 against the second's 5, so the longer candle is always the first. I used that to argue the two functions must report opposite signs, which is true, and never asked what then covers the selector's other branch. Nothing did — deleting the comparison passes, as you showed. "No condition scenario can separate them" was the right observation and the wrong stopping point: the next question is which axis covers each thing the decision reads, and the length comparison was on none of them.

**c7's second arm** — agreed, a control that reverts to the detect bars proves the pattern still fires, not that the flipped quantity is what stopped it.

Thank you for taking these on rather than sending them back, and for running the mutation sweep across all 52 conditions — the two XSIDEGAP3METHODS waivers behaving as the `pb_expect` case claims is the part I most wanted an independent check on, since the whole coupled-pair argument rests on it.

Good to see all four languages green end to end, and 61 of 61.

Issue #225, opened 2026-08-16
C: partial buffer overlap is accepted and silently corrupts output


mario4tier:
The output-distinctness reject (#108) compares buffers for **equality**, not for **overlap**:

```c
/* src/ta_func/ta_BBANDS.c */
if( outRealUpperBand == outRealMiddleBand || ... ) return TA_BAD_PARAM;
```

Equality is a complete distinctness test only while the expressible cases are "same buffer" and
"disjoint buffers". With raw pointers a caller can also pass two views of the same memory at
different offsets:

```c
TA_BBANDS(0, n-1, in, 5, 2.0, 2.0, TA_MAType_SMA, &beg, &nb,
          out, out + 1, out2);      /* overlapping, not equal */
```

That is accepted, computes, and returns `TA_SUCCESS` with wrong values. No error, no warning.

Beyond the missed reject, three bodies branch on buffer identity as **algorithm** rather than as
validation, and take the wrong arm when the buffers do collide:

- `ta_BBANDS.c` — the scratch election `if( inReal == outRealUpperBand )`. On a partial overlap
  it elects a temp buffer that collides with its own input, then writes through what it is
  still reading.
- `ta_MAVP.c` — the in-place defence `if( outReal == inReal )`.
- `ta_STOCH.c` / `ta_STOCHF.c` — the same shape. These currently misfire in the *safe*
  direction (they fall back to allocating), but are one edit away from the same class.

## Only C is affected

Decidable from each language's type system rather than by testing:

| backend | expressible? | why |
|---|---|---|
| **C** | **yes** | raw pointers; `out + 1` is legal |
| Rust | no | two overlapping `&mut [f64]` cannot exist in safe Rust, and the crate is `#![forbid(unsafe_code)]` |
| Java | no | two array references are identical or disjoint, with nothing in between |
| C# | no longer | the span surface made it expressible; already fixed, guarded and sabotage-proven in the C# tier |

Worth stating in the fix so that nobody later "adds the missing check" to Rust or Java, where it
is unreachable by construction.

## Why no test catches it

This is the part worth doing first, because the defect is a symptom of it.

**Nothing in the tree tests overlap in the batch tier.** Every aliasing probe passes *whole,
identical* buffers, where equality and true overlap behave alike — so no existing probe can
distinguish a correct guard from a broken one:

- the C-tier alias probes in `server_gen.rs` — whole `input_arrays[0]` / `fbuf[0]`
[...]

PR #233, opened 2026-08-18
fix(c): reject overlapping buffers at the produced extent (#225)

kevinlincg:
Implements the ruling. Two rejects from the single site at `c.rs`, measured at the produced extent.

## The extent

```
first    = max(startIdx, TA_<NAME>_Lookback(<validated optIns>))
produced = (first > endIdx) ? 0 : endIdx - first + 1
outExt   = produced * sizeof(out elem)
inExt    = (endIdx + 1) * sizeof(in elem)
```

Emitted as one block after the optIn validation, so the lookback sees validated values. All four amendments are in, and the emitter comment records each one with the failure it prevents — a later reader has to actively decide to reopen them rather than tidy them away.

## Evidence

Red half first, on `dev` before any fix — the probe sits beside `verify_accbands_inplace_aliasing()`, no new file:

```
OVERLAP PROBE: out/out overlap accepted (rc=0)
OVERLAP PROBE: in/out overlap accepted (rc=0)
OVERLAP PROBE: out = in + startIdx accepted (rc=0) -- reads [71..299], writes [100..299]
```

With the fix:

```
Buffer-overlap rejects: 3 colliding shapes refused, whole-buffer in place
                        accepted and bit-identical
Stream verify: 174 functions, 20200 legs bit-exact vs batch
C: 161 passed, 0 failed
```

The probe pins the accepted side as hard as the rejected one: whole-buffer in place must stay `TA_SUCCESS` **and** bit-identical to the disjoint result, and the empty range (`TA_SMA(0, 5, buf, 30, ..., buf + 2)`) must stay `TA_SUCCESS`/nb=0. Those are the amendment-2 and false-reject cases.

**Reverting sabotage**, the red half of record: dropping the in/out rule from the emitter and regenerating reopens 2 of the 3 shapes, including `out = in + startIdx`. That is the coupling in amendment 4 demonstrated rather than asserted.

`git diff` shows the Rust, Java and C# outputs unchanged and the `TA_S_` bodies unchanged; the float variant carries no reject, since a `const float *` input cannot alias a `double *` output.

## What I got wrong on the way, since it shaped the patch

I applied your §2 "do not couple to `TA_X_Lookback()`" to the output side as well, which is where it is load-bearing, and then read the resulting false rejects as a property of internal composed calls. I built a census of those calls — 40 sites, 22 passing the caller's own output inward, 8 false-rejectable — and offered you four structural options, all of which the correct extent dissolves. The census was real work aimed at the wrong question.

Two of the four options I offered were also unsound on their own terms, and I withdrew them in-thread: "land out x out first" was based on an isolation run I had not repeated after the build drifted, and is dead anyway under amendment 4.

## Not in this PR

- **`stoch.c`'s error path**, reported separately in-thread: the `memmove` at line 363 runs before the `retCode` check at 369, so a failing inner call copies through a stale `*outNBElement`. Measured — a sentinel past the caller's buffer is overwritten. Unreachable today because that inner `TA_MA` has never failed; my first attempt made it fail and that is how it surfaced. It is a transcribed body and you ruled those off-limits here, so it is untouched and awaiting your call on whether it belongs to this issue, and whether the fix is moving the copy or zeroing the out-params in the reject preamble.
- **Open / Update / Peek** carry no reject, per your §3. The streaming page now states that explicitly instead of leaving it to be inferred from `OpenAndFill` being stricter.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -45,6 +45,7 @@ See [github commits](https://github.com/TA-Lib/ta-lib/commits) for complete list
 - ~10%: ATR and NATR
 
 ### Changed
+- (#225) C functions now return `TA_BAD_PARAM` when two output buffers overlap, or when an output *partially* overlaps an input. Computing wholly in place is unchanged and still supported; only partial overlap is refused, which previously corrupted the result silently. Calls that succeeded before and now fail were reading and writing the same memory at different offsets.
 - (#133) BBANDS default `optInTimePeriod` changed from 5 to 20, as intended by John Bollinger.
 - (#120) PPO and APO now default `optInMAType` to EMA (was SMA), matching Gerald Appel's original PPO/MACD definition. Pass `TA_MAType_SMA` explicitly to keep the previous behavior.
 - (#96) Fused multiply-add and other floating-point re-ordering produce minor output differences; an intentional modernization.
--- a/src/ta_func/ta_AC.c
+++ b/src/ta_func/ta_AC.c
@@ -128,6 +128,15 @@ TA_LIB_API TA_RetCode TA_AC( int    startIdx,
       return TA_BAD_PARAM;
    if( !outReal )
       return TA_BAD_PARAM;
+   {
+      int taFirst = (int)TA_AC_Lookback(optInFastPeriod,optInSlowPeriod,optInSignalPeriod);
+      int taProduced;
+      if( taFirst < startIdx ) taFirst = startIdx;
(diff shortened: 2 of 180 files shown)


mario4tier:
Closing unmerged — rationale on #225 (https://github.com/TA-Lib/ta-lib/issues/225#issuecomment-5331129513). Not a reflection on the work; the problem turned out to cost more than it is worth.

Issue #229, opened 2026-08-16
perf(streaming): trailing rings buffer every raw input array when only a derived per-bar scalar is read back (368 redundant rings across 63 functions)


mario4tier:
**This is an optimization, not a defect.** No output value is affected, no gate is
failing, and nothing here is user-visible except memory footprint. Filing it so the
measurement is not lost.

## What happens

When a batch body reads a trailing window, the stream emitter synthesises one ring
buffer **per input array read through that trailing index** (`RingSpec.arrays`,
`streaming.rs`), plus a mirror for each. That is correct and necessary when the raw
arrays are what the loop reads back.

But in most cases the loop does not read those arrays back individually — it reads
back a **single derived scalar that is a pure function of the bar**. The emitter does
not analyse the shape of the trailing read: it records *which arrays* the expression
touches and synthesises one ring per array. Nothing today asks whether those arrays
are only ever consumed by a single derived expression. The information is already in
the IR; the question is simply never put to it.

AO is the smallest example. Its steady loop subtracts

```c
sumFast -= (inHigh[trailingFastIdx] + inLow[trailingFastIdx]) / 2.0;
sumSlow -= (inHigh[trailingSlowIdx] + inLow[trailingSlowIdx]) / 2.0;
```

and the emitted handle carries eight buffers:

```c
struct TA_AO_Stream {
   ...
   double *ring_trailingFastIdx_inHigh;
   double *ringMirror_trailingFastIdx_inHigh;
   double *ring_trailingFastIdx_inLow;
   double *ringMirror_trailingFastIdx_inLow;
   double *ring_trailingSlowIdx_inHigh;
   double *ringMirror_trailingSlowIdx_inHigh;
   double *ring_trailingSlowIdx_inLow;
   double *ringMirror_trailingSlowIdx_inLow;
};
```

Only `(high+low)/2` is ever read back. One ring of `max(fast,slow)` medians, shared
by both trailing indices, would carry the same information: at the 5/34 defaults that
is **1296 → ~592 bytes and 8 allocations → 2**, and `Update` stops recomputing the
median for two trailing bars.

The candlesticks are the same shape and far worse: each trailing total subtracts one
`TA_CANDLERANGE(...)` value, but the macro touches all four OHLC arrays, so every
trailing index buffers four rings plus four mirrors to recover one scalar per bar.

## Scope

Census over the emitted `struct TA_*_Stream` definitions in `src/ta_func/`:
[...]

PR #234, opened 2026-08-18
perf(stream): collapse the candlestick derived rings, on one shared walk (#229)

kevinlincg:
Two commits: the structural item you named as the root of both defects you fixed, then the tranche it unblocks.

## One walk, not five

`derived_bar_value` in `streaming.rs` and the `reindex` nested in each of the four stream backends were the same traversal, differing only in what an `in<Array>[...]` becomes — the scalar update parameter for the per-bar store, the same array re-indexed to the fill counter for the open-time fill. That leaf mapping is now the parameter and the walk exists once.

`generate` is byte-identical for every backend after this commit, which is the whole proof that it changed nothing.

I did **not** hoist the fill loop itself into IR. The body is renderable by every backend already, but its operands are named differently at capture time than in the transition — C reads `sp->ringCap_v` where Rust reads a local `cap_v` — so it needs a capture-time `NameMap`. That is a second change, and folding it in here would have cost the byte-identical proof above. Happy to do it next if you want it before the window tranche.

## The candlestick rings

Lifting `mentions_candle_settings`, now that the backends handle the lagged form:

```
census: total saving 365 -- candlesticks 360 over 120 pairs, other 5
```

| function | ring pointers |
|---|---|
| CDLADVANCEBLOCK | 40 → 10 |
| CDLHAMMER | 32 → 8 |
| CDL3WHITESOLDIERS | 32 → 8 |
| CDLDOJI | 8 → 2 |

57 files, +1847/−6608.

The 49 lagged rings are reached for the first time here, since they are all candlesticks. Their fill emits `fillJ % cap`, the absolute-mod slot — I checked the generated Rust rather than trusting the emitter, because the linear form would be green on the C leg and wrong on the rotation phase.

## Ordering, and why I inverted yours

You listed per-shape slots before windows. I did neither yet, because per-shape slots buy nothing on the current corpus — VWMA is 2 arrays over 2 shapes and ACCBANDS 3 over 7, so both stay at `saving = 0` either way — which matches your own "invisible today; it bites on windows". The hold-back was the gate holding the largest tranche, and the semantic question behind it was settled in your merge comment, so I took that first. Per-shape slots land with the window work, where they are load-bearing.

## Verification

```
Stream verify: 174 functions, 20200 legs bit-exact vs batch
C: 161 passed, 0 failed
generated Rust crate: compiles clean
clippy: clean
```

**Not claimed:** Java and C# are read, not run — JDK 11 and .NET 6/7 here, the same limit as before. That is exactly the gap that hid the `back > 0` defect in my last PR, so I would rather say it plainly than let the C leg speak for four.


kevinlincg:
Strengthening one claim in the PR body, because the way I originally checked it was too weak to carry it.

I wrote that the lagged fills emit the absolute-mod slot, having read one generated Rust file. That is a sample, not a check, and the defect this PR builds on was precisely a per-backend divergence that a sample would miss.

Cross-backend, over every emitted derived fill:

```
C:    49 mod occurrences over 8 distinct vars, 76 linear over 12
Rust  mod-set == C's: True   linear-set == C's: True   counts 49/76
Java  mod-set == C's: True   linear-set == C's: True   counts 49/76
C#    mod-set == C's: True   linear-set == C's: True   counts 49/76
```

Comparing the *sets of variables* that take each form, rather than counts alone, is what makes this discriminating: a backend that fell through to the wrong arm for some rings would keep the totals plausible while moving vars between the sets. The 49 also matches your count of lagged rings independently.

Two false starts worth recording, since both produced confident-looking output:

- my first detector paired each fill against a `ringLag_<var>` marker searched **per file**. That works for C and C#, one function per file, and is meaningless for Java, where `Core.java` holds every function — so a marker belonging to another function made 47 correct fills look mismatched;
- for Rust it reported 49 mismatches for the opposite reason: the marker string I guessed does not exist in that backend at all, so every mod-form fill read as "should have been linear".

Neither was a finding about the code. I mention them because both printed a tidy count that would have been quotable, and the only thing separating them from the real result is that the real one does not depend on a name I guessed.

PR #275, opened 2026-08-28
perf(rust): a stream handle carries only the candle settings its step reads (#274)

### LINKED ISSUE #274: Rust optimization: a stream handle clones the whole 280-byte Core, of which 119 of 176 functions read nothing
Every generated Rust stream handle embeds a full `Core` by value:

```rust
pub struct SMA_Stream { core: Core, state: SMA_StreamState, out: OutRange }
...
Ok(SMA_Stream { core: self.clone(), state, out: ... })
```

`Core` is **280 bytes** (`[i32; 24]` unstable periods = 96, `CandleSettings` = 176, `Compatibility` = 1, padding). Almost none of it is reachable after the handle is built.

## What a handle actually reads

Measured over all 176 generated `_step_impl` bodies:

| reads | count |
|---|---|
| nothing from `Core` | **119 / 176** |
| `candle_settings` only | 57 / 176 |
| `unstable_period` | **0** |
| `compatibility` | **0** |

Of the 57 candlestick steps: 15 read 1 distinct `CandleSetting`, 28 read 2, 8 read 3, 5 read 4, 1 reads 5. A `CandleSetting` is 16 bytes, so the worst case needs **80 of the 280 bytes** and the median needs 32.

The unstable period and compatibility mode are consumed at `Open` (they set the lookback); nothing post-open consults them. The only two things a handle's `core` is used for are `self.core.clone_from(&src.core)` in `restore_from`, and as the receiver of its **own** `<N>_step_impl` (2 sites per file: `update` and `update_and_fill`). Composed handles reach their sub-handles through the sub's public `update`, not through a sub `_step_impl` — so there is no cross-function settings forwarding to arrange.

## Measured cost

Handle sizes today (`size_of`, x86-64):

| handle | bytes | of which `Core` |
|---|---|---|
| `EMA_Stream` | 320 | 280 (87%) |
| `ATR_Stream` | 320 | 280 |
| `SMA_Stream` | 352 | 280 (80%) |
| `CDLDOJI_Stream` | 344 | 280 (81%) |
| `STDDEV_Stream` | 712 | 560 — 2 nested |
| `MA_Stream` | 1136 | — |
| `STOCHRSI_Stream` | 2208 | — |
| `BBANDS_Stream` | 2168 | ~1400 — 5 nested |
| `MACDEXT_Stream` | 3720 | ~1960 — 7 nested |

Nesting multiplies it: `BBANDS_Stream` holds `MA_Stream` + `STDDEV_Stream`, `STDDEV_Stream` holds `VAR_Stream`, and `MA_Sub` is an enum over ten handle variants that each carry their own `Core` — so the dispatch pays the largest variant's copy whichever MA type is selected.

## Prototype result

Hand-patched EMA, SMA, CDLDOJI, STDDEV and VAR in a private copy of the crate (handle stores only the `CandleSetting`s its step reads; `_step_impl` loses its `&self` receiver). A checksum over 1000 peeks across five handles was **bit-identical** to the unpatched crate.

Size:

| handle | before | after |
|---|---|---|
| `EMA_Stream` | 320 | **40** (−87%) |
| `SMA_Stream` | 352 | **72** (−80%) |
| `CDLDOJI_Stream` | 344 | **80** (−77%) |
| `STDDEV_Stream` | 712 | **152** (−79%) |
| `BBANDS_Stream` | 2168 | **1608** (−26%, only its STDDEV/VAR/SMA legs were patched) |

Speed — best-of-7 per run, 4 interleaved runs of each binary, bands quoted:

| benchmark | before | after |
|---|---|---|
| `STDDEV::peek` | 26.3–29.8 ns | **19.0–19.2 ns** (−28%) |
| `BBANDS::peek` | 43.1–51.1 ns | **29.8–35.3 ns** (−31%, partial patch) |
| EMA fan-out `update`, k=64…16384 live streams | 2.20–2.80 ns | **1.98–2.25 ns** (−10…−15%) |
| CDLDOJI fan-out `update`, k=4096/16384 | 3.54–4.21 ns | **2.95–3.23 ns** (−15%) |
| `SMA::peek` | 11.1–12.7 ns | unchanged |
| `EMA::peek`, `EMA::update` (single handle) | 2.8 / 3.5 ns | unchanged |
| `EMA_Open` | 7.8 µs | unchanged |

Nothing regressed. Two honest caveats:

- On a single stack-local handle that inlines fully, LLVM already deletes the dead 280-byte copy — hence the flat `EMA::peek` / `EMA::update` rows. The win appears where the copy cannot be elided: the `restore_from` cascade of a composed `peek`, and a fleet of handles in a `Vec`.
- The fan-out win holds at k=64 (20 KiB vs 2 KiB, both L1-resident), so it is not purely cache capacity — a 320-byte stride to reach 40 useful bytes wastes lines and prefetch regardless.

The fan-out shape is the realistic one: one stream per symbol per indicator, updated round-robin per bar.

## Cross-backend

Rust is the only backend paying this. C embeds nothing in `struct TA_SMA_Stream` (settings are file-scope globals); Java's `SMA_Stream` holds `Core core` — a reference; C#'s holds `internal Core core`, and `Core` is a `class`. Only Rust turned the handle's view of `Core` into a 280-byte value copy.

## Proposed change

Emit into the handle exactly the `CandleSetting`s the step body reads, and nothing when it reads none. The required analysis already exists — `candle_settings::detect_candle_settings(steady_stmts)`, already called from `streaming.rs` to classify candle locals — so the emitter change is mostly plumbing:

- `backends/rust_stream.rs`: the struct emit (~792), `restore_from` (~805), the 11 `core: self.clone()` construction sites, the `_step_impl` signature (~1131), and the 2 call sites (~2405, ~2551) which become `Core::<N>_step_impl(<settings>, &mut self.state, …)`.
- `candle_settings.rs::emit_rust_unpacking` needs a receiver path: the batch/`Open` tiers keep `self.candle_settings.<snake>`, the step tier reads its own field or parameter.
- `Core` may become an unused import in the files that end up with no settings.

Cheaper fallback if the per-function narrowing is unwanted: store `CandleSettings` (176) instead of `Core` (280) for the 57, and nothing for the 119. That keeps one uniform shape and still captures the large majority of the win, since the 119 are the ones with the worst ratio.

## Risk

- **No public API change.** Handle fields are private; `Clone`, `Debug`, `Send`, `Sync` are preserved.
- **No semantic change.** Rust's `Core` is immutable (settings come from `CoreBuilder`, producing a new `Core`), so the handle's copy is already a snapshot, and the narrowed set is exactly the read set the generator emits.
- Covered by `regen-check`, `ta_regtest --codegen`, `--xlang-hash` and `cargo test --doc`. The stream `peek`/`update` legs are the ones that matter, so this wants a nightly-class run rather than the PR gate alone.
- `Core`'s own doc comment — "cloning is cheap (it is a small `[i32; N]` array plus two small fields)" — describes 280 bytes and should be corrected either way.

kevinlincg:
Closes #274.

## What

A generated Rust stream handle embedded a whole `Core` by value. It now carries
one `cs_<snake>: CandleSetting` field per setting its own step reads, and no
field at all where it reads none; the step drops its `&self` receiver and takes
those settings as parameters, so `Core` is the namespace it is declared in
rather than state it reads through.

```rust
// before
pub struct CDLDOJI_Stream { core: Core, state: CDLDOJI_StreamState, out: OutRange }
fn CDLDOJI_step_impl(&self, sp: &mut CDLDOJI_StreamState, ...)
    let BodyDoji_factor: f64 = self.candle_settings.body_doji.factor;

// after
pub struct CDLDOJI_Stream { cs_body_doji: CandleSetting, state: CDLDOJI_StreamState, out: OutRange }
fn CDLDOJI_step_impl(sp: &mut CDLDOJI_StreamState, cs_body_doji: &CandleSetting, ...)
    let BodyDoji_factor: f64 = cs_body_doji.factor;
```

The unstable period and the compatibility mode are consumed at `Open`, where
they set the lookback; nothing post-open consults them. `Core` is immutable, so
the handle's copy was already a snapshot — the narrowed set is exactly the read
set the generator emits, which is why this is a size change and not a semantic
one.

## Where the set comes from

`candle_settings::detect_candle_settings` over the function's whole stream
source, not over one tier's steady statements. A superset can never leave a step
naming a field its handle does not have, and the two coincide for every shipped
function: the emitted field counts are **119x0, 15x1, 28x2, 8x3, 5x4, 1x5** —
exactly the read-set census in the issue — because a candlestick's batch
prologue averages exactly the settings its steady state tests.

`emit_rust_unpacking` grew a variant that lets the caller say where each
`CandleSetting` is read from. The batch and `Open` tiers, which do run on a
`Core` receiver, still read `self.candle_settings.<snake>`; only the step tier
moved.

## Size

`size_of`, x86-64, measured before and after on this tree:

| handle | before | after | | handle | before | after |
|---|---:|---:|---|---|---:|---:|
| `EMA_Stream` | 320 | **40** | | `STDDEV_Stream` | 712 | **152** |
| `ATR_Stream` | 320 | **40** | | `HT_TRENDMODE_Stream` | 960 | **680** |
| `RSI_Stream` | 328 | **48** | | `MA_Stream` | 1136 | **576** |
| `MAVP_Stream` | 336 | **56** | | `BBANDS_Stream` | 2168 | **768** |
| `SMA_Stream` | 352 | **72** | | `STOCHRSI_Stream` | 2208 | **808** |
| `CDLDOJI_Stream` | 344 | **80** | | `MACDEXT_Stream` | 3720 | **1760** |

The "before" column reproduces the issue's table exactly. Nesting is where it
compounds: `MA_Sub` is an enum over ten handle variants that each carried their
own `Core`, so a dispatch paid the largest variant's copy whichever MA type was
selected.

## Speed

`scripts/stream_ab.py --base=<upstream dev> --lang=rust --call=peek`, 6
interleaved rounds, `iters=500000`. `peek` is `self.clone()` then `update`, so
it is the call the handle's size lands on directly.

**Read the floor first.** Every one of the 176 handles changed here, so there is
no byte-identical control function inside the A/B. The substitute is the same
tool run with `--base=HEAD` against an unmodified tree — identical code in both
arms, so every row is pure build-layout and measurement noise. Over 176
functions the worst excursion there is **-7.4%**, with only 7 functions past
-5%. Nothing smaller than that is claimed below.

Against that floor, the head of the real run (worst-first, top 20 of 176):

| func | base ns | head ns | chg |
|---|---:|---:|---:|
| `MFI` | 43.18 | 16.53 | **-61.7%** |
| `MAVP` | 704.35 | 270.80 | **-61.6%** |
| `PVO` | 43.96 | 18.63 | **-57.6%** |
| `MACDEXT` | 79.52 | 34.99 | **-56.0%** |
| `PPO` | 45.57 | 20.58 | **-54.8%** |
| `APO` | 44.71 | 21.09 | **-52.8%** |
| `STOCH` | 70.04 | 35.56 | **-49.2%** |
| `STOCHRSI` | 68.66 | 35.64 | **-48.1%** |
| `CDLDOJISTAR` | 32.55 | 17.22 | **-47.1%** |
| `CDLGRAVESTONEDOJI` | 31.22 | 16.61 | **-46.8%** |
| `CDLDOJI` | 28.57 | 15.59 | **-45.4%** |
| `AO` | 31.00 | 18.18 | **-41.4%** |
| `CMF` | 47.54 | 28.57 | **-39.9%** |
| `BBANDS` | 65.11 | 39.49 | **-39.3%** |
| `STDDEV` | 26.90 | 17.43 | **-35.2%** |
| `WMA` | 31.06 | 20.54 | **-33.9%** |
| `STOCHF` | 40.35 | 27.83 | **-31.0%** |
| `WILLR` | 23.66 | 16.50 | **-30.3%** |
| `MIDPRICE` | 20.35 | 14.37 | **-29.4%** |
| `MAX` | 18.41 | 13.72 | **-25.5%** |

Two independent runs of the A/B agree on these rows to within about two points
(`MFI` -61.7 / -59.9, `MAVP` -61.6 / -61.8, `CDLDOJI` -45.4 / -46.2), and the
head-arm absolute ns match what the identical-arms run reads for the same
functions.

`--call=update` was not run: on a single stack-local handle updated in a loop,
LLVM already deleted the dead 280-byte copy before this change — the issue
measures that row flat and this does not move it. The win is where the copy
cannot be elided, which is `peek`'s `restore_from` cascade and a fleet of
handles in a `Vec`; `peek` above is the first of those, and the size table is
the direct evidence for the second.

## Gates

- `a_stream_handle_carries_only_the_settings_its_step_reads` — a one-setting
  step (`CDLDOJI`), a five-setting step (`CDLADVANCEBLOCK`) and a no-setting
  step (`SMA`), pinned against the field set, the step signature and the call
  site. Each row is the control on the others: going back to `core: Core` fails
  all three, widening to "every setting" fails the first two on the count,
  narrowing to nothing fails them on the field. Verified red by mutating
  `handle_candle_settings` to return the empty set.
- `a_stream_step_reads_candle_settings_from_its_parameters` — the step unpacks
  from its parameters and names no `self.candle_settings`; the control in the
  same test is that the batch tier still does.
- `the_transition_tier_is_step_impl_in_every_backend` updated for the Rust row's
  new signature and call site.
- `cargo test` (generator, 382 + 309 + 24), `cargo test --tests -p ta-lib`,
  `cargo test --doc -p ta-lib` (358), `cargo clippy --all-targets` clean on both.
- `ta_regtest --xlang-hash --language=rust`: **PASS**, 176 functions,
  278,274 Rust cases, 0 mismatches, bit-identical against the in-process C
  library — including the tier-agreement leg's `Open` / `OpenAndFill` compares.

## Cross-backend

Rust-only by construction: C embeds nothing in its handle (settings are
file-scope globals), Java holds a `Core` reference, C# a `Core` class reference.
A full `generate` leaves `src/ta_func`, `src/ta_abstract`, the Java, C# and C
outputs, the four servers and both benches byte-identical — the only generated
paths that move are `ta_codegen/output/rust/library/src/ta_func/*.rs`.

`Core`'s own doc comment claimed cloning is cheap because it is "a small
`[i32; N]` array plus two small fields"; it describes 280 bytes, so it is
corrected here as the issue asks.

## Environment note

This tree had no .NET SDK, so `--xlang-hash` ran for Rust (its documented
language today) and the C# server was not built. `--codegen` needs
`bin/ta_ref_serve`, which `scripts/regtest.py` builds from the pinned
`reference-pre-cutover` worktree; that was not available here either, so the
value evidence above is the `--xlang-hash` gate rather than the pre-cutover
oracle. CI covers both.

## Not done here

The issue's "cheaper fallback" (a uniform `CandleSettings` for the 57, nothing
for the 119) is not implemented — the per-function narrowing subsumes it.

---

### DIFF
--- a/ta_codegen/generator/src/backends/rust_stream.rs
+++ b/ta_codegen/generator/src/backends/rust_stream.rs
@@ -29,7 +29,7 @@
 //!   requires the rest to match that length (`Err(BadParam)` otherwise).
 
 use std::cell::Cell;
-use std::collections::{HashMap, HashSet};
+use std::collections::{BTreeSet, HashMap, HashSet};
 use std::fmt::Write;
 
 use crate::helper_registry::HelperRegistry;
@@ -760,6 +760,71 @@ fn scale_by_stride(idx: Expr) -> Expr {
 }
 
 // ---------------------------------------------------------------------------
+// The handle's candlestick settings (issue #274)
+// ---------------------------------------------------------------------------
+//
+// A handle used to embed a whole `Core` by value — 280 bytes, of which a step
+// can reach only the `CandleSetting`s it names: the unstable period and the
+// compatibility mode are consumed at `Open`, where they set the lookback, and
+// nothing post-open consults them. 119 of the 176 generated steps read no
+// setting at all, so they were carrying 280 bytes to read none of them.
+//
+// The handle now carries exactly the settings its own step reads, one
+// `cs_<snake>: CandleSetting` field each, and nothing when it reads none. The
+// step drops its `&self` receiver and takes those settings as parameters, so
+// `Core` is a namespace for it rather than state it reads through.
+//
+// The set is derived from the whole stream source rather than from one tier's
+// steady statements: a superset can never leave a step naming a field the
+// handle does not have, and the two coincide for every shipped candlestick
+// (the batch prologue averages exactly the settings the steady state tests).
+
+/// The candlestick settings this function's handle must carry.
+fn handle_candle_settings(func: &FuncDef) -> BTreeSet<String> {
+    crate::candle_settings::detect_candle_settings(func.stream_source())
+}
+
+/// The handle field, step parameter and unpacking source for one setting:
+/// `"Near"` -> `"cs_near"`.
+fn cs_binding(setting: &str) -> String {
+    format!("cs_{}", crate::candle_settings::pascal_to_snake_case(setting))
+}
+
+/// `, cs_near: &CandleSetting, …` — the step signature's settings parameters.
+fn cs_step_params(settings: &BTreeSet<String>) -> String {
+    let mut s = String::new();
+    for setting in settings {
+        let _ = write!(s, ", {}: &CandleSetting", cs_binding(setting));
+    }
+    s
+}
+
+/// `&self.cs_near, …` — the step call's settings arguments, as a handle reads
+/// them. Distinct field places, so they borrow alongside `&mut self.state`.
+fn cs_step_args(settings: &BTreeSet<String>) -> String {
+    let mut s = String::new();
+    for setting in settings {
+        let _ = write!(s, "&self.{}, ", cs_binding(setting));
+    }
+    s
+}
+
+/// `cs_near: self.candle_settings.near, …` — the handle's settings at every
+/// construction site, where `self` is the opening `Core`.
+fn cs_ctor_fields(func: &FuncDef) -> String {
+    let mut s = String::new();
+    for setting in handle_candle_settings(func) {
+        let _ = write!(
+            s,
+            "{}: self.candle_settings.{}, ",
+            cs_binding(&setting),
+            crate::candle_settings::pascal_to_snake_case(&setting)
+        );
+    }
+    s
+}
+
+// ---------------------------------------------------------------------------
 // Structs
 // ---------------------------------------------------------------------------
 
@@ -779,6 +844,21 @@ fn emit_handle_struct(o: &mut String, func: &FuncDef) {
     let state = state_type_name(func);
     let sn = snake(func);
     let n = func.name.to_uppercase();
+    let settings = handle_candle_settings(func);
+    // Exactly the settings this function's step reads, and nothing where it
+    // reads none (issue #274). Ahead of `state`, so the fields a step touches
+    // every bar sit together at the front of the handle.
+    let mut cs_fields = String::new();
+    let mut cs_restore = String::new();
+    for setting in &settings {
+        let field = cs_binding(setting);
+        let _ = write!(
+            cs_fields,
+            "    /// The `{setting}` setting this stream was opened with.\n\
+             \x20   {field}: CandleSetting,\n"
+        );
+        let _ = writeln!(cs_restore, "        self.{field} = src.{field};");
+    }
     let _ = writeln!(
         o,
         "/// Live {n} stream: one value per closed bar, bit-identical to [`Core::{sn}`]\n\
@@ -789,7 +869,7 @@ fn emit_handle_struct(o: &mut String, func: &FuncDef) {
          #[must_use = \"a stream does nothing unless updated; dropping it closes the stream\"]\n\
          #[derive(Debug, Clone)]\n\
          #[doc(alias = \"TA_{n}_Stream\")]\n\
-         pub struct {handle} {{\n    core: Core,\n    state: {state},\n\
+         pub struct {handle} {{\n{cs_fields}    state: {state},\n\
          \x20   /// The bars this handle has produced a value for — see [`Self::out_range`].\n\
          \x20   out: OutRange,\n}}\n"
     );
@@ -802,7 +882,7 @@ fn emit_handle_struct(o: &mut String, func: &FuncDef) {
          \x20   /// Overwrite from `src`, reusing this handle's buffers instead of\n\
          \x20   /// allocating new ones. See `{state}::restore_from`.\n\
          \x20   pub(crate) fn restore_from(&mut self, src: &Self) {{\n\
-         \x20       self.core.clone_from(&src.core);\n\
+         {cs_restore}\
          \x20       self.state.restore_from(&src.state);\n\
          \x20       self.out = src.out;\n\
          \x20   }}\n}}\n"
@@ -1113,6 +1193,9 @@ fn emit_step(
 fn emit_step_sig(o: &mut String, func: &FuncDef, fallible: bool) {
     let sn = snake(func);
     let state = state_type_name(func);
+    // No `&self`: a step reads nothing from `Core` but the candlestick settings
+    // it names, and those arrive as parameters from the handle (issue #274).
+    let cs_params = cs_step_params(&handle_candle_settings(func));
     let mut params = String::new();
     for bar in streaming::input_array_names(func) {
         let _ = write!(params, ", {bar}: f64");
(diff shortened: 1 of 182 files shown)


mario4tier:
Thanks Kevin — another high-quality contribution. Merged into main.

Issue #278, opened 2026-08-28
Rust/Java/C#: drop cross-language name parity, adopt per-language idiomatic casing (pre-publish)


mario4tier:
Decision from design discussion (2026-08-28): drop cross-language identifier parity for
the generated Rust/Java/C# APIs and adopt each language's own idiomatic naming convention
instead. This only applies to Rust, Java, and C# — **C is untouched**, stays the
canonical/legacy form (20+ years, already shipped, real users, immovable).

Doing this now specifically because none of the three affected backends have shipped yet
(Rust: #179, not yet on crates.io; Java: #172, not yet on Maven Central; C#: no NuGet
release) — all three are documented as not releasing before Q1 2027. Post-publish this
becomes a breaking change with real migration cost; pre-publish it's free. If this slips
past the first publish of any one backend, drop it for that backend and keep the current
C-mirrored names there.

Two rename axes, bundled per identifier:

1. **The generic verb suffix** (`_Open`, `_OpenAndFill`, `_Update`, ...) — mechanical
   recasing to the target language's own method-naming convention. Uncontroversial: it's
   what protobuf/gRPC codegen already does per target language from one canonical
   definition.
2. **The indicator acronym itself** (`SMA`, `RSI`, `MACD`, ...) — until now kept as 
   verbatim uppercase, specifically to mirror C. That exception is dropped: the
   acronym is treated as an ordinary word and re-cased like any other, per each language's
   own official style guide. Documentation/prose is unaffected — `SMA` stays the written
   form everywhere in docs; this is an identifier-only change.

## Rust

Fully idiomatic Rust: `snake_case` for functions/methods with no exceptions (this is what
the `#[allow(non_snake_case)]` on every generated `<NAME>_*` function currently works
around — that suppression goes away), `PascalCase` with the acronym single-capitalized for
types. Matches Rust API Guidelines' "acronyms are one word" rule (`Uuid`, not `UUID`); same
domain precedent in `nalgebra` (`.qr()`, `.lu()`, `.svd()`) and `rustfft` (`Fft`,
`FftPlanner`).

```rust
// before: core.SMA_OpenAndFill(...) -> Result<(SMA_Stream, OutRange), RetCode>
// after:  core.sma_open_and_fill(...) -> Result<(SmaStream, OutRange), RetCode>
```

## Java

Fully idiomatic Java: `camelCase` for methods (Google Java Style Guide — no underscores in
[...]

PR #280, opened 2026-08-29
feat(abstract): a name resolves under an ASCII case fold, in all four backends (#278)

kevinlincg:
> **This replaces the body PR #280 was opened with.** dev landed the fold
> itself in c4cecde3a a few hours before that PR went up, so the change this
> branch carries is no longer the one the old body described.

## What is left after c4cecde3a

c4cecde3a made the by-name lookup case-insensitive in all four backends. The
implementation half of this branch is therefore gone: C, Rust and C# now carry
dev's version verbatim — its `TA_StrCmpNoCase`, its `eq_ignore_ascii_case` —
and this branch has merged dev rather than competing with it.

Two things that commit did not bring are what remains here.

## 1. Java's fold widens, and throws on null

`Functions.byName` folds with `name.toUpperCase(Locale.ROOT)`. That is Unicode
uppercasing, not an ASCII fold, and `Locale.ROOT` does not make it ASCII — it
only removes the *narrowing* Turkish bug while leaving the widening one:

| spelling | `toUpperCase(Locale.ROOT)` | resolves to |
|---|---|---|
| `"sın"` (U+0131 dotless i) | `"SIN"` | **SIN** |
| `"ſma"` (U+017F long s) | `"SMA"` | **SMA** |

So two shipped functions currently answer to names that are not theirs. It is
not a locale-dependent bug — it reproduces under any locale, because
`Locale.ROOT` is the locale.

Separately, `byName(null)` now throws `NullPointerException` where it
previously answered `null` (the map lookup tolerated it; `name.toUpperCase`
does not).

Neither reproduces in C or Rust, whose folds are ASCII by construction. This
PR changes `java_metadata.rs` to emit the hand-written `asciiUpper` this
branch always had, plus the null guard. No other generator change.

## 2. Gates

Each is swept over the whole corpus rather than spot-checked on `"SMA"`. The
names a partial fold gets wrong are the long ones (`CDL3STARSINSOUTH`) and the
ones carrying a digit or an underscore (`CDL2CROWS`, `HT_DCPERIOD`), and no
single name stands in for those. Each probes an all-lower and an alternating
spelling, so every letter position is presented in both cases, and each asserts
the name reported back is still canonical.

| backend | gate | on dev today |
|---|---|---|
| C | `checkFuncHandleFoldsCase` in `test_abstract.c` — 352 spellings | **no gate at all** |
| Java | `byNameFoldsAsciiCase` in `MetadataTest` — part of 1716 checks | none |
| Rust | the corpus sweep is added to dev's own `lookup_is_ascii_case_insensitive`, not duplicated beside it | lower-case sweep + three `"SMA"` spellings |

Each also pins that the fold did not *widen*: U+0130 / U+0131 / U+017F, a
leading or trailing space, `-` for `_`, a name longer than any real one, and
the empty and null cases.

## Controls — run, and watched to fail

```
Java   restore BY_NAME.get(name.toUpperCase(Locale.ROOT)):
       3 of 1716 checks FAIL -- the null answer, U+0131 onto SIN,
       U+017F onto SMA.   <-- this is dev's current code

C      restore strcmp in the generated ta_abstract.c, rebuild:
       ta_regtest fails Abstract interface Tests, error 627

Rust   swap eq_ignore_ascii_case for a to_uppercase() compare:
       the_fold_does_not_widen_what_resolves FAILS and the other 71
       crate tests pass -- INCLUDING dev's own fold test, which is
       what shows the two are not redundant
```

## Verified

- `ta_regtest` — full C reference suite, all groups.
- `ta_regtest --codegen --language=c,rust,java` — 161 functions × 3 languages
  against the frozen `ta_ref_serve` oracle, 0 failures.
- The generator's own suites (382 + 310 + 24 + 4 + …).
- `cargo test --tests -p ta-lib`, `cargo clippy --all-targets`,
  `cargo doc --no-deps` on the crate — all clean.
- All five Java suites on the built jar (`MetadataTest: ALL PASS, 1716 checks`).
- `cargo run -- generate` then `git status` — clean.

## Not verified — please read this before merging

**The C# leg.** There is no .NET SDK on the machine this was prepared on.
Whether `StringComparer.OrdinalIgnoreCase` widens the same way `Locale.ROOT`
does is an **open question, not a claim** — I did not check it. This PR makes
no C# code change and adds no C# gate, precisely because I could not run one.
If the answer is that it does widen, C# needs the same treatment in a
follow-up.

## One thing outside this PR

c4cecde3a also left the generator failing
`cargo clippy --all-targets --manifest-path ta_codegen/generator/Cargo.toml
-- -D warnings` — two `clippy::similar_names` errors, at
`csharp_stream.rs:3430` and `java_stream.rs:3014`, under
`#![deny(clippy::pedantic)]`. `rust-toolchain.toml` pins 1.97.0, so it is not
runner drift: it reproduces on pristine dev. The nightly clippy job is a hard
gate. The two-line rename is on `fix/278-clippy-similar-names` rather than
here, so this PR is not widened by it.


kevinlincg:
This is conflicting, and I would rather say why than force a merge through it.

`c4cecde3a` on `dev` — "drop cross-language streaming-API name parity" — is your own implementation of #278's *first* axis, the one this PR deliberately left alone. It rewrites the same by-name lookup that this PR was making case-insensitive, so the rebase lands 20 conflict hunks across nine files: `ta_abstract.c`, all four metadata emitters, and their generated outputs.

Those are semantic, not mechanical. Resolving them means re-deriving the case fold on top of your new lookup structure, and re-running what backed the original claim — `ta_regtest --codegen=c,rust,java` plus the three controls that were each run to red (C reverted to `strcmp` → 352 spelling failures; Rust with the fold removed → that test fails and the other nine still pass; Java reverted to `BY_NAME.get(name)` → 528 of 1714 checks fail). Hand-merging the hunks and calling it rebased would not carry any of that.

So: leaving the branch as it stands rather than pushing a merge I have not verified. It will be re-derived on top of `c4cecde3a` and force-pushed here, or closed if your commit already covers the fold.

One question that decides which: does `c4cecde3a` change what a "name" is for lookup purposes? If the per-backend identifiers are now the canonical spelling, the ASCII fold may be answering a question that no longer exists in the same form.

Unchanged from the original: the C# leg is unverified here — no .NET SDK on this machine, so `FunctionCatalog.g.cs` and the C# `MetadataTest` were never compiled or run, and the Turkish-I probe was deliberately not written for C# because a probe that cannot run is a guess.

Issue #77, opened 2025-12-11
[CMake] libta-lib.so does not link against `libm` causing missing math symbols at runtime


BwL1289:
When building TA-Lib using the CMake build system, the produced shared library (libta-lib.so) is not linked against libm, resulting in multiple undefined math symbols at runtime and is only linked against libc:

```bash
[./lib/libta-lib.so.0.6.4]:
    linux-vdso.so.1 (0x0000ffff98a32000)
    libc.so.6 => /lib64/libc.so.6 (0x0000ffff986fe000)
    /lib/ld-linux-aarch64.so.1 (0x0000ffff989e5000)
undefined symbol: atan	(./lib/libta-lib.so.0.6.4)
undefined symbol: sin	(./lib/libta-lib.so.0.6.4)
undefined symbol: cos	(./lib/libta-lib.so.0.6.4)
undefined symbol: sqrt	(./lib/libta-lib.so.0.6.4)
undefined symbol: acos	(./lib/libta-lib.so.0.6.4)
undefined symbol: asin	(./lib/libta-lib.so.0.6.4)
undefined symbol: log10	(./lib/libta-lib.so.0.6.4)
undefined symbol: tan	(./lib/libta-lib.so.0.6.4)
undefined symbol: sinh	(./lib/libta-lib.so.0.6.4)
undefined symbol: exp	(./lib/libta-lib.so.0.6.4)
undefined symbol: cosh	(./lib/libta-lib.so.0.6.4)
undefined symbol: log	(./lib/libta-lib.so.0.6.4)
undefined symbol: tanh	(./lib/libta-lib.so.0.6.4)
```

This does not occur when building using the Autotools (configure) build, which correctly links `-lm` build [here](https://github.com/TA-Lib/ta-lib/blob/main/configure.ac#L38-L41).

The CMake build includes checks for the existence of math-related functions [here](https://github.com/TA-Lib/ta-lib/blob/main/CMakeLists.txt#L146-L182) however, it never links libm into the shared library, except in a special-case used only for tests [here](https://github.com/TA-Lib/ta-lib/blob/main/CMakeLists.txt#L543-L544). This linking step never happens for the actual build target (ta_lib or ta-lib), leading to missing symbols in the final shared library.

The fix should be something like:

```cmake
target_link_libraries(ta_lib PRIVATE m)
```

[...]

PR #284, opened 2026-08-29
Support pkg-config in ta-lib for CMake Builds

vikasnkumar:
The `autoconf`build supports `pkg-config` file generation and install, but the CMake build did not. Adding a couple of lines in CMake to support this on UNIX platforms.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -570,6 +570,16 @@ if(UNIX)
 		@ONLY
 	)
 	install(SCRIPT "${CMAKE_BINARY_DIR}/warn-stale-installs.cmake")
+
+    set(prefix ${CMAKE_INSTALL_PREFIX})
+    set(exec_prefix ${CMAKE_INSTALL_PREFIX}/bin)
+    set(libdir ${CMAKE_INSTALL_PREFIX}/lib)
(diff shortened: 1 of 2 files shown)


mario4tier:
Merged, thanks! Followed up with 8cf0839f9 on one detail in `ta-lib.pc.in`.

The added `Libs: ... -lm` wasn't quite right: since #77, `libta-lib.so` already self-declares its libm dependency (`DT_NEEDED libm.so.6`), so a dynamic consumer needs nothing extra — that's what `Libs.private:` is for (only pulled in for static linking via `pkg-config --static`). The actual gap was that `Libs.private: @LIBM@` relies on autoconf's `AC_SUBST([LIBM])`, which CMake's `configure_file` has no equivalent for, so `@LIBM@` was silently substituting to empty on the CMake path.

Follow-up gives CMake its own `LIBM` value (same `if(NOT WIN32)` idiom already used to link the target) and feeds that into `Libs.private` instead, so `Libs:` stays clean and matches autotools' output:

    Libs: -L${libdir} -lta-lib
    Libs.private: -lm

Verified with an actual CMake configure+build that the `.so` still carries `DT_NEEDED libm.so.6` and the generated `.pc` now matches the above.

vikasnkumar:
Thank you !

PR #285, opened 2026-08-29
test(abstract): the fold sweeps' third check asserted nothing, in all three remaining backends (#278 follow-up)

kevinlincg:
Follow-up to `4eb9a6d1d`, which fixed this in Rust and left the note:

> The same comparison is redundant after the identity check in the C, Java and C#
> sweeps too; left alone here, worth a follow-up.

It is redundant in all three, for the same reason — the line above it already
pins the looked-up object to the row being iterated — but the *consequence*
differs per backend, so the three do not get the same edit. Each backend's
reason is recorded at the site rather than in this description.

## C — the sweep gains cover it did not have

Two changes in `test_abstract.c`.

**The anchor.** C's identity check is `handle == canonical`, against a *second
lookup of the row's own name*, not against the row being enumerated. So the
string compare was the only thing tying the sweep to that row — and it does that
only while every name in the table is unique, which is true and which nothing in
this sweep checks. It is now pointer identity against `funcInfo`.
`TA_ForEachFunc` hands out `funcDef->funcInfo` and `TA_GetFuncInfo` answers that
same pointer, so this is the C spelling of what the other three already assert
directly (`lower == f` / `ReferenceEquals(lower, f)` / `Some(f.id)`).

**The new assertion:** the stored name carries no ASCII lower case. Every
spelling the sweep probes is derived from the stored one, and
`TA_GetFuncHandle` folds on both sides, so a lower-cased table entry resolves
under every casing and reads back as itself.

Sabotage, with the control:

| | before this PR | after |
|---|---|---|
| SMA's stored name lower-cased | **all of `ta_regtest` green**, gate prints `352 spellings resolved, all canonical` | `'sma' is not stored in its canonical upper case`, error 627, exit 115 |

The sabotage is a constructor over `TA_INFO_SMA`, so the *library's* table
carries the lower-cased name — not the probe's view of it. Nothing else in
`ta_regtest` reads a stored spelling, which is why the control is green
everywhere, not merely elsewhere.

## Java and C# — the line goes, nothing replaces it

Both already compare the looked-up object to the row being iterated, so the
third line compares one object's name to itself and cannot fail for any
registry. Neither needs the canonical-spelling assertion C now carries, and the
reasons are different:

| | why the fold sweep does not need it | measured, by lower-casing one row |
|---|---|---|
| **Java** | `BY_NAME` is keyed on the stored spelling while `byName` folds *upward*, so a lower-cased row is unreachable by its own name | SMA: four checks fail — `byName(SMA)`, `byName round-trips`, and both lookups in the sweep |
| **C#** | `_byName` IS `OrdinalIgnoreCase`, so the fold checks stay green — but the rest of the suite is keyed on the stored spelling ordinally | TRIX: four `MetadataTest` checks fail (no typed overload matched, compared every function, both XML checks), plus `NoPhantomIoTest` and `StreamApiTest` |

Adding the assertion to these two anyway would have traded one redundant
assertion for another. Recording *why* each backend does or does not need it is
the part that survives the next reader.

## What this costs

One assertion per function disappears in each of Java and C#, so their check
counts drop by exactly 176: **Java 1716 → 1540, C# 3431 → 3255**. That is
visible in the gate output and is the honest price of the change; nothing that
could fail stopped being checked. If you would rather keep a per-function line
there for the diagnosis it gives — `is not stored in its canonical upper case`
names the defect more directly than `lower-case lookup finds it` does — say so
and I will add it back to both, with the redundancy stated rather than implied.

## Verified here

* `ta_regtest` green: `352 spellings resolved to their own row, 176 name(s)
  stored upper case`, `All tests succeeded`.
* Java `MetadataTest`: `ALL PASS (1540 checks)` (compiled with `javac --release
  17` over `src/main` + `src/test`, not through `./mvnw` — this session has no
  Maven cache).
* C#: all seven suites pass on .NET 10.0.11 (`dotnet run -c Release` in
  `csharp/library/test`).
* `cargo run -- generate` leaves the tree clean, confirming all three files are
  hand-written.

**Not run here:** the cross-language `--codegen` legs, which need the language
servers built. No file this touches is compiled into a server.

### DIFF
--- a/src/tools/ta_regtest/test_abstract.c
+++ b/src/tools/ta_regtest/test_abstract.c
@@ -1732,6 +1732,7 @@ static ErrorNumber d2_param_vectors( const char *funcName, const TA_FuncHandle *
 typedef struct
 {
    int nbChecked;
+   int nbCanonical;
    int nbFailed;
 } NameFoldCtx;
 
@@ -1774,6 +1775,24 @@ static void nameFoldCb( const TA_FuncInfo *funcInfo, void *opaqueData )
       return;
    }
 
+   /* "Canonical" has to name something for a fold to fold onto it. Every
+    * spelling probed below is derived from this entry, so a table entry stored
+    * in lower case would fold onto itself just as happily and no probe here
+    * would notice; this is the only line that does. Same assertion the Rust
+    * registry sweep carries (abstract_api.rs, registry_tests).
+    */
+   ctx->nbCanonical++;
+   for( i = 0; i < len; i++ )
+   {
+      if( funcInfo->name[i] >= 'a' && funcInfo->name[i] <= 'z' )
+      {
+         printf( "  ABSTRACT ERROR: '%s' is not stored in its canonical upper case\n",
+                 funcInfo->name );
+         ctx->nbFailed++;
+         break;
+      }
+   }
+
    for( variant = 0; variant < 2; variant++ )
    {
       ctx->nbChecked++;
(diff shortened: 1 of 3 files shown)


mario4tier:
Merged to `dev` as 2c38a968a (rebased onto the current tip, so GitHub did not auto-close this).

Reproduced the C sabotage and its control here, in a throwaway worktree, with a `__attribute__((constructor))` over `TA_INFO_SMA` lower-casing the stored name:

| | result |
|---|---|
| before | `352 spellings resolved, all canonical` — **whole of `ta_regtest` green, exit 0** |
| after | `'sma' is not stored in its canonical upper case`, error 627, exit 115 |

So the sweep really did print `all canonical` about a table that was not, and nothing else in `ta_regtest` noticed. That is the part worth having.

The C identity swap checks out at the source: `TA_ForEachFunc` hands the callback `funcDef->funcInfo` and `TA_GetFuncInfo` returns that same pointer for the same handle, so `back != funcInfo` is the same claim the other three make against the row they iterate — and it no longer leans on name uniqueness. Per-backend reasoning confirmed too: Java's `BY_NAME` is keyed on the stored spelling (`put` → `m.put(f.name(), f)`) while `byName` folds upward, and C#'s `_byName` is `OrdinalIgnoreCase`.

On your offer at the end: yes, please, for both. The two deleted lines were vacuous and deleting them is right, but what replaces them in Java and C# is a failure somewhere else in the suite — for C# that is `NoPhantomIoTest` and `StreamApiTest`, which name nothing about the defect. C and Rust now say it outright; the other two should as well, with the redundancy stated at the site rather than implied. Same shape as the C one, its own counter. Separate PR is fine.

PR #301, opened 2026-08-30
ci(gate): drop #290/#291's PR-gate steps, already covered by the nightly

mario4tier:
Follow-up to #290 and #291. Both landed a step in \`pr-codegen-gate.yml\`'s
\`clippy\` job that runs a command already present in
\`dev-nightly-tests.yml\`'s \`clippy + generator tests\` job, verbatim:

- #290: \`RUSTDOCFLAGS="-D warnings" cargo doc --no-deps -p ta-lib\`
- #291: \`cargo test --tests -p ta-lib\`

Unlike #298's MSRV job (genuinely new coverage, moved to the nightly in
#300), the nightly already ran both of these before #290/#291 landed, so
keeping them on the PR gate too pays extra runner time on every PR for zero
net new coverage. Same "PR gate stays lightweight, nightly is the
only-ever-addition" reasoning, applied to a duplicate rather than a gap.

This removes exactly the two added blocks — the file is byte-identical to
its state before #290/#291 (\`git diff b3041f69b -- .github/workflows/pr-codegen-gate.yml\`
is empty).

### DIFF
--- a/.github/workflows/pr-codegen-gate.yml
+++ b/.github/workflows/pr-codegen-gate.yml
@@ -114,74 +114,10 @@ jobs:
             exit 1
           fi
 
-      # Rustdoc's lints are a set no other step on this gate can reach, and
-      # `rustdoc::broken_intra_doc_links` is the one that matters: only rustdoc
-      # resolves a `[`Core::<F>`]` link, so only rustdoc can report that one
-      # does not resolve. Measured, on a tree where the doc backend emitted 176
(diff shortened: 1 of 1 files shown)


mario4tier:
Reconsidered — closing. #290 and #291 are the same fast-partial-signal-on-PR + nightly-safety-net pattern the clippy job here already uses deliberately (see that job's own comment above), not redundant duplication. Both stay.

Issue #299, opened 2026-08-30
perf(tooling): stream_ab.py has no C# arm, so the .NET streaming tier is unbenchmarked


mario4tier:
`ta_bench_stream` measures the C streaming tier only. `scripts/stream_ab.py` covers Rust and Java. **Nothing measures C#** — the .NET `Update`/`Peek`/`Open` tiers have never had a benchmark in this tree.

That gap is not theoretical. The non-committing peek work needed a before/after for C#, and there was nothing to run it with; the numbers were taken from a throwaway console project built by hand against `TALib.csproj`, with the base arm extracted via `git archive <rev> ta_codegen/output/csharp`. That is not repeatable by anyone else and it is not in CI, so the next change to `csharp_stream.rs` starts from the same place.

### What it should do

A `--lang=csharp` arm matching the two that exist: parse the generated signatures, emit one timed block per streaming function, build the SAME harness against two copies of the library, interleave rounds with alternating arm order, and report marked functions against the untouched ones.

The pieces that differ from Java:

- **Signatures** come from `ta_codegen/output/csharp/library/src/Core_<NAME>.cs`, one file per function — no fragments directory needed, and the filename already names the function.
- The API is `public <Cls>Stream <Name>Open( ... )` on `Core`, the handle is `Core.<Cls>Stream`, and multi-output returns a `readonly record struct` rather than a class. `Peek`/`Update` are PascalCase.
- **Build**: a generated `.csproj` with a `ProjectReference` to the arm's `TALib.csproj`, `dotnet run -c Release`. Note `TreatWarningsAsErrors` is on in the library csproj, so a base arm from an older revision must still compile clean under the current SDK.
- **Warm-up**: the JIT needs the same discard-the-first-passes treatment the Java arm already does, and tiered compilation makes the FIRST timed block systematically slower than later ones — the hand bench showed p=30 reading high in both arms purely from block order. Either pin `DOTNET_TieredPMStubs`/`TieredCompilation` or run each function in its own process.

### Guardrail this arm should carry from day one

Both existing arms have been silently broken by an API rename. #278 recased Rust and Java; each arm's signature regex still spelled the old names, so it parsed zero functions. Rust's exited 0 and was fixed in `c308e789b`; **Java's stayed broken until 2026-08-30** and reported only "requested but not benchable".

[...]

PR #302, opened 2026-08-30
perf(tooling): a C# arm for stream_ab.py — and the Java arm has been measuring nothing since #278 (#299)

kevinlincg:
#299 asks for a `--lang=csharp` arm on `scripts/stream_ab.py` matching the two that exist, and a floor assertion so a benchmark that measures nothing cannot exit 0. Building the first one turned up the second's motivating case still live.

## The Java arm parses 0 of 176 functions on `dev` today

#299 records that #278's recasing broke both existing arms, that Rust's was fixed in `c308e789`, and that Java's "stayed broken until 2026-08-30". On `dev` at `12987063` it is still broken. `parse_java` looks for

    public (\w+)_Stream (\w+)_Open\( (.*?) \)

and the generator has emitted `public SmaStream smaOpen( double inReal[], ... )` since #280 landed. Measured on this tree before any change here:

    rust parsed 176 declined 0
    java parsed 0   declined 0

The consequence is not silent — `--lang=java` dies at the existing "parsed 0" guard, or reports every `--mark`ed function as "requested but not benchable" — but the Java streaming tier has had no A/B tool at all since 28 Aug. This PR fixes it: the Open regex, the handle-class regex, and the emitted `Core.SMA_Stream st = core.SMA_Open(...)` call, which had the same stale spelling baked into `JAVA_BLOCK`/`JAVA_OPEN_BLOCK`.

## The floor gate

`FUNC_FLOOR = 170` (against 176 today, in each of the three languages), asserted on what the parser *found*, not on what `--funcs` selected — it is a claim about the parser, so it must fire whether or not the invocation narrowed the set.

Each arm also now takes the function NAME off the filename rather than off the method, and decides indicator-ness on something outside the stream API (Rust: the batch `XXX_Lookback`; Java/C#: the `Core_XXX.` filename). A stream-API recasing therefore turns a function into a reported *decline* instead of a silent disappearance. The decline list is capped at 8 entries plus a count — when a rename takes out the whole API, 176 identical reasons bury the one line that says what to do.

**Control, and it goes red.** I edited each arm's Open regex back to its pre-#278 spelling, one arm at a time, and ran it:

    [rust]   parsed 0 streaming functions, below the floor of 170 - ... Fix parse_rust() ...     exit=1
    [java]   parsed 0 streaming functions, below the floor of 170 - ... Fix parse_java() ...     exit=1
    [csharp] parsed 0 streaming functions, below the floor of 170 - ... Fix parse_csharp() ...   exit=1

All three fire before any build work. Reverted; all three then parse 176.

## The C# arm

Signatures come from `ta_codegen/output/csharp/library/src/Core_<NAME>.cs`, one file per function. All 176 parse with no declines. The arm builds a generated `stream_ab.csproj` with a `ProjectReference` to that arm's `TALib.csproj`, taking the TFM off the arm's own csproj rather than pinning one here, since the two arms are different revisions and the library has retargeted before.

Two decisions worth stating, both costs rather than wins:

**`TieredCompilation` is pinned off** in the harness project. #299 reports that tiering makes the first timed block in the process read systematically slower than later ones purely from position, which discarding warm-up passes cannot fix because the effect is per-process and per-position. Off means every method is fully optimised on its first call. **The cost: dynamic PGO goes off with it**, so the C# ns columns are static-opt numbers, not what a long-lived process settles at. Both arms get identical treatment so the change column — the output — is unaffected, and the ns columns were never comparable across invocations anyway. I did *not* independently reproduce the block-order effect; I pinned the setting the issue names, which is deterministic and arm-symmetric either way.

**`--lang` now defaults to `rust,java,csharp`**, so a bare `scripts/stream_ab.py --base=X` now needs `dotnet` as well as `cargo` and `javac`. That is a real new prerequisite for the default invocation. There is a preflight check for every requested arm's toolchain that names the missing tool and says to drop it from `--lang`, run before any build so the rust leg's minutes are not spent first. If you would rather the default stay `rust,java`, that is a one-line change — say so and I will make it.

Smaller notes: the arm runs the built dll rather than `dotnet run` (as #299 suggested), because `measure()` invokes the command 8 times per round and `dotnet run` re-evaluates the project each time; building once keeps the arm's binary fixed for the whole run, as the other two arms do. And `stage_tree` now takes an `ignore` list, used for `bin`/`obj`: `git archive` never carries build output but the head arm's `copytree` does, and a stale `obj/project.assets.json` pins absolute paths from the tree it was restored in.

## Proving the arm can see something

Per the `ta-bench` skill's standing requirement for a C# lane — reproduce a *known* effect before trusting the tool on an unknown one.

**Null control.** Both arms byte-identical (`--base=HEAD`), full 176-function sweep, `--call=peek --rounds=4 --iters=100000 --passes=3`:

    marked   n=110  median -0.5%  (-0.45 ns)
    control  n=66   median -0.3%  (-0.07 ns)  past 10%: 2  past 25%: 0
    median control spread 10.7%

**Known effect.** 81 of the 176 generated C# `Peek`s reuse a cached `peekScratch` handle; the other 95 allocate a fresh copy. In a throwaway working-tree edit (never committed) I rewrote those 81 to allocate too, marked exactly them, and left the other 95 as byte-identical code. Same settings:

    marked   n=81   median +143.0%  (+101.82 ns)
    control  n=95   median -0.2%   (-0.09 ns)  past 10%: 18  past 25%: 1
    median control spread 22.5%

The effect group moves +143% and the untouched group sits at -0.2%. The arm sees a real change and does not manufacture one.

Both runs on .NET SDK 10.0.400 / `net10.0`, on a shared box — hence the 10–22% median per-row spread, which is why the group medians and not any single row are the readable part.

## What I did not check

- Only Linux. Nothing here was run on Windows or macOS.
- The full-sweep evidence above is `--call=peek`. `--call=update` and `--call=open` were smoke-run on small function sets for all three arms, not swept.
- The `TreatWarningsAsErrors` hazard #299 names — a base arm from an older revision must compile *warning-clean* under today's SDK, not merely compile — is documented in the docstring and the skill, and I saw it fire once (an unused-field CS0169 during the sabotage run above). A real older base does work: `--base=edb9c6c6 --lang=csharp` (29 Aug, a different C# library than HEAD) builds both arms clean under SDK 10.0.400. I did not survey how far back a `--base` can reach before it stops building, and a pre-#278 base will not work at all, since the emitted harness is written against today's signatures.

## Files

- `scripts/stream_ab.py` — the C# arm, the Java fix, the floor, the toolchain preflight, filename-derived names.
- `.claude/skills/ta-bench/SKILL.md` — the C# lane is no longer a gap; the floor and the tiering cost are written down where someone defending a number will read them.

### DIFF
--- a/.claude/skills/ta-bench/SKILL.md
+++ b/.claude/skills/ta-bench/SKILL.md
@@ -112,27 +112,40 @@ cd bin && ./ta_bench_stream --points=20000 --iters=50
 ./ta_bench_stream --points=20000 --iters=50 --min-ratio=0.35   # exits 1 if any func is below
 ```
 
-`ta_bench_stream` is **C only**. For the Rust and Java streaming tiers,
+`ta_bench_stream` is **C only**. For the Rust, Java and C# streaming tiers,
 `scripts/stream_ab.py` A/Bs `update` (or `peek`) per bar — or `open`, which times
 the whole warm-up instead of one bar — between the working tree
 and a git revision — same generated harness compiled against two copies of the
 library, interleaved rounds with alternating arm order, every streaming function
-so the untouched ones are the control. It reads only the generated Rust crate and
-Java fragments (no ta_abstract, no servers, no C build) and derives every call
-from the emitted signatures, so adding an indicator needs no edit there.
+so the untouched ones are the control. It reads only the generated Rust crate,
+Java fragments and C# library (no ta_abstract, no servers, no C build) and
+derives every call from the emitted signatures, so adding an indicator needs no
+edit there.
 
-**C# streams too, but has no lane in either tool** — there is no C# row in any
-benchmark table, and `ta_bench --mode=open` answers `unsupported_mode` for it.
-So the C# peek-scratch election is Java's predicate shipped unchanged, and the
-emitted docs deliberately state what `Peek` allocates rather than claiming it is
-cheaper, which is a comparison nothing here has measured. Adding a C# lane needs
-a control arm first: reproduce a *known* effect (switch the copy constructor to
-`Clone()` and require ~2x on array-owning functions while the array-less ones
-stay flat) before trusting the tool to settle an unknown one.
+Every arm asserts a **floor** (`FUNC_FLOOR`, 170 against 176 today) on how many
+functions it parsed. That is not a style check: #278 recased the Rust and Java
+stream APIs and both arms' regexes kept the old spelling, so each parsed zero —
+Rust's until `c308e789`, Java's for a further day. Nothing but a person trying
+to use the tool noticed either. If an arm dies saying it is under the floor, the
+generated API moved: fix the parser, do not lower the floor.
+
+The **C# arm** pins `TieredCompilation` off in the harness project, so every
+method is fully optimised on its first call. The cost, stated rather than
+hidden: dynamic PGO is off with it, so the C# ns columns are static-opt numbers
+rather than what a long-lived process settles at. Both arms get the same
+treatment, so the change column — the output — is unaffected, and the ns columns
+were never comparable across invocations anyway. Note also that `TALib.csproj`
+sets `TreatWarningsAsErrors`, so a `--base` from an older revision has to compile
+*warning-clean* under today's SDK, a higher bar than the other two arms impose.
+
+There is still **no C# row in `ta_bench`** — `ta_bench --mode=open` answers
+`unsupported_mode` for it — so batch-vs-stream ratios remain a three-language
+table.
 
 ```bash
-scripts/stream_ab.py --base=origin/dev                                  # both languages
+scripts/stream_ab.py --base=origin/dev                                  # all three
 scripts/stream_ab.py --base=HEAD~1 --lang=rust --call=peek --mark=MIN,MAX
+scripts/stream_ab.py --base=origin/dev --lang=csharp --call=peek
 scripts/stream_ab.py --base=origin/dev --call=open --mark=BBANDS,STDDEV   # the Open tier
 ```
 
--- a/scripts/stream_ab.py
+++ b/scripts/stream_ab.py
@@ -1,14 +1,15 @@
 #!/usr/bin/env python3
-"""A/B the Rust and Java streaming tier between two revisions of this tree.
+"""A/B the Rust, Java and C# streaming tier between two revisions of this tree.
 
 `ta_bench_stream` measures the C streaming tier only. This measures the other
-two: it times `update` (or `peek`) per bar for every streaming function -- or
+three: it times `update` (or `peek`) per bar for every streaming function -- or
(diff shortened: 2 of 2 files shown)


kevinlincg:
Rebased onto `da303aab`. Two things in the original description no longer describe this tree, and I would rather say so than quietly re-word the PR.

**The Java arm is no longer mine to fix.** `87b625e31` landed the same repair independently — filename-derived name, class and open spellings carried in the func record — an hour or so after this went up. I took `dev`'s version wholesale in the rebase, including its `cls`/`blk` naming, and dropped my `handle` key from the C# record since nothing read it. What survives from my Java hunk is only the `declined[name] = "no Open signature matched"` line, which merged cleanly: a rename now shows up as a named decline instead of a silent disappearance.

**The known-effect evidence is stale.** It was measured by making 81 functions that reused the `[ThreadStatic] peekScratch` handle allocate instead, and `0546c0fa5` has since deleted that tier outright. The `marked +143.0% / control -0.2%` split was real when it was run, but it was run against code that is not in the tree any more, so it is no longer a reproducible demonstration that this arm can see an effect. I have not re-run it: the local box here is on SDK 7.0.200 and that sweep was SDK 10.0.400 / `net10.0`. Treat the C# arm as unproven on discrimination until someone re-runs a known effect against the post-`0546c0fa5` sources.

What I did re-verify on the rebased tree, since it is cheap and does not need .NET:

    rust    parsed 176  declined 0
    java    parsed 176  declined 0
    csharp  parsed 176  declined 0   (FUNC_FLOOR = 170)

and the floor still discriminates — reverting each arm's Open regex to its pre-#278 spelling, one at a time, drops that arm to `parsed 0` and trips the floor:

    [rust]   parsed 0 -> below floor
    [java]   parsed 0 -> below floor
    [csharp] parsed 0 -> below floor

So the parts I would still defend are the C# arm itself, the floor, the toolchain preflight and the `bin`/`obj` ignore in `stage_tree`. The `TieredCompilation`-off cost and the new `dotnet` prerequisite for the default `--lang` are unchanged from the description above, and the offer there stands: if you would rather the default stay `rust,java`, say so and it is a one-line change.

mario4tier:
Landed on `dev` as c88254457 — rebased, so this shows closed rather than merged; git authorship is preserved.

Verified before the push: all three arms parse 176/176 (the Java one was 0 on `dev`), C# runs end-to-end against a two-arm build, and `FUNC_FLOOR` was sabotage-proven — recasing the C# open regex back to the pre-#278 spelling exits 1 with the diagnostic, so a run that measures nothing can no longer report a clean sweep.

The red clippy check was not from this PR. It came from generator lints on `dev` that I introduced and fixed in f360b8b49; re-running the job reused the original merge ref, so it kept reporting the old failure. On the rebased result the generator and crate clippy, regen-check and the 18 generator suites are all green.

PR #308, opened 2026-08-30
test(stream): the managed peek sweeps read `=` and no other store operator

kevinlincg:
`no_java_peek_copies_the_handle` and `no_csharp_peek_copies_the_handle` state the
rule the peek frame lives by — a peek may assign only to a name the frame
DECLARED, and a bare write to a handle field is an offender. They were written
two commits ago (`c7b17bc9b`) to assert *the class* rather than the instance.
They assert it for one operator.

`assign_target` bails on any line whose left side ends in `+ - * /`:

```rust
if lhs.ends_with(['=', '!', '<', '>', '+', '-', '*', '/']) {
    return None; // ==, !=, <=, +=, ...
}
```

`<=` and `+=` are in the same comment and only one of them is a comparison. So
every compound assignment is classified as writing nothing, and `++`/`--` are
never looked at at all.

## Why that is the shape the corpus is made of

These are not operators the frames might one day use. They are what the frames
are written in:

| tier | the line | operator |
|---|---|---|
| loop (SMA) | `periodTotal += (double)inReal;` | `+=` |
| extrema (BETA) | `pkIdx0 = i++ & sp.xMask;` | `=` **and** `++` |
| Hilbert / extrema (15 handles) | `while( ++i <= today ) {` | `++` |

Each of those names is a handle field that the frame has to localize first
(`double periodTotal = sp.periodTotal;`, `int i = sp.i;`). Drop the
declaration and the bare name resolves to the field and commits it — which is
exactly the defect `c7b17bc9b` fixed, stated there as "the frame declared no
local for it". For two of the three operators above, neither sweep could have
said so.

## The change

`write_targets` replaces `assign_target` in both sweeps. It reads every
operator that stores, and returns *each* write on the line rather than the
first:

- the assignment, compound or not — and a compound one never declares, which is
  the whole point (`sp.x += e` cannot introduce `sp.x`);
- each `++`/`--` wherever it sits, because an increment inside a `while` or
  `if` header commits exactly as one on its own line does;
- the assignment first, so a `for( int i = 0; ...; i++ )` header still declares
  `i` before its own increment is read.

`==`/`!=`/`<=`/`>=` stay comparisons; `<<=`/`>>=` are stores, which is why the
`<`/`>` case looks one character further back. The Java sweep now skips comment
lines in the write scan, as the C# one already did.

## Controls

Each sabotage was run against dev's sweep and against this one, in both
backends:

| sabotage | old sweep | new sweep |
|---|---|---|
| drop SMA's `periodTotal` local, so the frame's own `+=` resolves to the field | **green** | red: `sma: writes the handle: periodTotal += (double)inReal;` (and the `-=` line) |
| a `sp.i++;` in MIN's frame | **green** | red: `min: writes the handle: sp.i++;` |
| clean tree | green | green |

The first is a real emitter defect rather than an injected line — one `continue`
in `peek_frame_arm_named`'s local-declaration loop, which is the `c7b17bc9b`
class one operator over. Both go red in Java and in C#.

`cargo test --no-fail-fast` is green (386 + 312 + every suite), `cargo clippy
--all-targets` is clean, and `generate` leaves the tree byte-unchanged: this
touches no emitter.

## What it does not cover

- **C is deliberately unchanged.** Its frame runs on a stack copy of the
  struct, so only the shared pointer buffers can be committed, and
  `peek_suite`'s `buffer_stores` already reads `=`, `+=` and `-=` on them —
  a superset of what the emitter produces. I checked: no `*=`, `/=`, `++` or
  `--` reaches a `sp-><buf>[...]` slot anywhere in `src/ta_func/`. Adding those
  operators there would be a check with no shape to constrain.
- **Rust needs none of it.** Its frame binds `&self.state`, so a commit does not
  compile.
- **An increment as the SOLE write to a name is not proven.** Every incremented
  name in today's corpus is also plainly assigned somewhere in the same frame,
  so dropping its declaration turns the old sweep red anyway (via the plain
  assignment) and discriminates nothing. That arm's control is the injected
  `sp.i++;` above — it proves the detector reads the operator, not that a
  realistic emitter defect would reach it alone today.
- **A store through an array alias is still invisible**, in both managed
  sweeps: `double[] r = sp.ring; r[k] = v;` declares `r` and the store counts as
  local. No such bind is emitted — every array a frame localizes is `.clone()`d
  (Java) or `Array.Copy`d (C#), and the read-only buffers get no local at all —
  so I left it, rather than adding a rule with nothing to hold.
- **I did not run the four-language `stream_verify` or `--xlang-hash`** for this
  change. Neither can see the property (a peek that commits what the following
  update commits is bit-identical on every value leg, which is why the sweep is
  structural), and the diff reaches no emitted code.


kevinlincg:
Sequencing note, since this lands in files #307 is also editing: this branch is cut from `dev` (`f5909c8b`) and touches only `backend_suite.rs` and `java_stream_suite.rs`, both of which #307 also changes. It is not a competing change — #307 adds a fixed-size-accumulator gate, this widens the operator set the two existing `no_*_peek_copies_the_handle` sweeps read — but whichever lands second will want a rebase, and #307 should be the one to go first since it is the larger change and mine is the one that is cheap to redo.

Worth flagging in the other direction too: if #307's new sweeps classify writes with `assign_target`, they inherit the same one-operator blind spot this fixes, and `write_targets` would drop straight in.

Issue #310, opened 2026-08-30
Java streaming: multi-output update() allocates a Value record every bar because it is cached for value()


mario4tier:
Java's multi-output stream `update()` allocates a `Value` record on **every bar** — 40 B and roughly 26 ns of a ~30 ns call, on the 15 multi-output streamable functions. C#'s identical generated function measures 5.78 ns and 0 B. The outputs are already fields, so the allocation is not the numerics; it is the return record, and specifically the fact that it is *stored* before being returned.

`ta_codegen/output/java/fragments/Core_MACD.java:683`:

```java
public Value update( double inReal ) {
   if( !Double.isFinite(inReal) )
      throw new TaLibArgumentException("MACD update: BadParam", RetCode.BadParam);
   core.macdStepImpl(this, inReal);
   if( this.outRangeCount < MAX_INDEX ) this.outRangeCount++;
   this.cachedValue = new Value(this.cur_outMACD, this.cur_outMACDSignal, this.cur_outMACDHist);
   return this.cachedValue;
}
```

The store to the `cachedValue` field (`:615`) is what puts the record on the heap: it escapes, so escape analysis cannot scalar-replace it. `peek` runs the same arithmetic and returns the same record **unstored** (`:757`) — and measures **0 B / 2.74 ns**. That is the whole delta.

The field exists only so `value()` (`:765`) is a field read. `docs/streaming-api-design.md` states the intent: *"multi-output `update` caches the immutable `Value` it returns, so `value()` is allocation-free."* The measurement says the trade goes the wrong way — every bar of the hot path pays a guaranteed heap allocation to make an accessor free that most callers never call, and callers who do call it call it at most once per bar.

C# does not have the field at all and computes `Value` on demand (`Core_MACD.cs`, `readonly record struct`). So the two managed backends already disagree on this, and the one without the field is the faster one on both paths.

**Measured** (`ThreadMXBean.getThreadAllocatedBytes`, min-of-N, contended box, noise floor ±7% on the min — the deltas here are far outside it):

| | Java `update` | Java `peek` | C# `Update` |
|---|---:|---:|---:|
| MACD (3 outputs) | ~30 ns, 40 B | 2.74 ns, 0 B | 5.78 ns, 0 B |
| BBANDS | ≥15.8 ns (typ 19–45), 40 B | — | 24.67 ns, 0 B |

Single-output functions are unaffected (0 B in all four backends).

[...]

Issue #287, opened 2026-08-29
feat(streaming): C/Rust lack value()/copy() parity with Java/C#, and copy()'s range fields are untested


mario4tier:
Filed from a website-docs review: writing the new **Utility Calls** section for the streaming API docs (`website/src/api/{,rust/,java/,csharp/}stream/README.md`) surfaced an asymmetry across the four backends' utility accessors, worth deciding before the streaming tier ships (all four backends are still "Not yet released").

## Current surface, verified against the generator

| accessor | C | Rust | Java | C# |
|---|---|---|---|---|
| range (`OutRange`) | `TA_StreamOutRange` (generic, one fn for all streams) | `stream.out_range()` | `stream.outRange()` | `handle.OutRange` |
| re-read last value w/o recompute | — | — | `stream.value()` | `handle.Value` |
| fork an independent copy | — | `Clone` (derived; free, not hand-emitted) | `stream.copy()` | `handle.Clone()` |

`c_stream.rs` / `rust_stream.rs` emit exactly `Open[AndFill]` / `Update[AndFill]` / `Peek` / `Close` per function, plus C's separate generic `TA_StreamOutRange` in `ta_utility.c` — no `value`/`copy` emission in either. Rust's `Clone` comes from `#[derive(Clone)]` on the handle struct, not a generated method, and is never exercised by name in the docs or (see below) the test harness.

So: **C is missing both `value()` and `copy()`. Rust is missing `value()` only** — it already has `Clone` structurally, just not as a documented or tested utility call.

## Whether these are worth adding — for this issue to decide

- `value()`: in C/Rust, every call that can produce a value (`Open`/`Update`/`Peek`) already hands it back through a return/out-param, so a caller who threads the value alongside the handle never needs it. It only matters once a handle crosses a boundary that doesn't carry the value with it (stored in a collection, returned from a function, etc.) — decide whether that's common enough to justify the accessor in these two backends.
- `copy()`/duplicate in C: no exposed way to fork a live `TA_<N>_Stream *` today; the only way to get a second independent trajectory is to replay history into a fresh `Open`. A generic `TA_StreamCopy`, mirroring `TA_StreamOutRange`'s "one accessor, not one per function" shape, would need a way to know each concrete struct's size — which `TA_StreamOutRange` doesn't, since it only reads the shared head.

## A confirmed test gap found along the way (measured 2026-08-29, not hypothetical)

[...]

PR #311, opened 2026-08-31
perf(java): a multi-output stream `update` returns its `Value` unstored (#310)

kevinlincg:
Closes the investigation #310 asked for, and implements what it proposed. The
allocation claim replicates exactly. The **cost** claim does not go the way the
issue expected, and that is the part worth reading before merging.

## The change

One generator file. `backends/java_stream.rs` no longer emits the
`cachedValue` field on a multi-output stream handle:

- `update` returns `new Value(...)` **without storing it**. The store was what
  made the record escape, and an escaping record is a heap allocation the JIT
  cannot remove.
- `value()` builds its `Value` from the same `cur_<out>` fields when it is
  called — which is exactly what the C# tier already does.
- `updateAndFill` loses the `int done` / `try` / `finally` that existed only to
  keep the cache fresh on a partial fill. `value()` now names the last committed
  bar by construction, for the same reason C# always did.
- `copy()` copies one fewer field; the three open-capture sites seed one fewer.

Regenerating touches the 15 multi-output streamable functions and nothing else:
`ACCBANDS AROON BBANDS HT_PHASOR HT_SINE MACD MACDEXT MACDFIX MAMA MINMAX
MINMAXINDEX SMI STOCH STOCHF STOCHRSI`. **C, Rust and C# output is byte-identical**
(`generate` then `git status`: only `output/java/` moved). Net −135 lines of
generated Java.

## Allocation: the mechanism claim, confirmed

Throwaway probe, `ThreadMXBean.getThreadAllocatedBytes`, two arms compiled from
the same probe source against two copies of the generated library (this branch
vs `dev` at `f5909c8`), min-of-N, 3 interleaved rounds with alternating arm
order, 1M updates per timed block.

| | base B/update | this B/update | base ns | this ns | chg |
|---|---:|---:|---:|---:|---:|
| MACD (3 out) | 40.00 | **0.00** | 7.62 | 4.19 | −45.0% |
| BBANDS (3 out) | 40.00 | **0.00** | 16.95 | 13.18 | −22.2% |
| STOCH (2 out) | 32.00 | **0.00** | 25.25 | 21.42 | −15.2% |
| AROON (2 out) | 32.00 | **0.00** | 16.59 | 13.75 | −17.1% |
| MINMAX (2 out) | 32.00 | **0.00** | 15.27 | 12.31 | −19.4% |
| MAMA (2 out) | 32.00 | **0.00** | 59.25 | 57.37 | −3.2% |
| *SMA (control)* | *0.00* | *0.00* | *4.21* | *4.19* | *−0.4%* |
| *RSI (control)* | *0.00* | *0.00* | *8.24* | *8.22* | *−0.2%* |

The two single-output controls read 0 B in **both** arms and move −0.4% / −0.2%,
which is what says the probe itself allocates nothing and the box was quiet.

## Timing across the whole tier, with 161 functions as the control

`scripts/stream_ab.py --base=f5909c8 --lang=java --call=update --rounds=7`
(500k iters/pass, interleaved, alternating arm order):

| group | n | median | p05 | p95 |
|---|---:|---:|---:|---:|
| multi-output | 15 | **−1.90%** | −3.84% | −0.13% |
| everything else (control) | 161 | −0.13% | −1.17% | +0.73% |

14 of the 15 sit below the control's p05; the 15th (HT_SINE, +0.4%) costs ~1000
ns/bar in this harness and would not notice 40 B either way.

Two caveats on the magnitudes, stated rather than buried:

- The **absolute** ns here are not the issue's. #310 measured MACD `update` at
  ~30 ns; this box reads 51.8 ns through `stream_ab` and 7.6 ns through the
  isolated probe. `stream_ab` compiles all 176 functions into one class, and the
  per-block inflation is what the gap is. Both arms get it identically, so the
  change column is unaffected — but do not read a ns column here against #310's.
- I could not reproduce the issue's ~26 ns saving on MACD. On this hardware the
  saving is 3.4 ns isolated and ~1 ns in the crowded harness. The direction and
  the control separation replicate; the size does not.

## The cost, which is larger than the issue expected

The issue's reasoning was that `value()` is *"an accessor that most callers never
call, and callers who do call it call it at most once per bar."* The first half
holds. The second half does not survive measurement.

| caller shape, MACD | base B/bar | this B/bar | base ns | this ns | chg |
|---|---:|---:|---:|---:|---:|
| `update` only | 40 | **0** | 7.62 | 4.19 | −45.0% |
| `update` + `value()` **1×/bar** | 40 | **0 or 80** | see below | | |
| `update` + `value()` **4×/bar** | 40 | **200** | 9.13 | 33.01 | **+261.6%** |
| `update` + `value()` **1 bar in 8** | 40 | 45 | 8.57 | 8.03 | −6.3% |
| `value()` polled, no update | 0 | **40** | 1.20 | 5.59 | **+366.4%** |
| *`SMA` + `value()` (control)* | *0* | *0* | *4.60* | *4.54* | *−1.4%* |

Two things there need naming:

- **`value()` is no longer free.** Polled on its own it goes 0 B → 40 B and
  1.20 ns → 5.59 ns. That is the trade, not a surprise — but it is a 4.7x on
  that accessor, not a rounding error.
- **At exactly one `value()` per bar the result is not stable.** I measured that
  row twice, from runs differing only in what *else* was compiled into the same
  probe class: once at **0 B / −61.6%**, once at **80 B / +50.8%**. Escape
  analysis fired in one and not the other. The issue already flagged that EA is
  a JIT decision; this is what that looks like in practice, and it means the
  1×/bar caller can land either side of today's behaviour.

So the honest summary of the trade: strictly better for a caller that reads
`update`'s return (the shape the API is built around), strictly worse for a
caller that polls `value()`, and JIT-dependent in between. Whether that is the
right trade is a maintainer's call, not mine.

**A third option I did not implement**, in case it is preferred: keep the field
but populate it lazily — `update` writes `cachedValue = null` (a null store does
not make the returned record escape) and `value()` fills it on first read per
bar. That would be 0 B/update *and* at most 40 B/bar for any number of `value()`
calls, i.e. better than both this branch and `dev` on every row above. It costs
the field back on `copy()`, a store per bar, and a benign race on `value()`
(safe under JLS 17.5 final-field publication, which the design doc already
leans on). Say the word and I will build and measure it.

## One observable API change

`value()` no longer returns the *same instance* `update` returned. It returns an
**equal** one — `Value` is a record, so `equals`/`hashCode`/`toString` and every
component read are unchanged. Only reference identity moved, and it moved to
match C#, where a record struct never had identity to share.

`StreamSmokeTest` asserted the old identity (`m.value() == v1`) and now asserts
both halves of the new contract: equal to what `update` returned, and a distinct
instance. Its `HashSet` check becomes a real test of the record's
`equals`/`hashCode` — before, identity satisfied it.

## Gates, and which of them I watched go red

- `tests/java_stream_suite.rs::test_java_mama_value_class_protocol` — asserts no
  `cachedValue` and the exact unstored return. **Verified red** against the
  unfixed generator (`assertion failed: !s.contains("cachedValue")`).
- `tests/update_and_fill_suite.rs::no_managed_update_and_fill_caches_the_multi_output_value`
  — replaces `the_java_value_cache_is_refreshed_on_every_exit`; sweeps Java and
  C#, `updateAndFill` and the whole handle, for any cache or `finally`.
  **Verified red** against the unfixed generator.
- `StreamSmokeTest`'s partial-fill check (`value()` must name the last committed
  bar after `updateAndFill` throws) is what says the `finally` was removable
  rather than load-bearing. It passes here unchanged.
- `StreamSmokeTest`'s new identity assertions — **verified red** by compiling
  this branch's test against the `dev` library: `FAIL: multi-output value() is
  not a cached instance`.

Green here:

- `cargo test` in `ta_codegen/generator` — every suite, 0 failed.
- `cargo clippy --all-targets -- -D warnings` — clean.
- The shipped Java suites, on the jar: `StreamSmokeTest` 3860 checks,
  `MetadataTest` 1722, `BatchApiTest` 181, `CoreApiTest` 66, `DivZeroTest` 91,
  `SMathOverflowTest` 4 — all ALL PASS.
- `ta_regtest --xlang-hash --language=java` — **PASS**, 176 functions,
  276,983 Java cases, 0 mismatches, bit-identical against the in-process C
  library (batch, `Open` and `OpenAndFill` tiers, plus the L2/L3 tier-agreement
  and golden legs).
- `ta_regtest`'s own C reference suite — all groups pass, including the
  streaming finite-input gate and the `UpdateAndFill` partial-commit gate.

Two things I did **not** check, named rather than implied:

- **The `--codegen` leg.** It needs `bin/ta_ref_serve`, built from the pinned
  `reference-pre-cutover` worktree, which this box does not have; the run stops
  at "cannot start ta_ref_serve". So the `Stream verify: N legs bit-exact vs
  batch` line is not in evidence here. `--xlang-hash --language=java` and
  `StreamSmokeTest`'s own stream-vs-batch sweeps are what stand in its place.
- **Anything .NET.** No `dotnet` SDK on this box, so the C# half of the new
  `updateAndFill` sweep was exercised only through the generator's in-process
  rendering, and no C# was compiled. The generated C# is byte-identical to
  `dev`, so there is nothing new for it to break — but I did not compile it.

### DIFF
--- a/docs/streaming-api-design.md
+++ b/docs/streaming-api-design.md
@@ -50,8 +50,11 @@ Every stream, in every language, is the same lifecycle:
    true.
 2. **`update(handle, bar) → value` — once per closed bar.** Takes the handle
    and the new input(s) (OHLCV for multi-input functions) and always produces
-   the new current value. `update` performs **no allocation** — the handle is
-   sized at open.
+   the new current value. `update` allocates **no handle state** — the handle is
+   sized at open. On the managed backends a multi-output `update` also returns a
+   `Value`; C#'s is a record struct and costs nothing, and Java's is a record
+   *returned unstored*, which is what leaves it eligible for scalar replacement
+   rather than a guaranteed per-bar heap allocation (#310).
 3. **`close(handle)`.** Releases the stream — explicit in C, implicit
    in managed languages (Rust/Java).
 
@@ -492,8 +495,11 @@ OutRange r = s2.outRange();                      // the bars this handle has an
   `"<NAME> peek:"`), leaving the handle's *state* untouched
   (`docs/error-handling-spec.md` §2.4).
 - `value()` re-reads the value(s) at the last bar the stream counted — the bar
-  `outRange()` ends on — without recomputing; multi-output `update`
-  caches the immutable `Value` it returns, so `value()` is allocation-free.
+  `outRange()` ends on — without recomputing. Multi-output `value()` builds its
+  `Value` from those committed fields on each call rather than returning a
+  cached instance: the cache cost a guaranteed heap allocation on every
+  `update`, to make an accessor free that a caller invokes at most once a bar
+  and usually not at all (#310).
 - `clone()` is the universal deep copy (arrays cloned, sub-handles copied
   recursively, the `Core` reference shared), and it is spelled the same in all
   four backends: `TA_<N>_Clone`, `.clone()`, `clone()`, `Clone()`. It is the
@@ -536,9 +542,10 @@ emitter and accepted the later delivery date.
 One place the C# tier deliberately departs from the Java one, because the
 language offers something Java does not:
 
-- **Multi-output values are a `readonly record struct`, not a class.** Java
-  caches the boxed `Value` so that `value()` allocates nothing; a struct is
-  allocation-free by construction, so C# has no `cachedValue` field at all.
+- **Multi-output values are a `readonly record struct`, not a class.** A struct
+  is allocation-free by construction; Java's `Value` is a real object, so the
+  best Java can do is never store it (#310) and leave the JIT free to
+  scalar-replace it — a JIT decision, where C#'s is a language guarantee.
   The consequence for callers is that C# equality is .NET's `double` equality
   — `NaN` equals `NaN` **and** `+0.0` equals `-0.0` — where Java's record
   compares bitwise and disagrees on the second. Compare
--- a/ta_codegen/generator/src/backends/csharp_stream.rs
+++ b/ta_codegen/generator/src/backends/csharp_stream.rs
@@ -33,11 +33,12 @@
 //!   `Core` instance method. Measured, `core.Step(sp, x)` versus
 //!   `this.Step(x)` is 3.44–3.54 against 3.39–3.47 ns/bar — indistinguishable.
 //!
-//! - **No `cachedValue` field.** Java caches the boxed multi-output `Value` so
-//!   that `value()` allocates nothing; a `readonly record struct` return is
-//!   0 B/update by construction (measured against 40 B/update for the
-//!   Java-shaped class). One fewer field, one fewer store per bar, and one
-//!   fewer thing the copy path can get wrong.
+//! - **No `cachedValue` field.** A `readonly record struct` return is
+//!   0 B/update by construction, where a class return is 0 B only if the JIT
+//!   proves it does not escape. Java once cached the boxed multi-output
+//!   `Value` in a field so that `value()` was a field read, and paid 40 B per
+//!   `update` for it; #310 dropped that field and brought Java to this shape,
+//!   so the two managed backends now agree here rather than differing.
 //!
 //! - **The `NameMap` prefixes are Java's, verbatim** (`sp.x`, `sp.cur_y`,
 //!   `sp.ring_v_a`, ...). This is load-bearing, not cosmetic:
@@ -286,7 +287,8 @@ fn field_type_and_default(ty: &VarType) -> (String, String) {
 ///
 /// There is no `cachedValue`: the C# `Value` is a `readonly record struct`
 /// returned by value, so caching it would cost a field and a store per bar and
-/// buy nothing.
+/// buy nothing. Java reached the same conclusion the other way round, by
+/// measuring what the store cost it (#310).
 fn base_fields(func: &FuncDef) -> Vec<Field> {
     let mut fields: Vec<Field> = Vec::new();
     for p in &func.optional_inputs {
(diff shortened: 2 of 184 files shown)


kevinlincg:
Rebased onto `69284b1d5`, which now includes #307. One textual conflict, in `java_stream.rs`'s module doc: #307 rewrote the `peek` bullet (this branch still carried the "deep-copy constructor + step on the throwaway" wording you swept as stale), and this branch rewrote the `Value` bullet. I kept your peek bullet and mine for `Value`, so neither correction is lost.

Re-verified after the rebase, on the pinned toolchain: `generate` leaves the tree byte-unchanged, the diff against `dev` still touches `ta_codegen/output/java/` and nothing else (18 files — C, Rust and C# output remain identical), `cachedValue` appears nowhere in any generated output, the generator suites are 874 pass / 0 fail, and `clippy --all-targets -- -D warnings` is clean.

What I did **not** re-run: the JMH-style allocation and timing tables in the description above were measured against `f5909c8b`, before #307. #307 changes `peek`, not `update`, so the `update` rows should be unaffected — but that is an argument, not a measurement, and I have not re-measured them on the new base. The mechanism claim (0 B/update on the 15 multi-output functions) is re-checkable from the generated source; the ns columns are not.

kevinlincg:
Rebased onto `0e1b1228`, which brings in #287's C `Clone`/`Value` work and the `copy()` → `clone()` rename across all four backends.

Two textual conflicts, both in prose, both resolved by keeping each side's own correction rather than one side wholesale:

- `docs/streaming-api-design.md` — `dev` rewrote the deep-copy bullet for the rename (and the reasoning about `Cloneable`/`super.clone()`); this branch rewrote the `value()` bullet. Kept `dev`'s `clone()` bullet and this branch's `value()` bullet.
- `java_stream.rs` module doc — same shape: `dev`'s `peek` bullet (updated again since my last rebase), this branch's `Value` bullet.

**One line in the description above is now stale.** It says "`copy()` copies one fewer field". The method is `clone()` as of this base; the substance is unchanged — the handle carries one fewer field, so whatever copies it copies one fewer — but the name in that sentence is wrong and I would rather flag it than quietly rewrite the description after you may have read it.

Re-verified on the new base, pinned toolchain: `generate` leaves the tree byte-unchanged; the diff against `dev` touches `ta_codegen/output/java/` and nothing else (18 files); `cachedValue` appears in no generated output; `Core.java` has 0 `copy(` and 176 `clone(`, so the rename is consistent and this branch did not resurrect the old spelling; generator suites 874 pass / 0 fail; `clippy --all-targets -- -D warnings` clean.

Still not re-measured on the new base: the allocation and timing tables were taken against `f5909c8b`. #287 and #307 change `peek` and the copy path, not `update`, so the `update` rows should be unaffected — an argument, not a measurement, and the same caveat I flagged on the previous rebase still stands.

mario4tier:
A better solution is being proposed in #310 — please check there for details before continuing this PR.

Issue #322, opened 2026-09-01
test(java): the cross-language sweep measures the server's spliced Core, and the shipped Core.java is value-checked only near the default parameters


mario4tier:
Java is the only backend whose cross-language sweep measures a **copy** of the library
instead of the shipped artifact, and the one gate that does reach the shipped
`io.github.talib.Core` only drives parameter vectors near the declared defaults. A
defect in the shipped Core that needs a *relationship* between two optional parameters
is therefore invisible.

Found while adding `TA_KC` (#273), whose two periods have exactly such a relationship.

## The asymmetry

| backend | what the server's direct methods execute |
|---|---|
| Rust | `use ta_lib::{Core, ...}` — the shipped crate |
| C# | `using TALib;` — csproj compiles the shipped library sources (`partial class Core`) |
| **Java** | its **own inline `class Core`**, spliced by `inline_java_core_methods()` (`server_gen.rs:191`) |

`generate_csharp_server`'s own doc comment (`server_gen.rs:4551-4556`) names the splice as
the thing C# avoids:

> compiles the **shipped library sources directly** … not a spliced copy — so what the
> cross-language harness measures is byte-for-byte the shipped code. That is the same-text
> identity proof `inline_java_core_methods` gives Java, without the splice

The splice keeps the two texts equal *by construction* (both come from the same
`fragments/Core_<NAME>.java`), so this is not a claim that they compute different things.
It is a claim about **what the gate would notice if they ever did**.

## What actually reaches the shipped `Core.java`

Only `abstract_call` (`TaCodegenServe.handleAbstractCall`), which goes through
`io.github.talib.metadata.Functions` / `ParamHolder`. Its parameter coverage is
`callWithDefaults` → metadata defaults, plus `d2_non_default()`
(`test_abstract.c:1374`), which perturbs each slot to `default + (slot+1)`.

For `KC` (`optInTimePeriod` default 20, `optInATRPeriod` default 10) that is (20,10) and
(21,12). KC's centre-line anchor offset is `max(N-1, M) - (N-1)`, non-zero only when
`M > N-1`. Both vectors give 0. The sweep never reaches the case.

## Reproduction (measured, on `dev` + the #273 branch)

1. In the shipped `output/java/library/.../Core.java`, replace KC's
   `emaOffset = lookbackTotal - emaLookback;` with `emaOffset = 0;` (3 sites).
   Do **not** touch the input or the fragments.
[...]

Issue #164, opened 2026-08-04
ta_abstract: behavioural + coverage parity across C / Java / C# / Rust (umbrella)


mario4tier:
Umbrella issue for the `ta_abstract` layer across C / Java / C# / Rust: the
behavioural divergences, the coverage holes that let them survive, and the half
of the layer Rust never shipped.

Found while closing #162. Sources: a six-dimension audit (69 agents, 63 findings,
19 confirmed / 44 confirmed-with-corrections / 0 refuted), plus hand-verification
of everything marked **[verified]** below.

## The rule this is measured against

**The contract must match; the API shape adapts to each language.** An overload
set where C could only afford one function, an exception where C returns a
`TA_RetCode`, a `Result` instead of an out-param — these are adaptations to a
backend's canonical idiom, not divergences, and extra type-robustness is welcome.
What must not differ is *which calls are accepted and what numbers come back*.

That splits everything below into three buckets: **fix**, **document as
intentional**, and **needs a ruling**.

---

# A. Contract divergence — fix

### A1. Java `ParamHolder.setOptInput(int,int)` rejects `TA_INTEGER_DEFAULT` **[verified]**

`ParamHolder.java:179` range-checks `value < 0 || value >= MAType.values().length`
before storing, so `Integer.MIN_VALUE` throws. C's `TA_SetOptInputParamInteger`
performs no validation and the function substitutes the declared default.
Java's own *unset* path (`ParamHolder.java:275-276`) already resolves an
`INTEGER_LIST` to its declared default, so unset and explicit-sentinel disagree
**with each other**.

C# could route the sentinel through to the typed API because a C# enum is an
`int`; Java's `Core` takes a real `MAType`, so Java must resolve at the holder
boundary — which is where it already resolves the unset case.

### A2. Java binds a choice list by enum **ordinal**, not by declared **value**

`ParamHolder.java:188` (`maTypeOpts[idx] = all[value]`) and `:206`
(`intOpts[idx] = value.ordinal()`). C, C#, and the shared row model all use the
declared value from the `IntegerList` domain. Identical today only because
`MAType`'s values happen to be `0..10` with no gaps. A choice list whose values
are not dense-from-zero silently binds the wrong member.

Related: `ParamHolder` hardcodes `MAType` as *the* integer-list type in two
places, so a second `enum:` domain would compile and misbehave. C# guards the
equivalent assumption with an assertion that all choice lists in the corpus are
identical; Java has no such guard.

[...]

PR #323, opened 2026-09-01
test(abstract): drive both orderings of every integer parameter pair (#322)

kevinlincg:
Takes option 2 of #322 — the parameter-interaction half. It does **not** touch the Java splice (option 1), which is the real fix for the asymmetry and is a separate change.

## What was missing, verified rather than quoted

`d2_param_vectors`'s three classes are each a one-slot move:

    non-default       d2_non_default(oi, k) → def + (k+1) for every integer slot
    default-sentinel  one slot at a time set to TA_INTEGER_DEFAULT
    out-of-range      one slot at a time set to min-1 / max+1

So the only multi-slot vector in the sweep is the non-default one, and it moves every slot by a *fixed* per-slot offset. Neither shape can express a relation between two slots. For the KC case in the issue that is (20,10) and (21,12), both on the same side of `optInATRPeriod > optInTimePeriod - 1`.

## The change

A fourth class: for every pair of `TA_OptInput_IntegerRange` slots, one vector per ordering, every other slot at its default. Derived from the declared domains, so a new function or a new integer pair is covered the day it lands with no list here. Both directions per pair, because which side of a relation matters is a property of the function and this sweep does not know it.

`d2_lo_hi` picks `min+2` / `min+9`, clamped, falling back to the full range when the domain is narrower, and returns 0 when there is no room for two distinct values — a slot that cannot be ordered is skipped rather than failing. Small on purpose: the point is the relation, and a large period only buys runtime. For KC that would be (4, 10) — the exact vector the issue's reproduction uses.

**62 vectors** over the corpus, on top of the 457 already driven:

    Binder parameter contract (#164): 519 vector(s) driven through both binders
      — 84 non-default, 133 default-sentinel, 240 out-of-range, 62 ordering

The count is asserted alongside the other three in the non-vacuity guard, for the reason already written there: a sweep that quietly stopped producing a class reads exactly like a passing one.

## Control

A throwaway divergence in the C server's `handle_abstract_call` (`templates/c/ta_abstract_serve.c`): for `ADOSC` only, when the first integer slot exceeds the second and both are in a plausible band, the server perturbs the second slot before calling. That is an ordering-conditional defect and nothing else.

    with the ordering class     ABSTRACT ERROR [ADOSC]: real output[0][0]
                                C=-0.380571049705841 server=-0.320904805316095
                                ABSTRACT ERROR [ADOSC]: the ordering vector diverged
    without it (this commit
    reverted, sabotage kept)    519 → 457 vectors, 0 ordering, run passes

**Two earlier versions of this control were not discriminating, and finding that out is most of why I trust this one.** My first sabotage fired on `v0 > v` alone: the default-sentinel vector sets a slot to `TA_INTEGER_DEFAULT`, a large negative, so `v0 > v` was true there too and the *old* sweep caught it — via the sentinel class, not the ordering one. Guarding with `v0 > 0 && v > 0` then let the out-of-range probe trigger it (`retCode C=2 server=0`), so the old sweep caught it again, that time via the reject class. Only the third guard — both values inside a plausible band — is inert for every vector the three existing classes produce. A control that goes red for the wrong reason proves nothing about the class it is supposed to be exercising.

## What I did not check

- **No non-C language server ran here.** `--language=java` needs JDK 17 to build (`error: release version 17 not supported` on this box's JDK 11), and the C# SDK is 7.0.200 against a `net10.0` library. So the sweep ran C-against-C, which is identity on the value path — what it proves locally is that the vectors are produced, are accepted by the binder, and that a divergence injected at that ordering is detected. **The cross-language value comparison this class exists for is exercised for the first time by CI on this PR.**
- **The `--codegen` run does not complete on this box**: it stops at `cannot start ta_ref_serve (the reference oracle)`, which is built from the pinned `reference-pre-cutover` worktree I do not have. The abstract sweep runs to completion before that point — 519 vectors, no `ABSTRACT ERROR` — and plain `ta_regtest` passes in full.
- **Runtime cost is unmeasured on a runner.** 62 extra vectors against 457 is a ~14% increase in that sweep's vector count; I did not time it on CI hardware.
- I did not touch option 1 or option 3. Option 3 (asserting the two Java texts are identical at generate time) is cheap and independent of this; happy to send it separately if you want it.

### DIFF
--- a/src/tools/ta_regtest/test_abstract.c
+++ b/src/tools/ta_regtest/test_abstract.c
@@ -94,6 +94,8 @@ static long long g_d2Vectors = 0;
 static long long g_d2NonDefault = 0;
 static long long g_d2Sentinel = 0;
 static long long g_d2Reject = 0;
+/* Pairs of integer slots driven at both orderings; see d2_param_vectors step 1b. */
+static long long g_d2Ordering = 0;
 /* Non-finite parameter probes; see d2_nonfinite_params. */
 static long long g_d2NonFinite = 0;
 static long long g_d2NonFiniteFuncs = 0;
@@ -167,6 +169,7 @@ void test_abstract_set_server(CodegenPipe *cp, const char *lang)
    /* Per-server, so the summary line reports the server it names rather than a
       running total across every server tested so far. */
    g_d2Vectors = g_d2NonDefault = g_d2Sentinel = g_d2Reject = 0;
+   g_d2Ordering = 0;
    g_d2NonFinite = g_d2NonFiniteFuncs = 0;
    g_d2OorNotRejected = g_d2SentNotDefault = 0;
    if( cp )
@@ -1462,6 +1465,7 @@ static void d2_set_opts( TA_ParamHolder *paramHolder, const TA_FuncHandle *handl
 #define D2_CLASS_NON_DEFAULT 0
 #define D2_CLASS_SENTINEL     1
 #define D2_CLASS_REJECT       2
+#define D2_CLASS_ORDERING     3
 
 /* The all-defaults C result, captured before the sweep so the sentinel class can
  * assert what it exists to assert. */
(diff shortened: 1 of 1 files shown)


kevinlincg:
Rebased onto `d8af8b5d`, and changed to match the direction that commit sets rather than merely resolving the conflict.

The conflict was the two `printf`s reporting the sweep's counts on success — which `d8af8b5d` deletes, since its whole point is that a passing run should say nothing. I kept your version: no success output.

That left my ordering count printed nowhere and asserted only as `!= 0`, which is weaker than the standard the same commit sets a few lines above (`g_d2NonFinite < 144 || g_d2NonFiniteFuncs < 14`, with the comment explaining that a derived count moves with the corpus and lets a sweep go quiet unnoticed). So the ordering class now carries a **literal floor** too:

    || g_d2Ordering < 64

**64, not the 62 in the PR description above** — KC landed in `c5dd353a` and brought two more integer parameters into a pair, so the class picked it up with no edit. That is the generator-driven property working, and it is also why the floor has to be literal: had it been derived, the same corpus growth would have moved the bar with it and proved nothing.

Re-verified on the rebased branch: the abstract sweep runs to completion with no `ABSTRACT ERROR`, and plain `ta_regtest` passes in full. The `--codegen` run still stops afterwards at `cannot start ta_ref_serve` on this box, as noted originally.

The floor itself is control-tested the only way a floor can be: setting it to 99999 makes the guard fire and print `(531/85/136/246/64)`, which is where the 64 comes from.

PR #324, opened 2026-09-01
build(java): fail the server splice when the shipped Core has gone stale (#322)

kevinlincg:
Takes option 3 of #322 — the staleness half. Independent of #323 (option 2, parameter coverage) and of option 1, which stays the real fix.

## Where the check goes, and why there

`build.py servers` re-splices the Java server from `fragments/Core_<NAME>.java` and does **not** rewrite `Core.java`. That is the window the issue names, and it is the moment the two texts can part — so the assert runs there, in `generate_server` for the Java backend, right after `inline_java_core_methods`.

Not in a test: on a clean checkout there is nothing to catch, which is exactly why `regen-check` does not backstop this (it also excludes the files being worked on, as the issue notes).

## The check is equality, not containment — and that mattered

My first version compared per line: every fragment line must appear somewhere in `Core.java`. It passed its own control, and it was wrong. I sabotaged `Core.java` by appending ` // STALE` to a line, ran it, and it stayed green — because `"foo // STALE"` still *contains* `"foo"`. **Containment cannot see an addition**, which is a large share of the drift worth catching.

The shipped version compares the whole GENCODE section against the concatenated fragments for equality, normalized on both sides the way `inline_java_core_methods` already normalizes (drop the `/* Generated */` prefix, drop blank lines, trim), since the two differ in indentation by construction. It reports the first differing generated line with both sides and the two lengths.

## Control

Three shapes of drift, each applied to `Core.java` alone and then reverted, with `build.py servers --language=java`:

    clean tree   0 false positives
    append       fires, first difference at generated line 3602
    modify       fires, first difference at generated line 3601
    delete       fires, first difference at generated line 3601

The append row is the one the first version missed.

## Verification

- Full `scripts/build.py generate` then `git status`: no drift outside the two generator files this commit touches.
- `cargo test --no-fail-fast` in `ta_codegen/generator`: 884 pass, 0 fail.
- `cargo clippy --all-targets -- -D warnings`: exit 0.

## What I did not check

- **The Java library does not compile here** — this box's JDK is 11 and the pom asks for release 17 (`error: release version 17 not supported`), so `build.py servers --language=java` fails at maven *after* the generation step. The assert runs before that and its output is what the control above reads; nothing downstream of maven was exercised.
- The `--func=` path is unaffected on purpose: it prints `skipping shipped Core.java` and does not generate the server, so it never reaches this. This makes the stale state loudly visible the next time a server is built, rather than blocking the partial-regenerate workflow the message documents. If you would rather `--func=` itself warn, that is a different and slightly noisier change.
- Only the Java backend. C# compiles the shipped sources directly and has no splice to check; C and Rust likewise.

### DIFF
--- a/ta_codegen/generator/src/backends/mod.rs
+++ b/ta_codegen/generator/src/backends/mod.rs
@@ -352,6 +352,18 @@ impl LanguageBackend for JavaBackend {
         std::fs::create_dir_all(&tools_dir).unwrap();
         let template = server_gen::generate_java_server(funcs, enums);
         let output = server_gen::inline_java_core_methods(&template, &frag_dir, funcs);
+        // The server and the shipped library are the same text only for as long
+        // as both are regenerated together. `generate --func=<NAME>` rewrites a
+        // fragment and deliberately skips Core.java, and `build.py servers`
+        // re-splices the server from the fragments — so in that window the
+        // cross-language sweep measures new code while the shipped library is
+        // old, and `--codegen --language=java` reads green (#322).
+        //
+        // Checked here rather than in a test: this is the moment the server is
+        // built from the fragments, so it is the moment the two can part. The
+        // fragment is the same text `java_shipped` splices into Core.java's
+        // GENCODE section, so containment is exact rather than approximate.
+        server_gen::assert_java_core_matches_fragments(&frag_dir, out_base, funcs);
         let path = tools_dir.join("TaCodegenServe.java");
         crate::emit::write_if_changed(&path, &output).unwrap();
         println!("  Java server -> {}", path.display());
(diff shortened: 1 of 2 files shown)


mario4tier:
Thanks for this — and in particular for writing up why containment failed its own control. That the check has to be *equality* is the part that mattered, and it shaped what landed.

We took a different route in the end, so I'm closing this one. Rather than comparing the two texts at generate time, both artifacts now carry a digest of the generated method text and `ta_regtest` compares them over the wire before it measures anything — so the check reads the loaded bytecode instead of source on disk, and a stale class directory or a `cargo run -- build` that skipped regeneration is visible too. The generator asserts the same pair at splice time, which is the place your version occupied.

One measurement from your PR's ground worth recording: the fragments concatenate byte-identically to the GENCODE section, so no normalization is needed at all — dropping it makes the check strictly stronger, since whitespace-only drift is then caught rather than excused.

Details in https://github.com/TA-Lib/ta-lib/issues/322#issuecomment-5499688094. Landed in `3aab6fde7`.

mario4tier:
Superseded by 3aab6fde7 — see the note above.

Issue #325, opened 2026-09-01
perf(streaming): two composed Java peek frames allocate a sink per call, and only a frame-emitter change removes it (#310 follow-up)


kevinlincg:
Splitting this out of #310 at the maintainer's request, so the analysis is on record rather than buried in a comment thread. **Not proposing to do it now** — the site count does not justify it yet.

## The residue

After #310 lands (Java multi-output `update`/`peek`/`value` taking a caller-owned `<N>Out`), Java streaming allocates zero bytes per bar everywhere **except two composed peek frames**:

    MamaStream.peek     called from the MA-dispatch tier's peek
    StochfStream.peek   called from STOCHRSI's peek

Those two allocate one `<N>Out` per call, as a per-call local. That is not a regression #310 introduces — they allocate today, for the same reason.

## Why the local is unavoidable at the API level

A `peek` must not write the handle (the javadoc promises peeks may run concurrently with each other, in every backend), and Java has no out-parameters for primitives. So N values can leave a call by exactly three routes: an object, the handle, or no call at all. The first is the local; the second breaks the concurrency promise; the third is this issue.

## Why escape analysis does not save it

Measured on the shipped jar (bytecode size vs the default `FreqInlineSize` of 325 bytes):

| method | bytecode | |
|---|---:|---|
| `MamaStream.peek` | ~1354 B | 4x over |
| `StochfStream.peek` | ~698 B | 2x over |
| `MacdStream.peek` | ~165 B | under |
| `MacdStream.update` | ~94 B | under |

Neither composed callee can be inlined at default flags, so the sink crosses a call boundary C2 will not open and is allocated unconditionally. EA is not marginal at those two sites — it never fires.

## The fix, and its cost

Inline the sub-handle's peek transition into the outer frame, so the sub's outputs land in the outer frame's locals and there is no call to leave. No allocation, no shared scratch, no concurrency compromise.

The cost is that it is a **frame-emitter change bought for two call sites**. That is the trade this issue exists to record, not to make.

## Not blocked on, and not helped by, #321

Checked rather than assumed: MAMA's peek body is 226 lines with the last sink store at line 194, so the tail #321 deletes is ~13% — nowhere near the ~76% that would bring it under the inline budget. #321's win there is time (an `atan`), not bytecode size.

## When to revisit

If the number of composed multi-output peek sites grows. Today it is two. It may stay two indefinitely, in which case this should be closed rather than done.

mario4tier:
[...]

PR #326, opened 2026-09-01
perf(java): a multi-output stream writes a caller-owned Out (#310)

### LINKED ISSUE #310: Java streaming: multi-output update() allocates a Value record every bar because it is cached for value()

kevinlincg:
Closes #310, implementing the shape you settled on. Splits out #325 for the composed-peek residue, as asked.

## The API

    void update( double inReal, MacdOut out );   // commits, writes out
    void peek(   double inReal, MacdOut out );   // commits nothing, writes out
    void value(  MacdOut out );                  // writes the last counted bar

`MacdOut` is a mutable public-field class at `Core` level, a sibling of `MacdStream` — where C# already puts `MacdValue`. Single-output handles are byte-identical, and so are C, Rust and C#.

Both decisions you called are in:

- **No `equals`/`hashCode`.** The javadoc says why, and says the other half too: *"Its contents are only valid until the next call that writes it… a reference kept past that call, or one put in a collection, sees the value change underneath it."*
- **Composed peek: one `Out` per call.** Two sites, `ma` and `stochrsi`.

**Composed `update` allocates nothing at all** — it commits, so the forwarded value comes off the sub-handle's own `cur_*` field and there is no sink on that path. That is better than the local I had proposed for symmetry, and it follows from your own observation on the issue.

`updateAndFill` loses the `done` / `try` / `finally` that existed only to keep the cache fresh on a partial fill.

## The gates, and one made stricter rather than looser

`no_java_peek_copies_the_handle` went red on the two composed sinks. **I did not exempt the allocation.** The sweep now pins the set of composed peeks that allocate to *exactly* `{ma, stochrsi}`:

- a third site fails — a new composed multi-output peek nobody costed;
- **losing one also fails** — that is #325 landing, and the bound should tighten with it rather than silently over-permitting.

`the_java_value_cache_is_refreshed_on_every_exit` asserted a mechanism that no longer exists. Replaced, not deleted, by `no_managed_handle_caches_the_multi_output_value`: a sweep over Java and C#, over `updateAndFill` and the whole handle, for any cache or `finally` — with a non-vacuity check that the handles it sweeps really are multi-output, since a single-output handle's lack of a cache proves nothing.

The three remaining expectation updates are contract changes: the record → the Out class, and the dispatch `update` path now reading `cur_outMAMA` with an explicit assertion that `MamaStream.Value` appears nowhere.

## Verification

- `cargo test --no-fail-fast`: **884 pass, 0 fail**.
- `cargo clippy --all-targets -- -D warnings`: exit 0.
- Full `generate` twice: idempotent, no drift.
- Diff touches `ta_codegen/output/java/` and three generator files. **Zero files under `output/c`, `output/rust`, `output/csharp`.**
- `cachedValue` and `record Value(` appear nowhere in the generated Java.
- Rebased onto `c5dd353a` (KC). KC is multi-output and picked up `KcOut` with no work — the emitter is driven off the corpus, so a new indicator is covered the day it lands.

## What I did not check

- **Nothing Java was compiled or executed.** This box's JDK is 11 and the pom asks for release 17 (`error: release version 17 not supported`), so `build.py servers --language=java` fails at maven, and the four-language regtest cannot run its Java leg here. Everything above is generator-side: text assertions, the sweeps, and clippy. **The first compile and the first bit-exactness run of this change are the CI on this PR.** That is a real gap on a change of this size and I would rather name it than let the green ticks above imply more.
- **The allocation claim is not re-measured.** The 0 B/bar for the 15 multi-output `update`s follows from there being no allocation site left in the generated text, which is checkable by reading it — but I have not run `ThreadMXBean` against a built jar, for the same reason.
- **`StreamSmokeTest` is not updated here.** It lives in the Java library sources, which I cannot compile to check I have not broken it; its identity assertions on the returned `Value` will need to become field comparisons. Say the word if you want that in this PR anyway and I will write it blind and mark it as untested.

### DIFF
--- a/ta_codegen/generator/src/backends/java_stream.rs
+++ b/ta_codegen/generator/src/backends/java_stream.rs
@@ -105,6 +105,14 @@ pub fn stream_class_name(func: &FuncDef) -> String {
     format!("{}Stream", common::pascal_words(&func.name))
 }
 
+/// The caller-owned output object a multi-output stream writes through.
+/// A sibling of the stream class at `Core` level, where C# already puts its
+/// equivalent -- it is the type the CALLER allocates, so it should not need the
+/// stream class to name it (#310).
+pub fn out_class_name(func: &FuncDef) -> String {
+    format!("{}Out", common::pascal_words(&func.name))
+}
+
 fn out_is_int(func: &FuncDef, name: &str) -> bool {
     func.outputs
         .iter()
@@ -247,7 +255,7 @@ fn field_type_and_default(ty: &VarType) -> (String, String) {
     }
 }
 
-/// The params + `cur_<out>` (+ cachedValue) fields every tier's handle carries
+/// The params + `cur_<out>` fields every tier's handle carries
 /// (dispatch/period-bank/loopless-composed build on exactly this base).
 fn base_fields(func: &FuncDef) -> Vec<Field> {
     let mut fields: Vec<Field> = Vec::new();
@@ -259,9 +267,6 @@ fn base_fields(func: &FuncDef) -> Vec<Field> {
     }
     for out in &func.outputs {
         fields.push((format!("cur_{}", out.name), out_java_type(func, &out.name).to_string(), "0".into()));
-    }
-    if has_value_class(func) {
-        fields.push(("cachedValue".into(), "Value".into(), "null".into()));
     }
     fields
 }
@@ -385,9 +390,6 @@ fn state_fields_from(
     for out in &func.outputs {
         let t = out_java_type(func, &out.name);
         fields.push((format!("cur_{}", out.name), t.to_string(), "0".to_string()));
-    }
-    if has_value_class(func) {
-        fields.push(("cachedValue".into(), "Value".into(), "null".into()));
     }
     fields
 }
@@ -730,10 +732,13 @@ fn emit_handle_class_with_members(
     let _ = writeln!(o, "         this.outRangeCount = other.outRangeCount;");
     let _ = writeln!(o, "      }}");
 
-    emit_value_class(o, func);
     emit_update_peek_value_copy(o, func, frame);
 
     let _ = writeln!(o, "   }}");
+
+    // At Core level, a sibling of the stream class: it is the type the CALLER
+    // allocates, so it should not need the stream class to name it (#310).
+    emit_out_class(o, func);
 }
 
 /// The immutable multi-output value record (batch output order, components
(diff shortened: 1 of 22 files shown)


kevinlincg:
Rebased onto `19f95fe9` (KC's per-leg lookback fix).

Every conflict was in `ta_codegen/output/java/` — the generator source applied clean. That is the split working as intended: `java_stream.rs` and the two suites needed no reconciliation, so the resolution was to take `dev`'s outputs and re-run `scripts/build.py generate` rather than merge generated text by hand.

Re-verified on the new base: `generate` twice is idempotent with no drift, the diff still touches `ta_codegen/output/java/` and the three generator files only (**zero files under `output/c`, `output/rust`, `output/csharp`**), `cargo test --no-fail-fast` is 884 pass / 0 fail, and `clippy --all-targets -- -D warnings` exits 0.

KC still carries `KcOut` after its fix, which is the point worth noting: it picked the Out shape up when it landed and kept it through a change to its own seeding, with no edit here either time.

The verification gap in the description is unchanged and still the important one — nothing Java has been compiled or run on this box (JDK 11 against a pom asking for release 17), so this PR's CI remains the first compile and the first bit-exactness run of the change.

kevinlincg:
`2e4171ce1` gave `build.py` a `ta_ref_serve` target, so the local gap in the description — `--codegen` stopping at `cannot start ta_ref_serve` — is now closable. I built the oracle and ran it on this branch:

    scripts/build.py ta_ref_serve     ta_ref_serve: OK (from pinned-tag worktree)
    ta_regtest --codegen --language=c C: 161 passed, 0 failed, 16 skipped
                                      * All tests succeeded. *

**That is not evidence for this PR, and I am not claiming it as such.** I ran the same thing on `dev` with the branch checked out nowhere near it, and it prints the identical line — the C leg never touches a Java artifact, so it cannot distinguish this change from its absence. What it establishes is only that the branch does not break the C tier, which was never in question since the diff has no file under `output/c`.

The verification that matters is still the Java leg, and it is still unreachable here: JDK 11 against a pom asking for release 17. So the position is unchanged — **this PR's CI is the first compile and the first bit-exactness run of the change**, and the allocation claim remains derived from reading the generated text rather than measured on a built jar.

Mentioning the run rather than staying quiet because a `--codegen` line in a log looks like coverage, and on this change it is not.

Separately: I re-ran the merge into `2e4171ce1` locally and it is clean, so no rebase is needed for the `-Wall -Wextra` and oracle commits.

Issue #328, opened 2026-09-02
perf(streaming): a peek frame keeps arithmetic whose only readers the tail took with it


mario4tier:
Follow-up to #321, which is where these six warnings come from.

Trimming a peek frame at its last output store orphans locals whose only readers
were in the deleted tail. Nothing prunes them: `temps_used` filters declarations
with `names_mentioned`, and an assignment target counts as a mention.

`gcc -fsyntax-only -Wall -Wextra -Wno-unused-parameter -Wno-unknown-pragmas`
(the set both build systems compile the library with — CMakeLists.txt, configure.ac):

```
src/ta_func/ta_CORREL.c:1133    'trailingX' set but not used
src/ta_func/ta_CORREL.c:1134    'trailingY'
src/ta_func/ta_HT_PHASOR.c:1490 'Q2'
src/ta_func/ta_HT_PHASOR.c:1491 'I2'
src/ta_func/ta_MAMA.c:1727      'Q2'
src/ta_func/ta_MAMA.c:1728      'I2'
```

Zero before #321. No `-Werror` on C anywhere, so this is warning noise in the
shipped library and in the release tarball, not a break. The other three backends
are silent for their own reasons — the generated Rust crate allows the lint
wholesale, C#'s CS0219 does not fire on a non-constant assignment, javac does not
warn on unused locals — so the wasted work is in all four and only C says so.

Not free arithmetic. `TA_HT_PHASOR_Peek` still runs
`Q2 = fma(0.2, Q1 + jI, 0.8 * sp->prevQ2)` and its `I2` twin in **both** Hilbert
arms, four fused multiply-adds computed and discarded; MAMA the same; CORREL
rebuilds `trailingX`/`trailingY` from two extrema-ring loads and two shadow
selects.

## Why #321 did not do it

Deleting a dead store is easy. Deleting THOSE dead stores costs the gate that
watches the frame.

`a_peek_frame_fuses_the_same_multiply_adds_as_its_update_frame` was re-anchored in
#321 from equality to a PREFIX check, on the strength of peek's fused sites being
a character-for-character leading run of update's. Prefix is what keeps it sharp:
a site re-fused, dropped or reordered in the MIDDLE still fails, which is the
hazard the gate exists for (`expr_is_float_typed` classifies a `Ternary` by its
THEN arm, so a shadow select with the local in the then-arm would silently unfuse
a multiply-add update still fuses — a ~1 ULP divergence in a comparison ta_regtest
[...]

PR #329, opened 2026-09-02
perf(streaming): prune a peek store whose readers left with the tail (#328)

kevinlincg:
Takes the "cheapest thing" line of #328: prune only the orphans whose defining expression is fusion-free. CORREL's two warnings go; HT_PHASOR's and MAMA's four stay, for the reason the issue records — their stores carry the multiply-adds the fusion gate anchors as a prefix of update's run.

## The pass

After `peek_tail_trimmed`'s cut, a kept top-level store is dropped only when **all** of:

- the target is a bare local identifier — not a qualified field write (`sp->x`), and never a `cur_*` localize;
- it is that name's only mention in the kept frame;
- the right side has no multiply and no call — decidable without importing `fma.rs`'s per-backend knowledge, since a fused site cannot come out of an expression with no multiply in it;
- the right side reads no other candidate (one pass, no fixpoint).

On the shipped corpus that is exactly CORREL's `trailingX`/`trailingY`, in all four backends. I read **every deleted line** of the regenerated diff rather than the numstat: the removals are those two stores and their declarations, C, Java, C# and Rust, plus the two `GENCODE_DIGEST` lines moving with the Java text (#322's mechanism, by design).

    gcc -fsyntax-only -Wall -Wextra (all 176 C files):   6 warnings -> 4

## Two of the four refusal arms earned their place by failing

Worth stating because the first version of this pass was wrong twice, and both times the numbers looked fine:

- **`cur_*` exclusion**: my first spelling was `starts_with("cur_")`, which matched nothing — the NameMap hands names back already qualified (`sp.cur_outReal`), so the pass was deleting output computations like `cur_outMACD = macdValue` and `cur_outRealUpperBand = tempUpper / period` (division: no multiply, no call). A peek would have answered the seed instead of the bar. The diff stats looked identical to the correct version's; only reading the deleted lines showed it.
- **bare-identifier restriction**: without it, the pass's own unit test deleted `sp->out = a` — a single-mention, fusion-free *sink* store. The corpus never showed this because real sinks are pointer derefs or `cur_*`; the constructed test is what caught it.

## The fusion guard is vacuous today, and its controls are unit tests

Sabotaging the fusion-free check changes **no generated byte** on the current corpus: every shipped orphan that survives the other three filters happens to be fusion-free (MAMA's `Q2`/`I2` never get that far — branch-nested, twice-mentioned). So a corpus diff cannot control this arm. Five unit tests over constructed transitions do: the positive prune, multiply kept, call kept, `cur_*` kept under all three qualifiers, reader kept, and candidate-reads-candidate kept.

## Verification

- Deleted lines audited one by one (above).
- `generate` twice: idempotent.
- Generator suite 896 pass / 0 fail; `clippy --all-targets -- -D warnings` exit 0.
- Full `ta_regtest`: pass. `--codegen --language=c` with the pinned oracle: 161 passed, 0 failed — **and this run is discriminating for this change**: CORREL's C peek is one of the compared legs, so its bit-exactness against batch ran against the pruned frame.

## What I did not check

- Java and C# are not compiled here (JDK 11 vs release 17; SDK 7 vs net10.0). The deleted lines in those backends are the same two dead local stores and their declarations — no signature, no reference elsewhere — so the compile-break surface is small, but per the lesson from #326 I would rather say it than imply it. The PR gate's javac step (added in `f2e1c637e`) covers the Java half of exactly this gap.
- HT_PHASOR/MAMA's four warnings remain by design. Closing them needs the different fusion-gate anchor #328 describes, which is not attempted here.

### DIFF
--- a/src/ta_func/ta_CORREL.c
+++ b/src/ta_func/ta_CORREL.c
@@ -1255,9 +1255,6 @@ TA_LIB_API TA_RetCode TA_CORREL_Peek( const TA_CORREL_Stream *stream, double inR
          ssY = 0.0;
       }
    }
-   /* Save the trailing values before writing the output, since the input
-    * and output might be the same array.
-    */
    sp->trailingIdx += 1;
    /* Output the new coefficient.
     *
--- a/src/ta_func/ta_CORREL.c
+++ b/src/ta_func/ta_CORREL.c
@@ -1130,8 +1130,6 @@ TA_LIB_API TA_RetCode TA_CORREL_Peek( const TA_CORREL_Stream *stream, double inR
    struct TA_CORREL_Stream *sp = &scratch;
    double x;
    double y;
-   double trailingX;
-   double trailingY;
    double ssX;
    double ssY;
    double spXY;
@@ -1260,8 +1258,6 @@ TA_LIB_API TA_RetCode TA_CORREL_Peek( const TA_CORREL_Stream *stream, double inR
    /* Save the trailing values before writing the output, since the input
     * and output might be the same array.
     */
-   trailingX = (((sp->trailingIdx & sp->xMask) != pkSlot0) ? sp->x_inReal0[sp->trailingIdx & sp->xMask] : pkVal0) - sp->shiftX;
-   trailingY = (((sp->trailingIdx & sp->xMask) != pkSlot1) ? sp->x_inReal1[sp->trailingIdx & sp->xMask] : pkVal1) - sp->shiftY;
    sp->trailingIdx += 1;
    /* Output the new coefficient.
     *
--- a/ta_codegen/generator/src/streaming.rs
+++ b/ta_codegen/generator/src/streaming.rs
@@ -7464,11 +7464,19 @@ fn peek_tail_trimmed(
 /// place, which is the trade #328 records.
 ///
 /// One pass, top level only, single-mention only: a name written twice (one
-/// store per branch arm) or read anywhere in the kept frame is left alone, and
-/// no fixpoint is attempted — a store this pass removes cannot orphan another,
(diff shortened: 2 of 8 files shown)


mario4tier:
Thank you for picking this up, and for the write-up — the "what I did not check" section and the
two bugs you caught in your own filter are exactly the kind of record that makes a PR easy to read.

I'm closing this one rather than merging, so you don't spend more time on it: we're landing a more
general form of the same idea, and it supersedes this cleanly.

The short reason is that the constraint this PR is designed around turned out to be removable. The
`fusion_free` filter exists to keep peek's fused sites a character-for-character *prefix* of
update's, which is what forced the trade of two warnings closed and four left. But `fma` is not an
`Expr::FuncCall` in this IR at all (`grep fma ta_codegen/generator/src/ir.rs` is empty) — fusion is a
render-time decision on a `BinOp` in `backends/fma.rs:404`, and the per-function site sets are built
once from `func.stream_source()` (`c_stream.rs:1154`, `java_stream.rs:421`, `csharp_stream.rs:442`)
and guard the emission of *both* frames. So a deletion pass can never re-classify a site peek still
evaluates; only the gate's anchor was in the way, not correctness.

So we re-anchored the gate instead of working around it, which let the purge take all six warnings
and the arithmetic behind them — the whole `jI`/`jQ` Hilbert chains in MAMA and HT_PHASOR fall out
once `Q2`/`I2` go, which needs the fixpoint the issue asked for.

Nothing here was wasted: your reading of the deleted lines one by one, rather than the numstat, is
the check that caught the `cur_` spelling, and it is the same check that mattered on our side.
Thanks again.

Issue #108, opened 2026-07-11
Multi-output functions don't reject output-buffer aliasing (same buffer for two output args)


mario4tier:
## Summary

Passing the **same buffer for two different output arguments** of a multi-output
function must return `TA_BAD_PARAM` (the outputs would clobber each other, so the
call has no correct result). Today every multi-output function instead returns
`TA_SUCCESS` and writes garbage.

This is a **pre-existing** gap — it is present in the released **0.6.4** library
(the `reference-pre-cutover` source contains only NULL checks, no output-distinctness
checks), so it predates the `ta_codegen` cutover. It is not a regression from any
recent change.

## Affected functions (14, all with ≥2 outputs)

`ACCBANDS`, `AROON`, `BBANDS`, `HT_PHASOR`, `HT_SINE`, `MACD`, `MACDEXT`,
`MACDFIX`, `MAMA`, `MINMAX`, `MINMAXINDEX`, `STOCH`, `STOCHF`, `STOCHRSI`.

## Reproduction

Calling any of them with two output pointers set to the same buffer returns
`TA_SUCCESS` (rc=0) instead of `TA_BAD_PARAM`:

```c
double A[200], D[200]; int b, n;
/* outMACD and outMACDSignal are the SAME buffer */
TA_RetCode rc = TA_MACD(0, 199, in, 12, 26, 9, &b, &n, A, A, D);
/* rc == TA_SUCCESS today; must be TA_BAD_PARAM */
```

A red test over all 14 functions and every output pair (21 combinations) shows
**21/21** returning success today.

## Scope of the fix

- The generator emits an output-distinctness guard in the **guarded** public
  function of every multi-output indicator, in **all four backends** (C, Rust,
  Java, .NET): `if (outA == outB || outA == outC || outB == outC) return TA_BAD_PARAM;`
  (pointer identity for C/Java/.NET; slice base-pointer identity for Rust).
- The **unguarded** cross-indicator variants keep skipping validation (internal
  calls always pass distinct buffers).
- **Input↔output aliasing is unaffected**: many functions document in-place
  operation (`inReal == outReal`), which stays allowed. Only *output↔output*
  aliasing is rejected.
- A **permanent, generic gate** in `ta_regtest` drives every function reported by
  `ta_abstract` as having ≥2 outputs with each output pair aliased and asserts
  `TA_BAD_PARAM`, so the invariant holds for all current and future multi-output
  functions.

## Notes

[...]

PR #341, opened 2026-09-03
feat: DONCHIAN — Donchian Channels with the original lagged window (proposal-drafts #38)

kevinlincg:
feat: DONCHIAN — Donchian Channels with the original lagged window (proposal-drafts TA-Lib/ta-lib#342)

Implements TA-Lib/ta-lib#342 as ruled: three overlap outputs
(`outRealUpperBand`/`MiddleBand`/`LowerBand`), `optInTimePeriod` default 20,
and an explicit `optInLag` defaulting to **1** — the preceding-window rule —
with `optInLag=0` reproducing the inclusive TradingView/NinjaTrader/pandas
form, where the three outputs are bit-identical to `MAX`/`MIDPRICE`/`MIN`.
Body is the cached-extremum loop `midprice`'s streaming tier runs, with the
scan cursor held `optInLag` bars behind the output bar. Lookback is
`(N-1)+L`. Docs carry the lag rationale and the "arriving from
TradingView/pandas: pass `optInLag=0`" note; `midprice.md` cross-links back.

## Batch-only, and the generator gap it exposed

Per the proposal's note, `flags: [stream]` is left off — the `L`-bar
hold-back has no shipped streaming tier — making DONCHIAN the first
batch-only function in the tree. That path had rotted: `server_gen`'s bench
warm-up arms were emitted unconditionally ("Every function streams"), so all
four generated servers referenced `TA_DONCHIAN_Open`/`_OpenAndFill` symbols
the library does not export and failed to compile. Fixed in `server_gen.rs`
for C/Rust/Java/C#: a batch-only function answers
`{"retCode":0,"timing_ns":0,"unsupported_mode":1}` for `bench_mode != 0` —
the exact non-measurement the frozen C reference server already reports, so
`ta_bench` needs no change — and emits no warm-up arms.

Two hand-written Java suites carried the same universal assumption. Their
sweeps now measure against the set whose metadata carries
`FuncFlags.STREAMING`, **in both directions**: a streaming core missing its
opener/handle/openAndFill still fails, and a batch-only core that *grows* a
streaming surface now also fails (flag and generated surface disagreeing is
the same bug from the other side).

## Gates

New `src/tools/ta_regtest/ta_test_func/test_donchian.c`, registered in both
CMakeLists.txt and Makefile.am, every comparison **bitwise** — the bands are
selections of input values and the middle is an exact halving, so there is
no divergence for a tolerance to absorb:

1. **pandas 2.3.3 oracle at `optInLag=0`** on the 252-bar TA_SREF corpus:
   six pinned bars × 3 outputs from `rolling(n, min_periods=n).max()/.min()`,
   `DCM = 0.5*(DCL+DCU)` (bit-identical to `(upper+lower)/2.0`: commutative
   add, exact power-of-two scaling). Valid at lag 0 only, where the pandas
   window and ours coincide; the lag ruling is the proposal's, not an
   oracle's — no library ships the lagged default to compare against.
2. **Differential, 0 ULP**: `DONCHIAN(N,L)[i] == MAX(high,N)[i]`,
   `MIN(low,N)[i]`, `MIDPRICE(N)[i]` — aligned index-for-index by
   construction (`donNb == refNb - L`) — over
   `N ∈ {2,3,5,20,50,252} × L ∈ {0,1,2}`, both edges included.
3. **Deterministic edges**: all-flat collapses all three outputs onto the
   price (NaN-free by construction — the only division is by the literal
   2.0); an oversize lag returns 0 elements and TA_SUCCESS.
4. **Aliasing**: each of the three outputs over each of the two inputs (six
   cases) must equal the separate-buffer run — the write index never passes
   the oldest read index, the same argument as MIDPRICE — plus the #108
   two-outputs-one-buffer rejection.

Plus the `server_verify` cross-language leg on both oracle parameter sets.

**Controls, run and watched going red** (not asserted): dropping the lag
offset from the scan cursor (`winEnd = today`) fails the differential at
`[N=2 L=1]` on its first bar *and* fails the framework's own
`[DONCHIAN][inputNegData]` sweep; restored, everything green again.

## Verification

- `generate` idempotent (second run leaves the tree unchanged); full
  `regtest` across all four language servers green — **`All tests
  succeeded`** — on this box with JDK 21 and .NET 10 installed.
- Generator `cargo test --release`: 901 passed, 0 failed.
  `cargo clippy --release --all-targets` clean.
- `--xlang-hash`: DONCHIAN bit-identical across C/Rust/Java/C# on every
  swept shape/seed/range. The sweep as a whole reports 28 pre-existing
  `TA_HT_TRENDMODE` tolerance-lane mismatches (C-vs-Java) on this machine —
  macOS x86_64 + Temurin 21 libm, an integer output flipping at a threshold —
  unrelated to this change: `git diff dev` names no HT_* file, and the
  Linux CI Java gate is the arbiter there.
- Bench sanity: `DONCHIAN` ~1509ns C / ~1515ns Rust on the shipped corpus
  (vs c-ref's MIDPRICE-shaped baseline), nothing anomalous.

## Not in this PR

- **The streaming tier** — the proposal's own sequencing: ship batch-only,
  add `stream` as a follow-up once the `L`-bar hold-back has a tier to run
  on (the PVO precedent).
- The pandas oracle **server arm** already exists in your ta-lib-oracles
  repo per the proposal (`compute_DONCHIAN`, `SPECS:395`), reordering
  `(DCL, DCM, DCU)` into band order — nothing to add there, and the goldens
  above were captured directly from pandas, not through it.


mario4tier:
Merged into `dev` as-is (`e0e323122`, `6e018348b` — your commits, unmodified), with a follow-up on top (`221fcd633`) that **removes `optInLag`** and ships the inclusive window with `flags: [overlap, stream]`.

The reason is one you surfaced yourself: the lag is what made DONCHIAN batch-only. `stream-census` rejects the lagged body at stage 1 — *"extrema automaton: expected exactly one window-start variable, found `[trailingIdx, winEnd]`"* — because the scan cursor trails the paced variable. Drop the lag and it reports `T4 state=7 lags=0 outs=3`, identical to MIDPRICE, and the whole `_Open*` family plus Update/Peek/Clone/UpdateAndFill generate with **no emitter change at all**. Weighed against a parameter that no other library exposes (ta4j, trading-signals, pandas-ta and TradingView all ship the inclusive window) and that is a pure output shift, being the tree's only non-streaming function wasn't worth it. Rationale in full on #342.

Consequence for your `server_gen.rs` commit: with all 179 functions streaming again, the batch-only path has no call site, so I reverted that file to its pre-PR state rather than carry unexercised branching in the one file that reaches all four backends. Your commit stays in history to cherry-pick the day a genuinely batch-only function lands — the analysis in its message is the valuable part and it holds. **Your two Java suite changes are kept**: measuring against the `FuncFlags.STREAMING` set in both directions is a better gate whether or not a batch-only function exists, and both suites now sweep 179/179.

Your cached-extremum body and test scaffolding carried over unchanged apart from the signature. I added a second oracle leg (ta4j-core 0.22.6 — different language and codebase, so it isn't two wrappers over the same `rolling()` call), the dual-alias case, and per-leg vacuity floors.

Thanks — this was a well-built PR, and the batch-only gap you found in `server_gen.rs` was real.

kevinlincg:
Merged manually as 12028998f (the PR head was a squashed re-push of the same tree, so GitHub could not auto-detect it) — closing this copy. The lag call makes sense: a full streaming tier for free beats a shift parameter nobody else ships, and the census error message pinpointing the second window variable is exactly the diagnostic I lacked. Glad the STREAMING-set sweeps earned their keep. Thanks for the thorough writeup!

Issue #343, opened 2026-09-03
streaming: every Java and C# peek frame seeds its output local from the handle and overwrites it unread (175 of 175)


mario4tier:
Found while shipping the `c_hygiene` phase (dev `46577145f`), which is C-only and cannot see this.

## The defect

Every Java and C# `peek` frame opens by seeding its output local from the handle, then overwrites it before any read:

```java
public double peek( double inReal ) {
   if( !Double.isFinite(inReal) ) throw ...;
   SmaStream sp = this;
   double tempReal = 0.0;
   double cur_outReal = sp.cur_outReal;   // <- dead: never read
   double periodTotal = sp.periodTotal;
   ...
   cur_outReal = tempReal / (double)sp.optInTimePeriod;
   return cur_outReal;
}
```

**175 of 175 seeds are dead at first mention, in both backends** (scanned over all 179 peeks in each; 0 live). Not confined to the stateless functions — SMA above is a stateful frame.

A peek commits nothing, so the previous bar's value is never an input to the transition. The seed is emitted because the `cur_` localization treats an output field like any other state scalar, which is correct for the COMMIT frame and wrong for peek.

## Why no existing gate sees it

- It is a dead **load**, not a dead store, so #328's peek dead-store purge does not reach it.
- The value is correct either way, so no value gate can see it.
- `no_rust_peek_copies_the_handle` and `no_c_peek_copies_the_handle` sweep the other two backends; neither has an equivalent (C writes through an out-param, Rust rebinds).

## Deliverables

1. The peek frame does not seed an output local from the handle; the local is declared at its natural default.
2. A gate that fails if a peek frame reads a `cur_` field it is about to overwrite, with its own counter, sabotage-proved. Java and C# both, since only these two carry the shape.
3. `--xlang-hash` stays bit-identical (179 functions, 0 mismatches).

## What it is worth

Unmeasured. One field load per output per peek call that a JIT may or may not fold; the C and Rust peeks do not pay it. Worth measuring before doing — `scripts/stream_ab.py --lang=java --call=peek` and `--lang=csharp --call=peek` against a branch, per the ta-bench skill's control-function rule. If it folds on both runtimes, close this rather than do it: the source-level argument alone is thin.

kevinlincg:
Ran the measurement this issue gates on, with a working implementation to measure. Verdict from here: **the fold hypothesis holds — this is a close, per your own bar.**

[...]

PR #351, opened 2026-09-04
refactor(streaming): managed peek frames stop seeding output locals they overwrite (#343)

kevinlincg:
Implements the ruling in #343: the managed peek frames stop seeding output locals they provably overwrite.

## What changes

`streaming::peek_seed_is_dead` is a conservative write-before-read query over the already-built peek-transition IR (83 lines, read-only — it rewrites nothing). At the declaration site, `java_stream.rs` / `csharp_stream.rs` replace `double cur_X = sp.cur_X;` with the type default when every mentioning path provably writes the local first. 176 of 177 seeds go; IMI keeps its one, because its sole store sits inside the period loop and the IR cannot prove a loop body runs.

The safety is one-sided: any unrecognised construct falls to `ReadOrUnknown` → keep the seed → today's behaviour. A new IR construct cannot break this pass; it can only make it miss an opportunity.

The generated diff is exclusively seed lines (`double cur_X = 0.0;` / `int cur_X = 0;`) plus the build-stamp digests. Nothing else moved. The managed backends now tell the same story Rust already does — a peek commits nothing, so the previous bar's output is never an input to it.

## Why this is `refactor(`, not `perf(`

There is no measurable perf, and the issue's own "close if it folds" bar was written on a perf premise. **That bar is superseded by the explicit ruling in #343, not met** — this lands as hygiene under the same standard as `46577145f` (the `c_hygiene` phase, which shipped on zero instruction-level change).

The measurement that killed the perf claim, for the record: `stream_ab --lang=java --call=peek`, 6 interleaved rounds × 500k iters, 179 rows — median +0.06%, sd 0.535. The control that settles it: MA and MAVP, dispatch functions that never carried a seed, report the same ≈−1.2% drift as touched rows, and the byte-identical `--call=update` frames spread *wider* (sd 0.645) than the touched peek frames. The effect is inside HotSpot's layout-noise floor. C# is unmeasured — neither bench tool has a C# lane.

## Gate, and its blind direction

`no_managed_peek_seeds_a_dead_output_local` sweeps both managed backends with non-vacuity floors (>150 swept, >150 seed-free) and pins the kept set to exactly `{imi}`. Sabotage-proved in both directions: forcing the analysis false reddens the seed-free floor; forcing it true reddens the exact set.

What the gate cannot do: `seeded == {imi}` pins the pass's *output*. It catches a seed growing back and it catches IMI being wrongly dropped, but it cannot catch a *different* function being wrongly dropped — that direction rests entirely on the runtime peek-vs-update legs (`stream_verify`, `StreamApiTest`). The exact-set assertion is a regression pin, not a correctness proof.

## Verification

- `generate` clean at head (181 functions — RMA and ZLEMA landed since the prototype; both frames are swept and seed-free, and the `{imi}` pin still holds without edits).
- Full generator suite green, gate included.
- `servers` + cross-language stream verify: C / Rust / C# zero mismatches; Java shows only the known Temurin-on-macOS libm tolerance-lane baseline (HT_TRENDMODE), unchanged from dev on the same machine.
- `StreamApiTest` all pass.

### DIFF
--- a/ta_codegen/generator/src/backends/csharp_stream.rs
+++ b/ta_codegen/generator/src/backends/csharp_stream.rs
@@ -1634,6 +1634,17 @@ fn peek_frame_arm_named(
         let (cty, default) = field_type_and_default(ty);
         let _ = writeln!(out, "{pad}{cty} {name} = {default};");
     }
+    // An output local whose every mentioning path writes it before any read
+    // does not need the handle's value: a peek commits nothing, so the previous
+    // bar's output is never an input to the transition (issue #343). Seeding it
+    // anyway is a dead field load on every call. Anything the analysis cannot
+    // prove keeps the seed.
+    let dead_seeds: BTreeSet<String> = func
+        .outputs
+        .iter()
+        .map(|o| format!("cur_{}", o.name))
+        .filter(|n| locals.contains(n) && streaming::peek_seed_is_dead(&body_ir, n))
+        .collect();
     for name in &locals {
         if predeclared.contains(name) {
             continue;
@@ -1645,6 +1656,9 @@ fn peek_frame_arm_named(
         if let Some(elem) = cty.strip_suffix("[]") {
             let _ = writeln!(out, "{pad}{cty} {name} = new {elem}[sp.{name}.Length];");
             let _ = writeln!(out, "{pad}Array.Copy( sp.{name}, {name}, sp.{name}.Length );");
+        } else if dead_seeds.contains(name) {
+            let zero = if cty == "int" { "0" } else { "0.0" };
+            let _ = writeln!(out, "{pad}{cty} {name} = {zero};");
         } else {
             let _ = writeln!(out, "{pad}{cty} {name} = sp.{name};");
         }
--- a/ta_codegen/generator/src/backends/java_stream.rs
+++ b/ta_codegen/generator/src/backends/java_stream.rs
@@ -1507,6 +1507,17 @@ fn peek_frame_arm_named(
         let (jty, default) = field_type_and_default(ty);
         let _ = writeln!(out, "{pad}{jty} {name} = {default};");
     }
+    // An output local whose every mentioning path writes it before any read
+    // does not need the handle's value: a peek commits nothing, so the previous
+    // bar's output is never an input to the transition (issue #343). Seeding it
+    // anyway is a dead field load on every call. Anything the analysis cannot
+    // prove keeps the seed.
+    let dead_seeds: BTreeSet<String> = func
+        .outputs
+        .iter()
+        .map(|o| format!("cur_{}", o.name))
+        .filter(|n| locals.contains(n) && streaming::peek_seed_is_dead(&body_ir, n))
+        .collect();
     for name in &locals {
         if predeclared.contains(name) {
             continue;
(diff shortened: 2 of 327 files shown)


kevinlincg:
Landed on dev as 88ba4c76 (generator patch is identical by patch-id), with the follow-ups a2880f15 / 4a4897ca / 4c3a63db on top — including the correction that 10 Rust peeks do not rebind at the type default, which my PR body overstated. Closing.

mario4tier:
Landed on `dev` as `88ba4c762`, with your authorship intact. GitHub shows this closed
rather than merged only because the branch was rebased onto the current `dev` on the
way in, so the SHA differs — the change itself is in.

`dev` moved three times during review (RMA → ZLEMA → two candlestick-comment fixes), so
the rebase-and-regenerate step ran again on landing. The generated diff against `dev` is
still exclusively seed lines and the two build-stamp digests — 436 `double cur_X = 0.0;`
/ 268 `int cur_X = 0;`, paired one-for-one with their removals, nothing else moved.

Three commits went on top:

- `a2880f15a` **`test(streaming)`** — 12 unit tests for `peek_seed_is_dead`. The shipped
  corpus exercises only 5 of its 9 decision points; a compound assignment, a
  self-reading write, an if-condition that reads the local, an else-arm that writes and
  returns, and a switch arm were reached by nothing, so the fallback that *keeps* the
  seed was untested in exactly the direction that would emit a wrong value. Sabotage
  partitions them 8 keep / 4 drop, and two of them catch mutations that move no
  generated line, so no existing gate could have seen those.
- `4a4897cad` **`refactor(streaming)`** — the pass now reuses `expr_var_names` /
  `stmt_var_names` instead of its own `reads` / `mentions_stmts`. Strictly coarser, so
  it can only keep a seed; it also closes the one gap the hand-rolled pair had
  (`Expr::PointerDeref` holds a name rather than a boxed sub-expression, so a `*cur_x`
  read was invisible to it). No generated line moves.
- `4c3a63db9` **`docs(streaming)`** — the gate docstring said "Rust rebinds its locals,
  so only these two ever carried the shape." Ten Rust peeks refute it, and there the
  local is wholly dead. Filed as #353. Also trimmed the emit-site comment, which
  restated the callee's docstring in both backends.

Two notes on things you flagged, both of which held up under checking:

- The one-sided-safety argument is right, and it now has an independent proof: stripping
  the initializer from all 175 declarations and recompiling, javac and Roslyn's own
  definite-assignment analyses accept every one, while IMI's kept seed immediately
  yields `CS0165`.
- The tempting simplification is wrong, and worth recording so nobody re-proposes it:
  `peek_seed_is_dead` cannot be replaced by `step_local_scratch` over the existing
  `DefAssign` walker. No peek `body_ir` ever ends in a `Return` — 198 end in `Assign`,
  182 in `If`, 2 in `ForC` — because the trailing `return cur_outReal;` is emitted by
  `emit_handle_class`, outside the IR the analysis walks. `DefAssign` therefore sees no
  read at all for IMI and drops its seed. Your "any mention inside a loop is a read"
  conservatism is load-bearing, not caution.

Thanks — the control-function measurement that killed the perf claim was the valuable
half of this, as said on #343.

mario4tier:
@kevinlincg It happen a few times that you are closing PR eagerly while I am in the process to commit and push it to dev, and therefore these show up "red close as unmerged" instead of green "merged". It works one way or the other, but it make things a bit misleading and not crediting your PR work as it should.

No big deal, just wanted to share the observation.

PR #356, opened 2026-09-04
fix(streaming): the unit-test helper borrows its body instead of consuming it

kevinlincg:
dev `4c3a63db` fails the PR clippy gate on itself: `a2880f15`'s test helper `dead(body: Vec<Statement>)` trips 1.97's pedantic `needless_pass_by_value` (denied in the generator), and since it landed by direct push it never crossed the gate — every open PR's merge preview is red on it (first seen on #354's checks; reproduced locally on unmodified dev with `build.py clippy`).

One-line-shape fix: the helper takes `&[Statement]`, its twelve call sites pass `&[...]`. `cargo test --lib` 432 passed, `build.py clippy` clean (generator + generated crate). No generated output is touched.

### DIFF
--- a/ta_codegen/generator/src/streaming.rs
+++ b/ta_codegen/generator/src/streaming.rs
@@ -10154,8 +10154,8 @@ mod tests {
         "cur_outReal"
     }
 
-    fn dead(body: Vec<Statement>) -> bool {
-        peek_seed_is_dead(&body, out_local())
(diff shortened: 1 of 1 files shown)


kevinlincg:
Superseded by ee28a826, which keeps the `dead(vec![..])` style and suppresses at the helper — the same treatment helper_registry.rs gives implicit_hasher, and a better trade than restyling thirteen cases. Closing.

Issue #362, opened 2026-07-20
(COPPOCK) Coppock Curve


mario4tier:
## What it is

Long-term momentum oscillator: a 10-period weighted moving average of the sum of a 14-period and an
11-period rate of change. Edwin S. "Sedge" Coppock, *Barron's*, originally published as the "Trendex
Model" (Wikipedia dates it Oct 15 1962; StockCharts says Oct 1965 — see References).

## Formula

Canonical (author's) form — the ROCs are **summed**, not averaged:

```
w = optInWMAPeriod (10), p1 = optInROC1Period (11), p2 = optInROC2Period (14)

R(j,p) = (inReal[j-p] != 0.0) ? ((inReal[j] / inReal[j-p]) - 1.0) * 100.0 : 0.0   /* == TA_ROC, incl. its zero guard */
S(j)   = R(j,p1) + R(j,p2)
out(j) = TA_WMA(S, w)(j)
```

Fused loop must reproduce `TA_WMA`'s recurrence verbatim, otherwise the composite differential is not
bit-exact (`ta_codegen/input/wma/wma.c:73-140`):

```
divider       = (w*(w+1)) >> 1            /* INTEGER divider, not pandas' 0.5*w*(w+1) */
lookbackTotal = max(p1,p2) + w - 1
if (startIdx < lookbackTotal) startIdx = lookbackTotal
trailingIdx = startIdx - (w-1);  periodSum = periodSub = 0.0
for (inIdx = trailingIdx, i = 1; inIdx < startIdx; inIdx++, i++) { t = S(inIdx); periodSub += t; periodSum += t*i; }
trailingValue = 0.0
for (inIdx = startIdx; inIdx <= endIdx; inIdx++) {
    t = S(inIdx);
    periodSub += t; periodSub -= trailingValue; periodSum += t*w;
    trailingValue = S(trailingIdx++);       /* read BEFORE the store => outReal==inReal is safe */
    outReal[outIdx++] = periodSum / divider;
    periodSum -= periodSub;
}
```

No seeding convention exists (no EMA/IIR anywhere); the only init is the `w-1` finite-window warm-up.
`w == 1` needs no special case — the generic recurrence yields `out = t` exactly, bit-identical to
`TA_WMA`'s `optInTimePeriod==1` memmove fast path.

**Competing variant (reject).** Tulip's `beta/copp.c:75` computes
[...]

PR #375, opened 2026-09-04
feat(coppock): Coppock Curve, WMA of the summed ROC pair (#362)

kevinlincg:
Implements #362: COPPOCK, the Coppock Curve — `WMA(w)` of `ROC(p1) + ROC(p2)`, defaults 10/11/14, lookback `max(p1,p2) + w − 1` (23 at the defaults). The **sum**, per every published definition; Tulip's half-scale mean stays rejected. `p1 > p2` is accepted — the formula is symmetric and the lookback keys off the max (issue decisions 1–3 followed; param names match the existing pandas oracle arm, so no oracle edit).

## Streaming: how the body earned `flags: [stream]`

The issue marked the T3 tier "plausible and unproven". It took three shapes to make it true, and the middle one is the cautionary tale:

1. Parameter-sized lag subscripts (`inReal[inIdx - optInROC1Period]`) are outside the classifier's index grammar — stage 1 rejects, loudly.
2. Per-lag trailing cursors (TA_ROC's own shape) **passed stage 1 but generated a broken streaming engine**: the rescan walk's cursors (`wj1`/`wj2`), recomputed each re-anchor, were modelled as persistent rings — `stream_verify` failed with batch/stream value divergence at vector 46 and `state_bad: ringCap_wj1`. Stage 1 acceptance is not streaming correctness; the runtime legs are the proof.
3. The shipped shape: S is a derived series carried in a `(w−1)`-slot ring, exactly TA_HMA's outer-stage pattern — trailing reads the expiring slot, the re-anchor walks the ring oldest-first, then adds the current bar at weight w. Each ROC denominator keeps its own plain trailing cursor. `stream_verify` green across all four backends, all vectors.

## One deviation from the issue text, found by the differential

"`w == 1` needs no special case — the generic recurrence yields `out = t` exactly" is not true in floating point: `periodSub`'s `(prev + t) − prev` round-off seeds `periodSum` with a residue from the third output on. Measured on TA_SREF at `w=1`: bar 24 differs from `TA_WMA(1)`'s copy fast path in the last bit. The identity case answers directly, matching TA_WMA's own `w==1` behaviour — the bit-exact differential is what surfaced it.

## Tests (`test_composite2.c`, tag now `SMI,COPPOCK`)

- **Composite differential, bit-exact** (`memcmp`): fused vs shipped `TA_ROC + TA_ROC + TA_WMA`, over a 7-cell param grid (defaults, `w=1`, `p1>p2`, `p1==p2`, all-minimal, lag 200, `w=20`) × 6 startIdx values — the startIdx axis is what pins the re-anchor phase on both sides. Non-vacuity floor: ≥5000 compared values.
- **External pins**: pandas 2.3.3, textbook composition (explicit ROC sum → rolling weighted dot 1..10/55), 5 spot bars with |value| ≳ 1, rel 1e-11. `outBegIdx/outNBElement = 23/229` asserted. Worst measured rel ~2e-12; the gap is shipped `TA_WMA`'s running-total residue (attribution commented in-file so nobody "fixes" it by loosening).
- **Flat input** ⇒ exactly 0.0 everywhere; **mid-series zero price** ⇒ all outputs finite through `TA_ROC`'s guard, with a non-vacuity check that something is nonzero.
- **`outReal == inReal` aliasing**, bit-compared against the separate-buffer call.

## Verification

- `generate` / `servers` / full generator suite / `build.py clippy` — clean on dev `efa925dd`.
- Full `--codegen` (all four languages, all functions) rc=0, including COPPOCK's `stream_verify`.
- `--xlang-hash`: zero non-baseline mismatches (COPPOCK has no transcendental — no `CODEGEN_TRANSCENDENTAL[]` entry, bitwise in all lanes; Java shows only the known HT_TRENDMODE macOS-libm baseline).
- `regen-check` and `check-source-lists` clean; CHANGELOG entry added.
- The issue's "stale count literal at `test_codegen.c:4555` (prints 162)" no longer exists at head — it left with the RMA/ZLEMA landings.


mario4tier:
Two things on head `f7958563` before this lands.

### 1. `wj1` / `wj2` are read uninitialized in the generated C `_OpenImpl` — C only

`src/ta_func/ta_COPPOCK.c:728-729` declares them with no initializer:

```c
int wj1;
int wj2;
```

The only assignment (`wj1 = trailRoc1Idx;`) is inside the `if( barsSinceReseed <= 0 )` re-anchor branch, but the reads are unconditional:

```c
sp->ringCap_wj1 = (int)(inIdx - wj1);   /* :952 */
sp->ringCap_wj2 = (int)(inIdx - wj2);   /* :960 */
```

The re-anchor only fires after `8 * optInWMAPeriod` tight-loop iterations, so any `TA_COPPOCK_Open` whose history is shorter than `lookbackTotal + 8*optInWMAPeriod` — **104 bars at the defaults** — reads an indeterminate value.

The other three backends all initialize it, so this is a C-vs-ports divergence rather than a style nit:

- `ta_codegen/output/rust/library/src/ta_func/coppock.rs:660` — `let mut wj1: usize = 0_usize;` (in `coppock_open_impl`)
- `ta_codegen/output/java/fragments/Core_COPPOCK.java:1018` — `int wj1 = 0;` (in `coppockOpenImpl`)
- `ta_codegen/output/csharp/library/src/Core_COPPOCK.cs:1094` — `int wj1 = 0;` (in `CoppockOpenImpl`)

On a short-history open the three ports deterministically compute `cap = historyLen`; C computes garbage.

Memory safety is not the issue — the `cap < 0 || cap > historyLen` guard, `allocN = max(cap,1)` and the mod-cap wrap keep every access in bounds for anything that passes the guard. The consequence is a spurious `TA_INTERNAL_ERROR(412)/(413)` or silently wrong stream state.

The root cause looks generator-side rather than input-side: a *branch-local* cursor promoted to a stream ring is a construct no shipped indicator had before (WMA's only ring cursor, `trailingIdx`, is unconditionally assigned), so the C local-init policy has not had to cover it until now.

### 2. The PR body describes the previous head's streaming shape

The "how the body earned `flags: [stream]`" section describes shape #3 — S carried in a `(w-1)`-slot ring, the `CIRCBUF_PROLOG(sRing,double,50)` form. That ring is gone at this head; the code is now the per-lag trailing-cursor shape, which the same body lists as shape #2 and reports as *"passed stage 1 but generated a broken streaming engine ... `stream_verify` failed with batch/stream value divergence at vector 46 and `state_bad: ringCap_wj1`"*. The rings named in that failure are back (`ta_COPPOCK.c:496-501`).

So the body's verification claims — `stream_verify` green on all four backends, full `--codegen` rc=0, `--xlang-hash` zero mismatches — describe code that is no longer in the PR.

Worth flagging that no PR gate can catch either of these: the seven green checks compile the generated sources and confirm regeneration is a fixpoint, and none of them runs `ta_regtest`. Could you re-run `--codegen`, `stream_verify` and `--xlang-hash` against this head, including at least one `Open` with `historyLen` below `lookback + 8*optInWMAPeriod`, and refresh the streaming section to match the shipped shape?

kevinlincg:
Both findings were right, and the root cause of #2 explains #1: head `f7958563` was a regression I introduced while re-basing across the VHF/FOSC landings — I replayed the branch from a stale local backup that predated the ring rewrite, so the shipped-and-verified shape #3 got silently replaced by the shape #2 the PR body itself calls broken. The `wj1`/`wj2` uninitialized read you found lives only in that resurrected shape.

Fixed at the new head `a6df0728`: the ring shape (#3) is restored from the original commit, `src/ta_func/ta_COPPOCK.c` contains zero `wj1`/`wj2` and the `sRing` carry, and everything re-ran **against this head**:

- Full `--codegen` (all four languages, all functions) rc 0, including COPPOCK's `stream_verify`.
- `--xlang-hash`: zero non-baseline mismatches.
- `regen-check`, `check-source-lists`, full generator suite, clippy — clean.
- **Short-history Opens, per your ask**: `stream_verify` driven by hand at `gen_n = 30 / 50 / 100` (all below `lookback + 8*w = 104` at the defaults), against all four servers — `ok:1, retCode:0` everywhere, `nb = 7 / 27 / 77`. In the ring shape there is no conditionally-assigned ring cursor to leave uninitialized: `sRing`'s capacity is the unconditional `w-1` and its cursor advances every bar.

The PR body's streaming section now matches the shipped shape again (it always described #3; the code briefly didn't).

Your generator-side observation stands on its own, ring shape or not: a branch-local cursor promoted to a stream ring is a construct the C local-init policy has never had to cover, the other three backends zero-init it, and nothing in the PR gates runs `ta_regtest`. That divergence is latent for the next input that writes a conditional cursor — worth its own issue if you want it tracked; I kept my repro of the old head (`git show f7958563` on my fork's reflog) if useful.

mario4tier:
Merged as `ff7e52220`. Thanks — and sorry for the rebases; every one of those was `dev` moving under you while I landed a batch.

Four changes were layered on top before landing, all in files this PR touched:

**1. `coppock.c` — the `optInWMAPeriod == 1` comment.** It described the arm as a 1-ULP nicety ("bar 24 at w=1 differs in the last bit"). Measured by removing the arm and regenerating: TA_SREF bar 16 at `(1,11,14)` moves from `-11.311839169954585` to `-6.4591709868291103`. `ringSize` clamps to 1, so the read-before-write ring hands back the wrong trailing value — the arm is load-bearing. The comment now says so, because it sits in the hot loop and reads as an invitation to delete it.

**2. `test_composite2.c` — the zero-guard leg could not see a wrong neutral.** It asserted finiteness and "not every output is 0.0", both of which any finite neutral satisfies. Changing both `: 0.0` neutrals to `100.0` passed every leg, `--xlang-hash` (all four backends generated from the same mutated source agree) and `--codegen`. It now memcmps against `TA_ROC + TA_ROC -> TA_WMA`, which carries TA_ROC's own guard; the same mutation fails at bar 41.

**3. `test_composite2.c` — the leg header claimed the flat-input half was the guard's non-vacuity evidence.** A flat series never divides by zero, so it takes the non-guard arm; its exact zeros say nothing about the guard.

**4. `coppock.md` — `## Implementation` restored to the source-link block.** It carried prose instead, so the generated page would have been the only function page with no link to its own sources and no wrapper pointer. The mechanics moved to `## Notes`.

Also: your error ids collided with CVI/MASSI, which took 416-419 while this PR was open. Since yours had not shipped, they were dropped and `generate` reallocated them to 420-423. Nothing already released moved.

Issue #372, opened 2026-09-04
(CUMSUM) Cumulative Sum


mario4tier:
> [!IMPORTANT]
> **RULED: ship it.** `TA_ADD`/`TA_SUB`/`TA_SUM`/`TA_MAX` are already non-indicator primitives. Oracle of record is **ta4j** `RunningTotalIndicator`; numpy is the second arm.
>
> The `Open questions` section further down is **historical** — it records what was open
> when the card was written, not what is. The ruling comment on this issue is the current
> state and wins over the body wherever they disagree.

## Proposal

**Ship `TA_CUMSUM`** — one `inReal` in, one `outReal` out, no optional inputs,
`TA_CUMSUM_Lookback() == 0`, flags `[stream, path_dependent]`, group `Math Operators`.

Spun out of board card TA-Lib/ta-lib-proposal-drafts#4 ((MBI) Market Breadth Indicator), which declined MBI as titled and
named the cumulative sum as the one primitive its published recipes were missing.

Three things make it worth a function rather than a caller-side loop:

1. **It is the only primitive between the shipped corpus and three named classical
   indicators.** A/D Line `= CUMSUM(SUB(A, D))`; A/D Volume Line `= CUMSUM(SUB(AV, DV))`;
   McClellan Summation Index `= CUMSUM(McClellan Oscillator)`. `TA_SUM` is a *rolling window*
   (`ta_codegen/input/sum/sum.yaml:13`, `optInTimePeriod` range `[2, 100000]`), and MEASURED,
   no shipped function takes a `real` series and returns its running total — the accumulators
   that exist all run over a term derived from a `price` bundle.
2. **An outside library exposes it as a named indicator**, so the publishable rule is met, not
   argued around: ta4j `RunningTotalIndicator`, class javadoc *"Running Total aka Cumulative
   Sum indicator"* — and MEASURED bit-exact against a sequential `+=` on this box (below).
3. **The emitter has never produced this combination.** MEASURED over
   `ta_codegen/input/*/*.yaml`: all 10 `path_dependent` functions carry `stream`, and every one
   of them takes a `price` bundle (ad adosc nvi pvi sar sarext supertrend vwap wad) or
   `(real, price)` (obv). `TA_CUMSUM` would be the first single-`real` × `path_dependent` ×
   `stream` function. Expect the surprise there, not in the arithmetic.

The arithmetic is one line. The content of this card is (3) and the `startIdx` contract.

## What it is

[...]

PR #376, opened 2026-09-04
feat(cumsum): Cumulative Sum, the running total from the anchor bar (#372)

kevinlincg:
Implements #372 as ruled: `TA_CUMSUM` — one `inReal` in, one `outReal` out, no optional inputs, `TA_CUMSUM_Lookback() == 0`, group `Math Operators`, flags `[stream, path_dependent]`.

## The contract

The accumulator re-seeds at the anchor: `TA_CUMSUM(3, 7, x)[0]` is `x[3]`, **not** `sum(x[0..3])` — the convention `ad.c`/`obv.c` already ship, and the published one (StockCharts: only the A/D Line's shape carries meaning). Left-to-right plain `+=`, uncompensated, per the issue's rejected-variants table. Two statements in the loop, never a value-producing compound assignment — it cannot lower to Rust.

## The pairing that had never generated

The issue predicted "expect the surprise there, not in the arithmetic": this is the first single-`real` × `path_dependent` × `stream` function (obv proves the accumulator streams, cosh proves the signature generates, nothing proved the pair). Measured outcome: **no surprise** — `generate` clean, `stream_verify` green across all four backends including the `OpenAndFill` per-bar compare, full `--codegen` rc 0, `--xlang-hash` bitwise with zero non-baseline mismatches, `regen-check` and `check-source-lists` clean.

## Tests (`test_cumsum.c`, registered in CMakeLists + Makefile.am + `DO_TEST`)

Item numbers from the issue's verification plan:

- **(1-adjacent) C40 golden, bit-exact**: the corpus published in full on the board card, `net` built through shipped `TA_SUB`, `memcmp`'d against an in-test sequential `+=`, with the published spot values (`700 800 -300 1000 1500 1000 -700 1200 … 4400`) pinning the reference loop itself. The card MEASURED ta4j 0.22.6 `RunningTotalIndicator` bit-identical on exactly these values.
- **(2) The `startIdx` contract**: the published `(3,7)` slice (`1300 1800 1300 -400 1500`) and single-bar anchors `TA_CUMSUM(k,k,x)[0] == x[k]` — the one place a well-meaning implementer silently diverges.
- **(3) `path_dependent` earned**, in `test_supertrend.c` Leg 6's shape: flag published via `TA_GetFuncInfo`; a sub-range call genuinely differs from full history (non-vacuous); and the sub-range call is bit-identical to a full-history call over the truncated input — a re-anchor, not stale state.
- **(6) Edges**: `(0,0)` and `(N−1,N−1)`; `outReal == inReal` aliasing asserted rather than assumed; an alternating `+x/−x` series landing on exactly `0.0`; an absorption series (`1e18` then `1.0`s) asserting the plain uncompensated answer.

## Left to the oracle infrastructure

The ta4j (`Server.java` arm — note the issue's `buildSeries` arity wrinkle) and pandas (`compute_CUMSUM`) arms live outside this repository. The goldens above are the issue's own MEASURED ta4j-bit-identical values, so the formula leg is covered here; the live arms remain the infrastructure's to add.

CHANGELOG entry added.


mario4tier:
Merged as `ff7e52220`. Thanks — and apologies for the rebases; those were `dev` moving under you.

Two changes were layered on before landing:

**1. `cumsum.md` — the ta4j claim.** The Formula section said the plain `+=` is "what ta4j's `RunningTotalIndicator` computes bit-exactly". Disassembled from the pinned jar (`ta4j-core-0.22.6`), that class carries a `private final int barCount` and its constructor is `(Indicator<Num>, int)` — it is a **windowed trailing total**, not a cumulative sum. The clause is removed; the `AD`/`OBV` convention statement stands on its own.

**2. `cumsum.md` — `## Implementation` restored to the source-link block**, with the lookback and no-compensation notes moved to `## Notes`. As written, the generated page would have shipped with no link to its own definition or native sources.

Issue #361, opened 2026-07-20
(ERI) Elder Ray Index


mario4tier:
## What it is

Elder Ray Index — Alexander Elder, *Trading for a Living* (1993). Two lines measuring how far
the bar's extremes sit from a 13-period EMA of close: Bull Power = High − EMA, Bear Power = Low − EMA.

## Formula

`n = optInTimePeriod` (default 13), `k = 2/(n+1)`. Both lines share ONE EMA of close.

```
lookbackTotal = ema_lookback(n) = n - 1 + TA_GetUnstablePeriod(TA_FUNC_UNST_EMA)
if (startIdx < lookbackTotal) startIdx = lookbackTotal;
if (startIdx > endIdx) { *outBegIdx = 0; *outNBElement = 0; return TA_SUCCESS; }

/* Seed - reproduces ta_codegen/input/ema/ema.c's DEFAULT arm op-for-op. */
today = startIdx - lookbackTotal;
i = n; tempReal = 0.0;
while (i-- > 0) tempReal += inClose[today++];   /* sequential sum from 0.0, THEN divide */
prevMA = tempReal / n;

/* Warm-up also consumes the EMA unstable period. */
while (today <= startIdx) prevMA = ((inClose[today++] - prevMA) * k) + prevMA;

outBullPower[0] = inHigh[startIdx] - prevMA;
outBearPower[0] = inLow[startIdx]  - prevMA;
outIdx = 1;

for (today = startIdx + 1; today <= endIdx; today++) {
    prevMA = ((inClose[today] - prevMA) * k) + prevMA;
    outBullPower[outIdx]   = inHigh[today] - prevMA;
    outBearPower[outIdx++] = inLow[today]  - prevMA;
}
*outBegIdx = startIdx; *outNBElement = outIdx;
```

No division in the per-bar map ⇒ no 0/0, no NaN path (issue #112 satisfied by construction).
Invariant `outBullPower[i] >= outBearPower[i]` on every bar, since high ≥ low.

No reference disagreement was found on the formula itself. TradingView additionally plots a third
line "Bull Bear Power" = Bull + Bear; that is not part of Elder-Ray proper (pandas-ta-classic,
StockCharts and TC2000 all ship two lines). **Ship 2 outputs.**

## Proposed API

```c
TA_RetCode TA_ERI( int    startIdx,
                   int    endIdx,
                   const double inHigh[],
                   const double inLow[],
                   const double inClose[],
[...]

PR #379, opened 2026-09-05
feat(eri): Elder Ray Index, Bull and Bear Power against one EMA (#361)

kevinlincg:
Implements #361: `TA_ERI` — Bull Power = `High − EMA(Close, n)`, Bear Power = `Low − EMA(Close, n)`, both lines against **one** shared EMA (Elder, *Trading for a Living*, 1993). Default 13, two `line` outputs, lookback = `ema_lookback(n)` with the EMA unstable period consumed in the warm-up; `UNSTABLE_MAP` gains `{"ERI", TA_FUNC_UNST_EMA}` per the issue's item 3. One fused loop per the issue's Notes — a composed `ema()` + combine form cannot stream (the ACCBANDS rationale), and no compatibility branch (EFI/SMI precedent). No division anywhere: no NaN path (#112 by construction).

## Two generator inventories, each caught by its own gate

ERI is the first function since those gates landed to be simultaneously **fma-fusing** (the inlined EMA recurrence lowers through explicit `fma()` via the #183 pass) and **transitively unstable** (through the hard-coded inner EMA). Both static inventories gained their row after their gate flagged the drift — `FUSING_INVENTORY` in `backends/fma.rs` and `INHERITED` in `stability_suite.rs` — which is exactly the drift those gates exist to catch (the `efi` incident in FUSING_INVENTORY's own docstring).

## Tests (`test_eri.c`, registered in CMakeLists + Makefile.am + `DO_TEST`; note `--function=ERI` substring-matches HT_DCP**ERI**OD, so the tag is exact)

- **Differential, bit-exact and non-tautological**: shipped `TA_EMA` + two subtractions, `memcmp`'d over 5 periods × 6 startIdx × 4 EMA-unstable-periods (≥10000-value floor), with the `bull ≥ bear` invariant free on every cell. The fused loop reproduces ema.c's DEFAULT arm op for op — seed-sum order and warm-up both covered by the unstable axis.
- **Golden pins at the defaults, ABSOLUTE tolerance 1e-12** per the issue's decision 1: ERI is a cancelling difference and relative error is unbounded at the zero crossings that carry the signal (the issue measured 1.9e-10 rel at an out-value of −1.5e-4). Rows pinned away from zero crossings (|value| ≥ 0.5). Provenance: from-scratch transcription of ema.c's DEFAULT arm on the committed TA_SREF corpus — the NVI/PVI pattern; the issue measured pandas-ta-classic 0.6.52 against the same transcription at max abs 7.1e-14, so the rows hold against either.
- **Edges**: all-flat ⇒ both lines exactly 0.0; `high == low` ⇒ the two lines bitwise identical; `n=1` ⇒ `bull == high − close` bit for bit (the EMA identity survives the fma lowering with no special case — verified, not assumed).
- **In-place aliasing**, each of the 2 outputs onto each of the 3 inputs — the widest aliasing surface of any recent addition; the bar's extremes are loaded into temps before either output is written.

## Verification

Full `--codegen` (all four languages) rc 0, `stream_verify` green (tier T2 scalar recurrence), `--xlang-hash` zero non-baseline mismatches, full generator suite + clippy + `regen-check` + `check-source-lists` clean. The pandas oracle arm already exists in the oracle infrastructure outside this repo; the trading-signals arm is left to it. CHANGELOG entry added.


mario4tier:
Reviewed. One blocker, one vacuous gate, and two doc items.

### Blocker — the inlined EMA drops `ema.c`'s period-1 identity arm

`ema.c:61` carries an explicit `if( optInTimePeriod == 1 )` short-circuit that returns a bit-exact copy of the input. `eri.c`'s transcription has none, and `eri.yaml` declares `range: [1, 100000]`, so period 1 is legal and reachable.

At `n == 1`, `k` is exactly 1.0 and the recursion degenerates to `fma(x - prev, 1.0, prev)` = `fl(fl(x - prev) + prev)`, which equals `x` only while consecutive closes stay within a factor of two (Sterbenz). So `TA_ERI(…, 1)` is **not** `high - TA_EMA(close, 1)`.

That breaks three things at once: the bit-exact contract issue #361 decision 2 ruled on, the assertion in your own comment at `eri.c:44-47` ("the differential against shipped TA_EMA holds bitwise"), and the premise of the tests below.

**The n=1 legs cannot catch it** — `test_eri.c:246` memcmps `outBull` against `high - close`, and `n=1` is in `eriPeriodGrid` at `:89` — because the TA_SREF corpus is Sterbenz-benign, so the recursion happens to return `x` on every bar of it. A series with a consecutive close ratio outside [0.5, 2] would turn both red.

### The aliasing sweep detects nothing it exists for

`test_eri.c:285` runs section 4 at `optInTimePeriod = 13` only. There `outBegIdx = 12`, so every store sits 12 slots behind the load and none of the six aliasing combinations can bite. Running it at the minimum period is what would exercise it.

### Documentation

- **`eri.md:5`** — the Summary hard-codes the default as though the period were fixed: "distance … from a 13-period EMA of the close". `optInTimePeriod` is a parameter with range 1..100000, so the definition is false for every other value. This sentence is the only definition carried into rustdoc, javadoc and the C# XML docs, so it ships to three backends.
- **`eri.md:9`** — "TradingView additionally plots the sum of the two as a third line" is not right. TradingView's built-in *Bull Bear Power* — which its own support page calls "otherwise known as the Elder-Ray Index" — plots **only** the sum, not three lines.

Issue #381, opened 2026-09-05
PR-gate IR check: a division whose accumulated divisor is guarded by something other than itself


mario4tier:
## The gap

Nothing in the tree checks that a **domain guard** is the right *kind* of guard, or
that any test can see it change. Two separate failures this week, both shipped past
every gate:

**1. Guards that are correct in exact arithmetic and wrong in floating point.** Three
instances, all in new functions, all caught only by human/agent review:

- `VORTEX` divided by a running true-range sum gated only on a flat-bar counter. FP
  absorption drives the sum to exactly 0.0 while the counter is short, so a legal
  40-bar OHLC series returned `NaN` and `+Inf` from a function that does not declare
  `TA_FUNC_FLG_NAN_INF_OUT`.
- `VORTEX` again, after the first fix: an over-broad reseed zeroed the two numerator
  sums on a predicate that only proves the denominator's terms are zero, producing a
  **negative `-VI`** — unreachable for a ratio of sums of `fabs()`.
- `ER` used its asymmetric KAMA-parity clamp (`sumROC1 <= periodROC`) as the zero
  guard. It cannot fire when `periodROC < 0`, so a zero denominator divided.

**2. Tests that cannot fail.** At least five this week, each built on the one input
shape where the thing it guards is a no-op — e.g. `ER`'s dead-flat leg used a series
flat from bar 0, where the priming sum is already 0.0 and the purge it exists to
protect never changes anything. Deleting the guard passed the whole suite
bit-identically.

## Why no existing gate sees either

- No PR gate runs `ta_regtest` at all — CI compiles and fixpoint-checks regeneration.
- All four backends are generated from one source, so `--xlang-hash` agrees *bitwise
  on the same wrong answer*.
- `--codegen` value-compares nothing for a post-cutover function (no frozen baseline).
- The reference corpus is benign in exactly the ways that matter: `TA_SREF` has no bar
  with `TR == 0`, no 14-bar halt, and no consecutive close ratio outside [0.5, 2].

## Proposal

A nightly **guard-mutation** leg, in the shape of `scripts/synth_gate.py`: for each
`ta_codegen/input/*/*.c`, enumerate the domain guards (`TA_IS_ZERO`,
`TA_IS_ZERO_SCALED`, `> 0.0`, `!= 0.0`, `<= 0.0`), apply one substitution at a time —
notably **`TA_IS_ZERO_SCALED` <-> `TA_IS_ZERO`**, which is the #253 axis — regenerate
[...]

Issue #56, opened 2025-01-20
where is source of mcpp.exe and how to use gen_code.exe to generate a new indicator?


cloudslee:
where is source of mcpp.exe and how to use gen_code.exe to generate a new indicator?
because i want to add a new indicator, magic nine turn

mario4tier:
I recommend an ubuntu or macOS setup for dev with TA-Lib... WSL on windows would also do. 

It should work on windows native, but I do not actively use it for development anymore (I use a windows VM only for packaging and running ta_regtest).

mcpp.exe
========
From https://mcpp.sourceforge.net/download.html

For convenience, I did embed the .exe in the repos https://github.com/TA-Lib/ta-lib/tree/main/src/tools/gen_code

how to use gen_code
===================
Just a heads-up, this is not trivial, but if you are willing to spend a few days to get started, I am here to help (when asking a question, consider putting your latest code in a fork so I can review all changes instead of going back and forth with small code excerpt).

gen_code generates the ta_XXXX.c files (and more). The function signature (the API) is defined in C in one of the "table file": https://github.com/TA-Lib/ta-lib/tree/main/src/ta_abstract/tables

Best way to start is to cut&paste&rename an existing "similar" function in a table, rebuild and run gen_code... and verify that a new ta_func/TA_xxxx.c file is created.

You then modify the non-generated portion of TA_xxxx.c file to put your own code. You can always add/remove parameters by changing the table and rebuilding+running gen_code.

To test the logic, rebuild ta_regtest after a gen_code. Again, cut&paste an existing "similar" function for testing.

Submit code
===========
Create a PR on the dev branch (not main).

PR #39, opened 2024-10-11
add macOS libtool support.

halohsu:
the libtool named glibtoolize on macOS.

PR #383, opened 2026-09-05
test(codegen): flag loop-accumulated divisors with no zero guard (#381)

kevinlincg:
Implements the narrower proposal from #381 — an IR check in the generator's own suite,
on the PR gate that already runs, in milliseconds. No regenerate, no rebuild, no
mutation.

## What it checks

For every division whose denominator is a variable accumulated across steady-loop
iterations by add/subtract, assert that a guard comparing that same variable **against
zero** dominates the division.

**Stricter than the proposal's wording, for a measured reason.** "Some guard on that
same variable" would have passed the ER defect: `sumROC1 <= periodROC` names the
divisor and still cannot fire when `periodROC` is negative. Only a comparison against
zero establishes non-zero-ness, so that is what the check requires.

## The sweep is mutation-validated

A structural sweep that reports nothing is indistinguishable from one whose matcher
silently stopped matching, so a second test strips ER's exact denominator test back to
the bare KAMA-parity clamp — the pre-#350 defect — and asserts the sweep flags it. That
test fails under the looser rule above, which is how the rule was chosen rather than
guessed.

It caught two of my own matcher bugs while I was writing it, both of which had produced
clean output: matching only top-level loop statements found none of ER's accumulators
(its loops nest under the priming block), and scanning `body` while ignoring
`private_body` skipped ER's divisions entirely.

## Findings

Nine on the first honest run, five of which were mine to fix rather than the tree's:

- **`fabs` aliases.** HT_DCPHASE guards correctly as `tempReal = fabs(imagPart); if(
  tempReal > 0.0 ) ... realPart / imagPart`. The sweep now follows `t = fabs(v)`, or it
  reports a false positive on the exact idiom it should be encouraging.
- **Divisions outside every loop.** `ht_trendline.c:113` is `rad2Deg = 45.0/tempReal`
  in the prologue, where `tempReal` is `atan(1)` — unrelated to the `tempReal` summed
  in the steady loop 270 lines later. C bodies reuse `tempReal` freely, so reporting is
  gated on being inside a loop.

Four remain and are annotated with what was concluded, not suppressed:

| function | divisor | conclusion |
|---|---|---|
| HT_DCPHASE / HT_SINE / HT_TRENDMODE | `smoothPeriod` | `period` is clamped to [6,50] before `smoothPeriod = 0.33*period + 0.67*smoothPeriod`, a positive convex combination, so it cannot reach zero |
| KAMA | `sumROC1` | **open, not cleared** |

## The KAMA row

KAMA divides by the same running `sumROC1` as ER, under the same asymmetric clamp, and
without the exact denominator test ER gained in #350. That is the shape of the defect
this card was opened for.

I did not change it. `er.c:56` records the difference deliberately — "the one thing
kama.c has no equivalent of" — and the #378 review asked for a fix that closed ER
"without touching KAMA parity". Whether the two should converge is a decision about a
shipped function's output on the absorption edge, not something to patch so a new sweep
comes out green. The entry keeps the question visible and the sweep honest.

## Scope

The check finds divisors that nothing tests. It does not claim a division it passes is
safe: a guard on the right variable can still be the wrong guard, which is precisely
what ER's clamp was. That half is what the adversarial review on new functions already
does; this covers changes to existing ones, which the review does not see.

Verified: `cargo test` green across all 33 suites; the tree carries no generated-source
change, and `kama.c` is byte-identical to `dev`.

### DIFF
--- a/src/tools/ta_regtest/ta_test_func/test_per_hlc.c
+++ b/src/tools/ta_regtest/ta_test_func/test_per_hlc.c
@@ -171,6 +171,22 @@ static TA_Test tableTest[] =
    /*   CCI TEST  */
    /****************/
 
+   /* KNOWN AND DELIBERATELY UNCOVERED: none of the CCI rows below can see which
+    * KIND of zero guard cci.c:127 uses. Degrading both TA_IS_ZERO_SCALED calls
+    * there to the fixed-band TA_IS_ZERO leaves `ta_regtest --function=CCI`
+    * entirely green -- measured, and the green was sabotage-proved (perturbing
+    * cci.c's own 0.015 to 0.0151 turns this same invocation red at index 0,
+    * 66.666 -> 66.225166, exit 122, so the group does run CCI).
+    *
+    * No choice of test DATA closes it, for the reason test_kdj.c:211 records
+    * about its small-magnitude leg: TA_IS_ZERO_SCALED is a RELATIVE dead-zone,
+    * so the fired/not-fired ratio is invariant under rescaling and a fixed band
+    * is indistinguishable from it at every magnitude. Separating the two
+    * requires mutating the guard, which is a mutation test rather than a test
+    * case -- costed and declined on issue #381. Do not "fix" this by adding a
+    * differently-scaled series.
+    */
+
    /* The following two should always be identical. */
    { 0, TA_CCI_TEST, 186,187,  2, 0, 0, TA_SUCCESS,   1, 0.0, 186,  2 },
    { 0, TA_CCI_TEST, 187,187,  2, 0, 0, TA_SUCCESS,   0, 0.0, 187,  1 },
--- a/ta_codegen/generator/tests/divisor_guard_suite.rs
+++ b/ta_codegen/generator/tests/divisor_guard_suite.rs
@@ -1,0 +1,513 @@
+//! Divisions by a loop-accumulated variable must be guarded on that same variable.
+//!
+//! Three defects in one week shared one shape (#381): the divisor was a variable
+//! accumulated across steady-loop iterations by add/subtract, and the guard that
+//! dominated the division tested *something else*.
+//!
+//!   - VORTEX divided by a running true-range sum gated on a flat-bar counter. FP
+//!     absorption drives the sum to exactly 0.0 while the counter is still short.
+//!   - VORTEX again, after the first fix: a reseed zeroed the two numerator sums on a
+//!     predicate that only proves the denominator's terms are zero, producing a
+//!     negative -VI — unreachable for a ratio of sums of fabs().
+//!   - ER used its asymmetric KAMA-parity clamp (`sumROC1 <= periodROC`) as the zero
+//!     guard. It cannot fire when `periodROC < 0`, so a zero denominator divided.
+//!
+//! No existing gate sees this class. All four backends are generated from one input
+//! body, so `--xlang-hash` agrees bitwise on the same wrong answer; `--codegen` has no
+//! frozen baseline for a post-cutover function; and the reference corpus is benign in
+//! exactly the ways that matter (no bar with TR == 0, no 14-bar halt, no consecutive
+//! close ratio outside Sterbenz range).
+//!
+//! Data cannot settle it either. `TA_IS_ZERO_SCALED` is scale-invariant, so no series
+//! magnitude distinguishes it from a fixed band — `test_kdj.c`'s small-magnitude leg is
+//! documented as unable to discriminate for that reason. The question is structural, so
+//! the check is structural: it reads the parsed body, costs milliseconds, and runs on
+//! the PR gate that already exists.
+//!
+//! What it does NOT claim: a division this sweep passes is not proven safe. A guard on
+//! the right variable can still be the wrong guard (ER's clamp names `sumROC1` and was
+//! still unable to fire on a negative `periodROC`). This finds divisors nothing tests,
+//! which is the cheap half.
+
+use std::collections::HashSet;
+use std::path::Path;
+
+use ta_codegen_lib::{
+    ir::{BinOp, Expr, FuncDef, Statement, VarType},
+    parser,
+};
+
+fn load() -> Vec<FuncDef> {
+    let base = Path::new(env!("CARGO_MANIFEST_DIR")).join("../input");
+    let mut funcs = Vec::new();
+    for e in std::fs::read_dir(&base).expect("read input dir").filter_map(Result::ok) {
+        let dir = e.path();
+        if !dir.is_dir() {
+            continue;
+        }
+        let n = e.file_name().to_string_lossy().to_string();
+        let (yaml, csrc) = (dir.join(format!("{n}.yaml")), dir.join(format!("{n}.c")));
+        if !yaml.exists() || !csrc.exists() {
+            continue;
+        }
+        let mut f = parser::yaml::parse_yaml(&yaml);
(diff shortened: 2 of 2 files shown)


mario4tier:
Both asks are honoured, and the annotation design is better than what I described.

`ANNOTATED` carrying a *reason* rather than a bare allowlist is the right shape, and
the sentence that makes it work is the one about expiry — "a bare allowlist states
nothing and silently keeps holding once its reason expires". The HT trio's entry
recording the `[6,50]` clamp "so that removing the clamp has something to contradict"
is exactly that in practice.

`the_sweep_detects_a_reintroduced_er_defect` is two-sided in a way I did not think to
ask for: silent on ER as shipped, loud once the exact test is stripped. Asserting the
quiet half is what stops a matcher that has silently stopped matching from reading as
a clean corpus.

**And it earned its place on the first run.** I verified the KAMA entry: `kama.c:150`
and `:206` guard with `sumROC1 <= periodROC` alone and then divide
`fabs(periodROC/sumROC1)` — the same asymmetric clamp, no `sumROC1 <= 0.0` in front.
`er.c:56` names the difference outright ("the one thing kama.c has no equivalent of").
Agreed it belongs OPEN and not cleared: closing it moves KAMA's output, which is a
decision about the function, not something to slip under a sweep. Worth its own card
once this lands, so the question outlives the annotation.

**One blocker:** `Rust clippy` is failing (34s, real — not the stale run I first saw).
Everything else is green and `ta_codegen output up-to-date` is still pending. Over to
you for that.

Nothing else from me — this is yours, and I am not going to re-litigate the design.

kevinlincg:
Both asks are in, and the sabotage requirement earned its keep immediately — **VORTEX
failed it.** The sweep reported nothing on the reintroduced defect.

## Why it was blind to VORTEX

The divisor is a per-bar *copy* of the accumulator: `curTR = sTR;` and then
`curVMP / curTR`, while `+=`/`-=` build up `sTR`. Nothing in `curTR`'s own definition
is additive, so it never entered the accumulated set. Copies are now propagated to a
fixpoint (a copy of a copy is the same quantity), and the reintroduced counter-gated
divide goes red.

The sabotage rewrites `curTR > 0.0` into `nullRun < optInTimePeriod` rather than
deleting the guard, so it would still fail if the sweep only noticed "no guard at all"
instead of "no guard on the divisor". Both that and the ER strip run as tests, so
neither can rot silently.

## Two more, one each way

**VWMA — open, not patched.** `vwma.c:115,124` reads `tempV = sumV;` then
`(tempPV/P) / (tempV/P)` with **no guard at all**. `sumV` is a rolling add/subtract sum
of `inVolume`, so it is exactly 0.0 on any window whose bars all report zero volume — a
halted session, or a feed that pads non-trading days — and reachable through absorption
on windows that are not all-zero. Unlike KAMA this is not a documented deviation; it is
simply unguarded. Recorded with its reasoning rather than fixed, since it is a decision
about a shipped function's output.

**MAMA — mine.** It divides under `if( tempReal > 1.0 )`, and a positive lower bound
establishes non-zero as well as a zero test does. Only *literal* bounds count, so ER's
`sumROC1 <= periodROC` — a variable that can be negative — still fails the check. The
ER test is what pins that distinction.

## CCI

Recorded at `test_per_hlc.c`, immediately above the CCI rows that cannot see it, in the
shape `test_kdj.c:211` uses: the measured green, the sabotage that proves the group
runs (`0.015` → `0.0151`, red at index 0, 66.666 → 66.225166, exit 122), and why no
test data closes it — a relative dead-zone's fired/not-fired ratio is invariant under
rescaling. It also says not to "fix" it by adding a differently-scaled series, so the
next reader does not spend the cycle I did.

## State

`cargo clippy --all-targets` clean, 938 generator tests green, `ta_regtest` green. The
tree carries no generated-source change; `kama.c` and `vwma.c` are byte-identical to
`dev`.

Standing entries in the annotation table, all four with reasons rather than
suppression: the Hilbert trio (`period` clamped to [6,50] before a positive convex
combination), plus KAMA and VWMA as open questions.

mario4tier:
The added scope is right, and two of the three additions were things I asked for — the
CCI hole documented in-code with its sabotage proof and the cross-reference to
`test_kdj.c:211`, and the second self-test using VORTEX's counter-gated divide. Good.

**One correction: VWMA is not an open question — it is the decided precedent.**

`vwma.c:124` divides by the rolling volume sum with no guard, and `vwma.yaml:4`
declares `nan_inf_output`. That is deliberate, and it is the standing answer for this
exact shape (it was set as the precedent for #56 RVOL and anything like it — not
`1.0`, not `0.0`, not carry-forward, which would force `TA_FUNC_FLG_START_DEP` onto an
otherwise windowed function).

The reasoning that settles it: **#112 does not demand a guard where the function is
stateless per bar.** Its concern is a NaN that *poisons an accumulator* — #39's KVO B2
is the live example. VWMA has no accumulator across bars, so there is no poisoning and
no guard is owed. A non-finite value there is contained to the bar that produced it,
which is what the declared flag says.

So the sweep is right that VWMA is unguarded and wrong to call it open. Suggest
re-annotating it as a clearance rather than a question, keyed to the flag:

    ("VWMA", "tempV", "unguarded by decision: stateless per bar, declares nan_inf_output")

That also makes the annotation self-checking in the way your own doc comment argues
for — if someone ever removes `nan_inf_output` from `vwma.yaml`, the stated reason is
contradicted and the entry should be revisited.

**KAMA stays genuinely OPEN** — that one has an accumulator (`sumROC1` is carried and
reused), so the #112 reasoning does not clear it, and `er.c:56` already records the
asymmetry. Worth its own card once this lands.

Clippy still needs whatever it needs; I have not looked at that beyond the red.

kevinlincg:
Corrected — VWMA is a clearance, and clippy was already fixed in the commit before your
review (you were looking at the 34s run from the first push; the rerun passes).

## VWMA

Re-annotated as you suggested, with the reasoning that actually does the clearing
written down rather than just the flag:

    ("VWMA", "tempV", "unguarded by decision: stateless per bar", "nan_inf_output"),

The comment above it records that statelessness is what settles it — #112's concern is
a NaN that *poisons an accumulator*, and VWMA carries nothing across bars, so a
non-finite value is contained to the bar that produced it. KAMA is named right there as
the contrast: it carries `sumROC1`, so the same argument does not reach it and the OPEN
entry stands.

I verified the flag independently before taking the correction: `vwma.yaml:4` is
`flags: [overlap, stream, nan_inf_output, period1_identity]`, `kama.yaml:4` has no such
flag, and eight functions declare it tree-wide.

## The flag is now asserted, not trusted

Your own argument for the table is that a stated reason can be contradicted where a
bare allowlist cannot — which only holds if something checks. Entries can now name a
YAML flag their reason rests on, and `annotation_reasons_still_hold` asserts it is
still declared:

> VWMA's divisor `tempV` is annotated "unguarded by decision: stateless per bar", which
> rests on the `nan_inf_output` flag — and VWMA no longer declares it. The reason has
> expired: re-read the division rather than re-adding the flag to silence this.

That is the message from an actual run with the flag removed from `vwma.yaml`; restored,
it is green again. Same two-sided treatment as the ER test — an assertion that only ever
passes is the failure mode this whole card is about.

The HT trio and KAMA carry an empty flag field, since their reasons rest on a clamp and
on an open question rather than on declared metadata.

## State

`cargo clippy --all-targets` clean, 939 generator tests green. `ta_codegen output
up-to-date` was pending on the earlier run rather than failing; this push should settle
it either way, and the tree still carries no generated-source change.

Happy to open the KAMA card once this lands, unless you would rather write it yourself.

PR #391, opened 2026-09-06
test(serve): C gains the short-history reject leg the other three have (#386)

kevinlincg:
Closes the C half of #386 B-3.

## The gap

`shortHistoryAccepted` reads 201 in Rust, Java and C#, and **0** in
`ta_codegen_serve.c`. `server_gen.rs` emits the leg at three sites — `:6874`,
`:7726`, `:9167` — none of them C. So C's `ok` could not fall for an `Open` that
accepts a history no output is defined over.

Worth naming, since it changes what the fix has to do: **`shortHistoryAccepted` has
no consumer.** Nothing in `server_verify.c` or anywhere else parses that key; it is a
diagnostic string, and the actual gate is that the leg clears `allOk`, which folds into
`ok`. So the fix is not "emit the key in C" — it is running the check at all.

## The leg

Open at exactly `lb` bars and require rejection with `TA_INSUFFICIENT_HISTORY`.
Accepting is the defect; rejecting with some other code is a second, distinguishable
one — the same split Java and C# already make between the typed exception and an
`IllegalArgumentException`/`ArgumentException`.

Reported as `short_history_checked` / `short_history_ok` / `short_history_bad` rather
than through a `diag` string: C has no `diag` mechanism, and it already spells exactly
this shape as a checked/ok/bad triple for `clone` and `value`. The counter also makes
"this leg never ran" visible, which a conditional diagnostic string cannot do — the
vacuity failure mode #386 B-2 describes for the Java/C# `Value()` floor.

## Sabotage-proved, both directions

A leg that cannot fail is worse than no leg, so it was made to fail before being
trusted. Opening at `svN` bars instead of `lb` (where `Open` must succeed):

```
TA_SMA: ok=0  short_history_checked=1  short_history_ok=0
        short_history_bad='open accepted a history shorter than one output'
TA_RSI: ok=0  short_history_checked=1  short_history_ok=0   (same)
```

Restored, both read `ok=1  short_history_checked=1  short_history_ok=1`. The
`checked=1` in the passing run is what rules out the third possibility — that the leg
silently never executed.

Also confirmed on the shipped binary rather than the source: the two new strings are
present in `bin/ta_codegen_serve_c` and a control string is not, so the probe can
distinguish a stale binary from a fresh one.

## Verification

- `ta_regtest` green.
- `cargo clippy --all-targets` clean; 935 generator tests green.
- Live responses: `TA_SMA`, `TA_EMA`, `TA_RSI`, `TA_ATR`, `TA_MACD` all report
  `short_history_checked=1, short_history_ok=1`.
- Longest sampled response is 641 bytes against a 16 MB buffer, so the three added
  fields are not near any limit.
- **Zero drift**: `git status` shows only `server_gen.rs` and
  `ta_codegen/output/c/tools/ta_codegen_serve.c`. The Rust, Java and C# servers are
  byte-identical, which is the evidence that the C-only emit did not perturb them.

## What I could not run, and why I do not think it blocks

`scripts/build.py servers` fails on this machine before it reaches anything of mine:
the JDK rejects `release version 17` and the .NET SDK is 7.0.200 against a net10.0
target. That blocks the four-server `stream_verify` sweep through `ta_regtest`.

Stating why I still think the change is verified rather than merely asserted: the
change is C-emit-only and the other three servers are byte-identical, so there is
nothing in those paths for the sweep to falsify. The C leg itself is exercised
directly against `ta_codegen_serve_c` and sabotage-proved in both directions above,
and the driver does not read the new keys, so the added fields cannot change how it
interprets a response. If you would rather see the full four-server sweep before
merging, say so and I will treat that as blocking rather than argue the point.

### DIFF
--- a/ta_codegen/generator/src/server_gen.rs
+++ b/ta_codegen/generator/src/server_gen.rs
@@ -2326,6 +2326,11 @@ fn generate_c_stream_verify(funcs: &[FuncDef], enums: &HashMap<String, EnumDef>)
         s.push_str("        int valueChecked = 0, valueOk = 1, valueLegs = 0;\n");
         s.push_str("        const char *valueBad = \"-\";\n");
         s.push_str("        const char *cloneBad = \"-\";\n");
+        // Short-history reject leg. Rust, Java and C# have carried this since
+        // the streaming tier landed; C never emitted it, so `ok` could not fall
+        // for an Open that accepts a history no output is defined over.
+        s.push_str("        int shortHistChecked = 0, shortHistOk = 1;\n");
+        s.push_str("        const char *shortHistBad = \"-\";\n");
         s.push_str("        const char *peekBad = \"-\";\n");
         s.push_str("        int fillOk = 1, fillChecked = 0, fillBars = 0;\n");
         emit_sv_state_decls(&mut s, name, steq);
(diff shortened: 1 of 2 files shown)


mario4tier:
Merged, with a follow-up fix on dev: 7625e2590.

**What needed fixing.** The leg is emitted after the handler restores `TA_SetUnstablePeriod(id, 0)` and `TA_SetCompatibility(savedCompat)`, but `lb` is computed earlier with `svK` still applied. So for the 43 unstable-period functions it compared a K-inclusive lookback against a K=0 library and reported every K>=1 case as `open accepted a history shorter than one output`. ADX at K=3: `lb` = 30, true K=0 lookback 27, so `Open(30 bars)` legitimately succeeds.

The library is fine — `TA_ADX_Open` at `lb` bars answers `TA_INSUFFICIENT_HISTORY` at every K when K is still set. The fix moves the leg above both restores; the generated diff is a pure reordering, 303 lines out and 303 back.

Worth naming why this was invisible from where you were standing: at K=0 the two agree, so the SMA/EMA/RSI/ATR/MACD responses you sampled all read green — the split only opens at K>=1. And no PR job *runs* `ta_regtest`, it only compiles it, so the gate stays green on the PR either way. Your call to name the missing four-server sweep as the risk was the right one, and it is exactly where this landed.

After the fix: 201 functions x K in {0,1,2,3,5,10} = 1206 requests, 1007 legs actually run, 0 failures; `ta_regtest --codegen --language=c` green; dev nightly green including `xlang matching (C vs Rust + Java + C#)` and the streaming ASan/UBSan/LSan job.

Three smaller things, none of them blocking:

1. `(void)stSH;` never reaches the output — `c_hygiene::scrub_void_casts` removes it because the enclosing block reads `stSH` (zero occurrences in the generated file). The `push_str` can go.
2. "Rust, Java and C# have carried this since the streaming tier landed; C never emitted it" is change history; CLAUDE.md puts that in the commit message rather than in the source.
3. The asymmetry now runs the other way: C checks the retCode, while Rust/Java/C# only check acceptance. Worth a follow-up under #386 so all four legs draw the same distinction.

Issue #389, opened 2026-09-06
Coverage gap: a non-success OpenAndFill can write the caller's buffer (RSI/CMO under Metastock), and no gate is shaped to catch it


mario4tier:
`TA_RSI_OpenAndFill` and `TA_CMO_OpenAndFill` return `TA_INSUFFICIENT_HISTORY` after having already written the caller's output buffer and both out-params. Reproduced on dev `2ef8a8cfd` against `libta-lib.so.0.8.1`:

```
--- DEFAULT ---
RSI      lb= 14 rc= 0 begIdx=   14 nb=    1 out[0]=50
CMO      lb= 14 rc= 0 begIdx=   14 nb=    1 out[0]=0
--- METASTOCK ---
RSI      lb= 13 rc=17 begIdx=   13 nb=    1 out[0]=66.6667   <-- VIOLATION
CMO      lb= 13 rc=17 begIdx=   13 nb=    1 out[0]=33.3333   <-- VIOLATION
```

(sentinels: `begIdx`/`nb` pre-set to `-999`, `out[]` to `-777`; anything else is a write.) The failed call did not merely set the indices — it wrote a **plausible RSI value**. A caller that trusts `outNBElement` over `rc` consumes one element of real-looking data from a call that failed. `website/src/api/README.md:181` states the opposite: on non-success, treat the indices as undefined and the buffers as untouched.

Under Metastock `TA_RSI_Lookback` reports **13** while `Open` requires **15** bars, so following the documented `historyLen >= Lookback + 1` rule lands exactly on the violating path. That half is #386's RSI/CMO entry; this issue is the coverage gap.

All four backends share the emitter shape — the transcribed batch prologue writes `outBegIdx`/`outNBElement`, the Metastock seed writes `outReal[0]`, and only then does `startIdx > endIdx` return `InsufficientHistory`: `ta_RSI.c` `TA_RSI_OpenImpl`, `rsi.rs:502 rsi_open_impl`, Java `rsiOpenImpl`, C# `RsiOpenImpl`.

The ordinary short-history path is clean — SMA, ATR and CMO-under-default leave all three sentinels intact, because the early `startIdx > historyLen - 1` guard answers before the body runs. Only a body that runs far enough to write and *then* discovers it has nothing to produce can violate this.

## Why phantom I/O did not catch it

Two independent reasons, neither trivial:

[...]

Issue #388, opened 2026-09-06
Remove Metastock compatibility (TA_Compatibility) from the TA functions


mario4tier:
Retire `TA_Compatibility` — the Metastock variant — from the TA functions, which are the source of truth. `TA_SetCompatibility()` / `TA_GetCompatibility()` are already deprecated (`ta_defs.h:160`, `ta_func.h:14028`); this removes the behaviour they select.

## Why removal, and not "make streaming ignore it"

The narrower alternative — have the streaming tier always behave as `TA_COMPATIBILITY_DEFAULT` — was considered and is worse. Bit-identity with batch is the streaming API's defining promise, and the semantic definition in `docs/streaming-api-design.md` qualifies it *"under the same compatibility and candle settings"*; `stream_verify` runs a Metastock leg specifically to prove it. Making streaming ignore compatibility would have batch and stream deliberately disagree under Metastock — a permanent exception written into the core contract of the headline feature, as scaffolding for something already slated for deletion.

## Scope

Compatibility is narrower than it looks. 12 functions branch on it, and most inherit it:

`cmo` `dema` `efi` `ema` `macd` `macdfix` `massi` `rsi` `smi` `tema` `trix` `tsi`

EMA's seeding is the root — DEMA, TEMA, TRIX, MACD and MACDFIX all cascade it. The genuinely distinct primitives are EMA's seed and RSI/CMO's seed-and-rewind.

To remove:

- `ta_codegen/input/{cmo,dema,efi,ema,macd,macdfix,massi,rsi,smi,tema,trix,tsi}/` — the compatibility branches in the C logic.
- `ta_codegen/generator/src/backends/compat_fold.rs` — a 212-line pass that exists only for this.
- `TA_Compatibility`, `TA_COMPATIBILITY_DEFAULT`, `TA_COMPATIBILITY_METASTOCK`, `TA_SetCompatibility`, `TA_GetCompatibility`, `TA_GLOBALS_COMPATIBILITY`, and the per-`Core` equivalents in Rust/Java/C#.
- The Metastock leg in ~16 gate files, including `server_verify.c`, `test_codegen.c`, `indicator_variants_suite.rs`, `streaming_suite.rs` and eight hand-written `ta_test_func/` suites. Census: `grep -rln "METASTOCK\|Metastock" src/tools/ta_regtest/ ta_codegen/generator/tests/ scripts/`.
[...]

PR #393, opened 2026-09-06
test(regtest): a rejected Open/OpenAndFill must not write the caller's buffer (#389)

kevinlincg:
Closes the coverage half of #389: the gate, corpus-wide, in the one backend
where the defect is reachable. It does not move RSI or CMO.

## What the sweep does

For every entry in the generated `TA_StreamTable` (201 functions), under both
compatibility modes, through both `Open` and `OpenAndFill`: arm the caller's
output buffers with a sentinel, call, and on any non-success return require
them back byte-identical.

Two rejection classes, deliberately both:

- **History** — `historyLen` ramps up from 0 and the leg stops at the first
  length that produces a value, so the last rejection it sees is the one at the
  boundary, whatever the boundary turns out to be. That is the point: under
  Metastock `TA_RSI_Lookback` reports 13 while `Open` needs 15 bars, so a probe
  computed from the lookback lands one bar away from the violating call.
  Nothing in the leg knows a lookback.
- **Parameter** — every optional parameter driven one step outside each of its
  declared bounds, on a history that otherwise produces. This is the class the
  public frame answers before the body runs, and having both is what makes
  "only the in-body class can write" a measurement rather than a claim.

## What it finds

`TA_RSI` and `TA_CMO`, under Metastock, on `OpenAndFill`, and nothing else:

```
OPEN-CONTRACT TA_CMO METASTOCK OpenAndFill: retCode=17 wrote the caller's buffer (outBegIdx=13 outNBElement=1)
     output 0 [0] = 99.1494
OPEN-CONTRACT TA_RSI METASTOCK OpenAndFill: retCode=17 wrote the caller's buffer (outBegIdx=13 outNBElement=1)
     output 0 [0] = 99.5747
```

(The values differ from #389's because the series does; the shape is identical
— `TA_INSUFFICIENT_HISTORY` with a plausible oscillator value already in
`outReal[0]` and `outNBElement` reporting one element.)

Both are listed in `ocKnownOpen` and **asserted to still violate**, so the run
is green today and a listed row that stops violating fails it. Same treatment
#383 gave PPO/PVO and STOCH/STOCHF: moving the write is a decision about a
shipped function, not something to slip in under a gate.

## The numbers

| | |
|---|---|
| functions swept | 201 (`TA_STREAM_TABLE_SIZE`), x 2 modes |
| rejections whose buffers were compared | 14 762 |
| of those, out-of-range parameter probes | 1 360, **none accepted** |
| per-function positive controls | 402 (201 x 2), all passed |
| rejections leaving `outNBElement` non-zero | **2** — the two rows above |
| added runtime | under 25 ms; the group runs faster than the smallest existing indicator group (`--function=CVI`) |

## Controls, each one broken and watched go red

Not asserted — run.

1. **`ocKnownOpen` emptied** → red, naming `TA_RSI` and `TA_CMO` under
   `METASTOCK` with the value each wrote. This is the live proof the sweep sees
   the real defect.
2. **A row that no longer violates** (added `{"SMA", 1}`) → red:
   `TA_SMA is listed as a known open write-on-rejection (#389) under Metastock,
   but the sweep no longer sees it.`
3. **The `outNBElement` ceiling** lowered by one → red, printing both offenders.
4. **The positive control**: output pointers aimed at a decoy buffer, so the
   functions write somewhere the comparison never looks → red at the first
   function (`TA_AC DEFAULT: OpenAndFill reported 1 elements at historyLen=38
   and wrote nothing`). Without this arm, "nothing was written" is satisfied by
   a fixture the function never writes through, over all 201 functions.
5. **An in-range parameter probe** (`min + 1` instead of `min - 1`) → red, 572
   probes accepted. This one was not hypothetical: with the obvious `min - 1`,
   **58 of the probes were no-ops**, because a real parameter's declared bounds
   are ±3e37 where adding one is the identity. `oc_out_of_range` scales the step
   for reals and keeps the exact ±1 for integers and enums.

## Two judgement calls worth naming

**`outBegIdx`/`outNBElement` are ratcheted, not asserted.** §3.4 of
`website/src/api/README.md` calls them *undefined* on a non-success return, and
most bodies legitimately zero them on the no-data path, so asserting them would
be inventing a rule. But a rejection leaving `outNBElement` non-zero is exactly
what turns an invisible write into a value a caller trusting the count reads —
and over 402 function-modes it happens twice, both of them the listed pair. So
it carries a ceiling equal to the known-open row count, which drops with the
list rather than being a second constant to forget. If you would rather this
gate said nothing at all about the indices, it is one `if` to delete.

**The parameter leg asserts that an out-of-range parameter is rejected.** That
is slightly more than #389 asked for. It is here because the leg is worthless
if the probe is not actually out of range (see control 5), and because
`test_abstract`'s equivalent drives `TA_CallFunc`, i.e. batch — the streaming
open tier had no such check. Zero acceptances today.

## Why C only, and why that is not a gap to close later

- **Rust cannot host it.** `Compatibility` is `pub(crate)`, pinned to
  `Default`, with `compatibility_is_pinned_to_default` asserting there is no
  setter. No Rust probe can reach the Metastock seeding path.
- **Java and C# could, but `codegen_lang_has_compatibility_api` already records
  the standing decision** that the ported backends expose no compatibility API
  and the Metastock legs are C-only, permanently.
- **The undersized-output class has no C expression**: `TA_<N>_Open` takes a
  bare output pointer and carries no length. It stays where it is, in the Rust
  crate's `stream_open_contract.rs`.

So the sweep is C-only for the same reason `stream_verify`'s state leg is, and
the emitter it exercises is shared: the four backends transcribe one body.

## The fix, and why it is not in this PR

#389 offers two directions and the choice is yours:

- **Structural** — have the open frame commit `outBegIdx`/`outNBElement` and
  the fill only on the success return. Free for the two scalars; **not** free
  for the fill, which would need a scratch buffer and a copy on every
  *successful* `OpenAndFill`, since the rejection is raised after the writes
  and there is nothing to un-write.
- **Reorder the seed** — the write at `rsi.c:164-171` happens before the
  `today > endIdx` test that follows it. Hoisting the test costs nothing at run
  time, but the generator does not derive "this body needs one more bar under
  Metastock" today; it would be new analysis in the open transform
  (`c_stream.rs`'s early-success mapping, which already names RSI/CMO in its
  comment).

And #388 may delete the compatibility mode outright, which removes this
instance without removing the class — which is why the gate is worth landing
either way.

## Verification

- `ta_regtest` full run green, with the new group; `--function=CONTRACT` green
  on its own.
- `cd ta_codegen/generator && cargo run -- generate` → **zero drift**;
  `git status` shows only the six files this PR touches, none of them
  generated. The change is test-side only: no generator, no input, no shipped
  library file.
- `scripts/build.py check-source-lists`: *"ta_regtest source lists agree across
  CMake and autotools (77 files)"*, after registering the new file in both.

**What I did not run.** `--codegen` and `--xlang-hash` (this machine has no
`bin/ta_ref_serve`, and its JDK/.NET are too old for the Java/C# servers), and
an actual autotools build — the Makefile.am entry is verified by
`check-source-lists`, not by a configure-and-make. None of the three can be
affected by a change that adds one `ta_regtest` translation unit, but I would
rather say so than imply otherwise.

---

## Verification run before opening

The branch was built and exercised end to end, and both directions of the
known-open mechanism were sabotage-proved rather than taken on trust:

- **The sweep catches a real leak.** Removing `RSI` from `ocKnownOpen` makes the
  suite fail immediately with `OPEN-CONTRACT TA_RSI METASTOCK OpenAndFill:
  retCode=17 wrote the caller's buffer (outBegIdx=13 outNBElement=1)` — the exact
  defect #389 describes, found by the sweep rather than by a hard-coded probe.
- **A stale known-open entry is caught too.** Adding `SMA` (which does not leak)
  to that list fails with "listed as a known open write-on-rejection (#389) under
  Metastock, but the sweep no longer sees it", so the list cannot outlive the
  defect it records.

Restored, `ta_regtest` is green and the tree is clean. The vacuity floors
(`TA_OPEN_CONTRACT_VACUOUS`) require the full `TA_STREAM_TABLE_SIZE` corpus,
non-zero rejections including parameter probes, and a controlled producing call
per function per compatibility mode, so a sweep that quietly stopped covering
anything fails rather than passes.


kevinlincg:
Updated for #388, which landed on `dev` while this sat open and changed the outcome
rather than just the base.

## The known-open list is now empty, and that is the mechanism working

#388 retires the Metastock compatibility behaviour. That behaviour is what made
`TA_RSI_OpenAndFill` and `TA_CMO_OpenAndFill` write the caller's buffer before
returning `TA_INSUFFICIENT_HISTORY`, so the two rows stopped violating and the
stale-entry check failed the run:

```
Fail: TA_RSI is listed as a known open write-on-rejection (#389) under Metastock,
      but the sweep no longer sees it.
      If that is the fix landing, delete the entry from ocKnownOpen so the row
      stops claiming an open defect.
```

Rows removed. The header now records the outcome instead of the prediction it was
carrying — it said "issue #388 may delete the compatibility mode outright", and it did.
`OC_NB_KNOWN_OPEN` drops to zero with the rows, and since it doubles as the
`outNBElement` ceiling, that ratchet tightens to zero at the same time rather than
staying at a stale 2.

## The leg still discriminates with an empty list

Worth checking rather than assuming — an allowlist emptying out is exactly when a sweep
can quietly become vacuous. Injecting the #389 defect into a function that never had it
(a write to `outReal[0]` before SMA's lookback-driven rejection) fails immediately:

```
OPEN-CONTRACT TA_MA DEFAULT OpenAndFill: retCode=17 wrote the caller's buffer
```

Reverted, `ta_regtest` is green. The first sabotage attempt went into the wrong return —
`OpenAndFill` delegates through `OpenAndFillInternal` to `OpenImpl`, and the ramp reaches
the lookback-driven rejection, not the `startIdx > historyLen-1` one — so the probe was
placed by reading which return the ramp actually hits.

## One consequence worth flagging

With #388 landed, `TA_COMPATIBILITY_DEFAULT` and `TA_COMPATIBILITY_METASTOCK` are the
same behaviour, so this sweep's two-mode loop is now a redundancy check rather than two
behaviours. It still costs nothing and guards against the modes diverging again, but the
history ramp is what carries the leg now. Noted in the header rather than left for
someone to rediscover.

mario4tier:
Superseded by #399, which carries your commits — the sweep, the fixture, the `TA_StreamTable` drive, the per-function positive control and the vacuity floors are all yours and unchanged in substance.

What #399 adds on top, in case it is useful context:

#388 landed after you opened this, and emptying the known-open list left `static int ocKnownOpenHit[OC_NB_KNOWN_OPEN]` at zero length. That is a GNU extension MSVC rejects, and since no PR gate compiles `ta_regtest`, only the Windows nightly would have caught it — not something your box could have shown you.

Three coverage holes turned up by injecting defects into the sweep and watching it stay green:

- the positive control accepted "at least one output slot moved", so a fixture mis-binding slots `1..n-1` of a multi-output function read fully green
- the `Open` leg had no positive control of its own
- leg B counted any non-success as proof the parameter was rejected, but an out-of-range period run at the shortest producing history trips the insufficient-history guard too — so the upper half of every declared bound was unverifiable, and deleting one stayed green

Your control-5 finding (a real parameter's `min - 1` landing back inside the range) was the model for all three: the probe has to be checked for actually probing.

Thanks — the machinery held up under a lot of adversarial poking.

mario4tier:
Landed via #399, merged to `dev` as `122507748`. Your commits went in with it — `e234728cb` is on `dev`. Closing this one as superseded rather than unmerged.

PR #413, opened 2026-09-11
feat(rvir): add TA_RVIR, the 1995 refined Relative Volatility Index (#416)

### LINKED ISSUE #416: (RVIR) Relative Volatility Index, 1995 refined form
## What it is

Donald Dorsey, **"Refining the Relative Volatility Index", *TASC* V.13:9 (Sep 1995) 388-391** — the
revision of his V.11:6 (Jun 1993) close-only RVI. It runs the 1993 RVI over daily highs and over
daily lows and averages the two indices; the closes are the fallback when a feed carries no
high/low (Chart manual).

The leg this function is made of ships: `TA_RVI` landed 2026-09-05 in commit `8c0fedbcd`
(TA-Lib/ta-lib#366), `ta_codegen/input/rvi/{rvi.c,rvi.yaml,rvi.md}` at HEAD. RVIR is that body run
twice and averaged; it invents nothing, and it inherits every variant choice that card made.

The refined form is a different series, not a re-parameterisation: **MEASURED at (14,10) on the
252-bar corpus, max |RVIR − RVI(close)| = 18.05 index points, mean 5.52, Pearson r 0.709.**

**Name collision, and it must be said on both docs.** The Chart manual (*"which is the RVI used in
Chart"*) and DXcharts reserve the bare name **RVI** for *this* function and call
TA-Lib/ta-lib#366's function **`RVIorig`**; pandas-ta-classic and trading-signals default the other
way. `TA_RVI` has shipped under the close form, so the pair cannot be swapped. `rvi.md:21` already
carries the collision note; `rvir.md` owes the reciprocal.

## Formula

`n1` = standard-deviation window (default 10), `n2` = Wilder smoothing period (default 14).
`RVI(...)` is the shipped `TA_RVI` **unchanged** — its ties-feed-neither rule
(`ta_codegen/input/rvi/rvi.c:137-141`) and its `50.0` on a zero denominator (`rvi.c:225`) included.

```
RVIR_i = 0.5 * ( RVI(high)_i + RVI(low)_i )
```

**Primary-source trace.** The 1995 article's PDF is paywalled ($3.95) and was **not** purchased; the
store listing was fetched (HTTP 200) for title, author, pagination and the opening two paragraphs,
which are about trend and inertia and do not state the arithmetic. The arithmetic is traced instead
through a MetaStock formula-language reproduction of the article, published under the heading
*"New Relative Volatility Index"* and labelled *"Appeared in the September 95 issue of Stocks &
Commodities magazine"* — code, not prose, in the formula language the article was written into
(provenance qualified under **References**: the host credits `traders.com`, not Equis):

```
(100 * Mov(If(HIGH > Ref(HIGH,-1), Std(HIGH,10),0),14,E) /
 (Mov(If(HIGH > Ref(HIGH,-1), Std(HIGH,10),0),14,E) +
  Mov(If(HIGH < Ref(HIGH,-1), Std(HIGH,10),0),14,E)) +
 100 * Mov(If(LOW > Ref(LOW,-1), Std(LOW,10),0),14,E) /
 (Mov(If(LOW > Ref(LOW,-1), Std(LOW,10),0),14,E) +
  Mov(If(LOW < Ref(LOW,-1), Std(LOW,10),0),14,E))) / 2
```

That settles four things structurally, in one source:

1. **Two full RVI legs**, each with its own direction gate and its own `Std` — not RVI of an averaged
   price.
2. **Denominator is `EMA(up) + EMA(down)`.** A bar with `HIGH == Ref(HIGH,-1)` makes **both** `If()`
   legs zero, so its sigma enters neither numerator nor denominator — **ties feed neither**, which is
   exactly `rvi.c:137-141`.
3. **`Std` 10, smoothing 14** — the shipped defaults.
4. **No zero-denominator guard**, so TA-Lib's `50.0` is a house choice. App::Chart and
   trading-signals 8.3.0 independently pick 50 as well.

**The averaging spelling is not a variant axis.** `0.5*(a+b)`, `(a+b)*0.5`, `(a+b)/2.0` and
`0.5*a + 0.5*b` are bit-identical for operands in `[0,100]`: scaling by a power of two is exact, so
each is one rounding of `a+b`. **MEASURED: 0 mismatches in 200 000 random pairs from [0,100].** No
spelling gate is owed.

Both legs are the same function at the same parameters, so they share a lookback and seed on the
same bar — no anchor question, no `max()`. Fuse them into one loop rather than KC's
malloc-two-scratch shape (`ta_codegen/input/kc/kc.c:81,88`): the result is bit-equal to two `TA_RVI`
calls by construction, which makes the differential gate below free.

```c
lookbackTotal = (optInStdDevPeriod - 1) + (optInTimePeriod - 1)
              + TA_GetUnstablePeriod(TA_FUNC_UNST_RVI);   /* == TA_RVI_Lookback, rvi.c:15-19 */
if( startIdx < lookbackTotal ) startIdx = lookbackTotal;
if( startIdx > endIdx ) { *outBegIdx = 0; *outNBElement = 0; return TA_SUCCESS; }

/* Two independent copies of TA_RVI's state, one over inHigh, one over inLow:
 * var.c's shifted-sum cursors plus prevUp/prevDn.  Coefficients spelled ONCE,
 * wBeta first (ta_codegen/input/rma/rma.c:51-52); both legs share them.
 * Every read of bar i in BOTH legs must precede the store to outReal[i]: with
 * outReal aliasing inHigh or inLow the slack is exactly zero at n2 == 1. */
wBeta  = (double)(optInTimePeriod - 1) / (double)optInTimePeriod;
wAlpha = 1.0 - wBeta;
   ... seed both legs exactly as TA_RVI does, then per bar ...
outReal[outIdx++] = 0.5 * ( rviHigh + rviLow );
```

`TA_RVIR_Lookback == TA_RVI_Lookback`. **MEASURED** at (14,10), unstable period 0, on the 252-bar
reference: `outBegIdx = 22`, `outNBElement = 230`. The composed trading-signals arm reports the same
`beg` at all four tuples — `(n2,n1) = (14,10) → 22`, `(30,30) → 58`, `(1,10) → 9`, `(2,2) → 2`.

### Rejected variants — MEASURED at (14,10) on `TA_SREF_{high,low}_daily_ref_0_PRIV` (252 bars, `src/tools/ta_regtest/test_data.c:38,65`), max abs deviation from the proposed form

| Variant | Source | Max dev | Why rejected |
|---|---|---|---|
| plain **EMA** `2/(n2+1)`, ties→neither | community Pine; pandas' `mamode` default `"ema"` | **12.94** (bar 73); tail bar 251 **3.80** | Dorsey defines RVI as *"the relative strength index (RSI) with the standard deviation over the past 10 days used in place of daily price change"* (V.11:6 p.253, quoted verbatim on the Equis 1993 page) — and RSI's smoother is Wilder's. The 1993 transcription is that recursion exactly, `((PREV*13)+x)/14`, alpha = 1/14. **Counter-evidence, stated because it is real:** the 1995 page writes `Mov(...,14,E)`, a plain EMA, against the 1993 page's exact Wilder form — the familiar MetaStock shorthand. It loses to the 1993 transcription and to Dorsey's own RSI sentence, and the smoother is TA-Lib/ta-lib#366's axis in any case. |
| plain **EMA** + ties→down | as above | **13.78** (bar 73) | Both flips at once; the isolated EMA cost is the row above. |
| `RVI( (high+low)/2 )` — average the *prices*, not the *indices* | none published; the reimplementer's shortcut | **13.58** | The 1995 code has two independent `Std()` calls and two independent gates. Looks like a simplification; is not one. |
| thirds: `(RVI(h)+RVI(l)+RVI(c))/3` | pandas `thirds=True` (`rvi.py:45-55`) | **6.02** at (14,10); **33.33** at L=3 and **15.86** at L=5 in pandas' *own* arithmetic | Not Dorsey's; no other source carries it. A separate indicator, not a variant. |
| tie (`x[i] == x[i-1]`) → **down** | App::Chart 275 `RVI.pm` (`den = EMA(stddev)`) | **4.64** (bar 88); tail bar 251 still **2.24e-1** | Both `If()` legs of the 1995 code are 0 on a tie. App::Chart is the documented negative control. |
| tie → **up** | ProRealCode | **3.76** (bar 81) | Only source that does this. |
| 9-period stddev | fmlabs; DXcharts prose | **3.23** (bar 83) | One source family against five; also moves `outBegIdx` 22 → 21. DXcharts contradicts itself — prose says *"9-period standard deviation"*, its own legend says *"SD - 10-days' standard deviation"* — so cite DXcharts for structure, never for the period. |

## Proposed API

```c
TA_RetCode TA_RVIR( int    startIdx,
                    int    endIdx,
                    const double inHigh[],
                    const double inLow[],
                    int    optInTimePeriod,     /* From 1 to 100000, default 14 */
                    int    optInStdDevPeriod,   /* From 2 to 100000, default 10 */
                    int   *outBegIdx,
                    int   *outNBElement,
                    double outReal[] );

int TA_RVIR_Lookback( int optInTimePeriod, int optInStdDevPeriod );
/* (n1-1) + (n2-1) + TA_GetUnstablePeriod(TA_FUNC_UNST_RVI) */
```

- Group **`Volatility Indicators`**, matching the shipped `rvi.yaml:2` and pandas' `category`. One
  output `outReal`, `flags: [line]` — well under the cap of 4 (`src/tools/ta_regtest/test_codegen.h:16`).
- **Input bundle `inPriceHL`**, `price_components: [high, low]` — the existing DONCHIAN bundle
  (`ta_codegen/input/donchian/donchian.yaml:6-8`). Arity 2. This is why the refined form cannot be a
  boolean flag on `TA_RVI`, which takes a bare `inReal` (`rvi.yaml:5-7`).
- Parameters, ranges and defaults are the shipped `rvi.yaml:8-22`, unchanged and in that order.
  App::Chart is a **third** witness that the two periods are separate and defaulted 10/14, after
  thinkorswim and trading-signals 8.3.0.
- Flags `[stream]` — **not** `unstable_period`, **not** `path_dependent`.
- **No new `TA_FUNC_UNST_*` id, and no second one either.** RVIR reads `TA_FUNC_UNST_RVI`
  (id 26, `src/tools/ta_regtest/test_internals.c:1158`), the way `efi.yaml:4` declares
  `flags: [stream]` only while `efi.c:22` reads `TA_FUNC_UNST_EMA`. KC's two-id case does not arise:
  KC inherits two because its legs are *different* functions with *different* lookbacks; RVIR's legs
  are the same function at the same parameters, so a per-leg unstable period has no user story and
  would break `outBegIdx == lookback` (TA-Lib/ta-lib#99) by over-warming one leg. One id means
  `doRangeTestEx`, not `doRangeTestMulti` — `ta_test_priv.h:254-256` still names KC as the only
  two-id function, and the shipped `test_rvi.c:856` already uses `doRangeTestEx`.

## Decisions

1. **Ties feed neither leg** — inherited from the shipped `TA_RVI` (`rvi.c:137-141`) and supported by
   the 1995 *and* 1993 MetaStock reproductions, whose denominators are `EMA(up) + EMA(down)`. Tally:
   ties→neither = 1993 formula + 1995 formula + pandas-ta-classic 0.6.52 + trading-signals 8.3.0 +
   shipped `TA_RVI`; ties→down = App::Chart alone, plus prose (Chart manual, DXcharts) that writes
   the denominator as *"EMA of SD"*.
2. **Wilder smoothing (`alpha = 1/n2`)**, not `2/(n2+1)` — inherited from TA-Lib/ta-lib#366; the 1993
   `((PREV*13)+x)/14` transcription plus Dorsey's own RSI sentence are the primary-source support.
   Cost of the flip is MEASURED at 12.94.
3. **`50.0` on a zero denominator** — inherited; house choice, no source states a guard, and
   App::Chart and trading-signals both land on 50 independently.
4. **`thirds` declined.** pandas ships it, nobody else does, and it is a different indicator, not a
   variant: MEASURED 33.33 at L=3 and 15.86 at L=5 against `refined=True` in pandas' own arithmetic.
   Not a parameter on this function; a separate card at most.
5. **Group `Volatility Indicators`; `flags: [stream]`; one unstable id, borrowed.**

## References

**[F]** = fetched this session, HTTP 200. **[C]** = cited, fetched for TA-Lib/ta-lib#366 rather than here.

- **[F] MetaStock-language reproduction of the 1995 article** — *"New Relative Volatility Index"* and
  *"Inertia"*, labelled *"Appeared in the September 95 issue of Stocks & Commodities magazine"*,
  quoted in full under **Formula**; the source that settles the tie rule. Read at
  `web.archive.org/web/20110803233955id_/http://trader.online.pl/MSZ/e-w-Inertia.html`
  (`trader.online.pl` is dead). **Provenance, stated exactly:** the page is a third-party MetaStock
  formula collection and credits `traders.com`; it does **not** carry an Equis byline, unlike the
  same site's 1993 page. App::Chart's header cites both, calling the 1993 one *"Reproducing
  equis.com original 1993 version"* and this one *"1995 revision"*.
- **[F] Equis/MetaStock, 1993 RVI** (same site, credited to `equis.com`) —
  `@RVI Down = ((PREV*13)+If(ROC(C,1,%)<0,Stdev(C,10),0))/14`, `@RVI Up` likewise with `>0`,
  `@RVI = (100*Fml("@RVI Up"))/(Fml("@RVI Up")+Fml("@RVI Down"))`. `((PREV*13)+x)/14` **is** the
  Wilder recursion. The same page quotes Dorsey verbatim from *TASC* V.11:6 (253-256): *"The RVI is
  simply the relative strength index (RSI) with the standard deviation over the past 10 days used in
  place of daily price change."* Read at
  `web.archive.org/web/20110514040046id_/http://trader.online.pl/MSZ/e-w-Relative_Volatility_Index_RVI.html`.
- **[F]** Dorsey, "Refining the Relative Volatility Index", *TASC* V.13:9 (Sep 1995) 388-391 —
  https://store.traders.com/-v13-c09-refinin-pdf.html. Paywalled at $3.95, **not purchased**;
  title, author, pagination and the opening two paragraphs are from the listing, and those
  paragraphs are about trend and inertia, not the arithmetic.
- **[F]** Chart manual (Kevin Ryde),
  https://user42.tuxfamily.org/chart/manual/Relative-Volatility-Index.html — *"In 1995 Dorsey
  revised his idea, to apply RVI to the daily highs and daily lows and average the result, thus the
  following, which is the RVI used in Chart"*, then `RVI = (RVIorig of highs + RVIorig of lows) / 2`;
  *"The stddev period of 10 days and the default EMA of 14 days are parameters"*; *"The EMA period
  follows J. Welles Wilder's reckoning … the same as an RSI."*
- **[F]** DXcharts, https://devexperts.com/dxcharts/kb/docs/relative-volatility-index-rvi —
  *"RVI = (RVIorig of highs + RVIorig of lows) / 2"*. **Internally inconsistent on the stddev period**
  (prose "9-period", legend "SD - 10-days'"); cite for structure only.
- **[F]** App::Chart 275 (CPAN `KRYDE/App-Chart-275`, GPLv3), read from `fastapi.metacpan.org`.
  `RVI.pm` `proc()` = `(high_proc(high//close) + low_proc(low//close)) / 2` over two independent
  `proc_original` closures — the refined form in code. `proc_original` is
  `num = EMA(v > prev ? stddev : 0)`, `den = EMA(stddev)` — an `EMA(S)` denominator **is** ties→down —
  and `rvi = (den == 0 ? 50 : 100*num/den)`. Its own header cites the 1995 store listing:
  *"TASC 1995, high/low combination, and introducing Inertia."* `EMA.pm` carries Wilder as
  `N = 2W−1` (alpha = 1/W exactly) but is weight-normalised from zero (`sum/weight`), a different
  transient from an SMA-seeded Wilder — converged tail only. `Stddev.pm` divides by `count`:
  **population**.
- **[C]** TradingView, thinkorswim, MotiveWave, eSignal, fmlabs, ProRealCode — fetched and quoted on
  card TA-Lib/ta-lib#366. TradingView's one structural sentence ("uses **high and low** price
  standard deviation") describes *this* form, not TA-Lib/ta-lib#366's.
- Read locally this session: pandas-ta-classic 0.6.52 `volatility/rvi.py` (`if refined: … return
  0.5 * (high_rvi + low_rvi), "r"`; `_rvi_compute` passes ONE `length` to both stdev and ma, so it
  cannot witness `n1 != n2`), `utils/_core.py` (ties→neither), `statistics/stdev.py` (ddof 0);
  trading-signals 8.3.0 `dist/volatility/RVI/RelativeVolatilityIndex.js` (WSMA on both legs,
  ties→neither, 50 on zero total, separate `interval`/`stddevInterval`, `stddevInterval < 2`
  rejected); `tulip_serve/vendor/tulipindicators/beta/rvi.c`.

## Sample input/output (board policy)

Input: `TA_SREF_{high,low}_daily_ref_0_PRIV`, 252 bars — **MEASURED** equal element for element to
`trading_signals_serve/corpus.json`. `optInTimePeriod = 14`, `optInStdDevPeriod = 10`, unstable 0;
`outBegIdx = 22`, `outNBElement = 230`. Left column: the proposed body's spelling (var.c shifted-sum
+ reseed, rma.c `wBeta`-first, `100.0*(U/T)` per leg, then `0.5*(a+b)`). Right: the composed
trading-signals arm, run. The hex column is the **oracle's** bits — those are what a golden test
freezes, not the left column's.

| bar | TA_RVIR (proposed spelling) | trading-signals 8.3.0 (2 legs, averaged) | oracle hex |
|---|---|---|---|
| 22 | 43.834741790037505 | 43.834741790037512 | `4045ead8d1a86866` |
| 23 | 40.773744308768570 | 40.773744308768570 | `4044630a0db2d046` |
| 24 | 37.279696373902450 | 37.279696373902446 | `4042a3cd173d5c43` |
| … | | | |
| 250 | 53.521391948420400 | 53.521391948420100 | `404ac2bcf8ab6e58` |
| 251 | 52.847645335992624 | 52.847645335991670 | `404a6c7fa4725878` |

**MEASURED over all 230 values: max abs 9.52e-13, max relative 1.80e-14; 42/230 bit-equal.** The
residual is TA-Lib/ta-lib#366's three house FP-spelling choices; it is ~4x that card's close-form
figure (2.42e-13) because averaging combines two legs' residuals.

## Oracles

| Oracle | Covers the refined form? | Arm status | Work to add the arm |
|---|---|---|---|
| **`inHigh == inLow` reduction** | **Yes, exactly** — both legs collapse to one, and `0.5*(a+a) == a` bit-exact (**MEASURED: 0/230 bit mismatches**) | **free, in-tree, no server** | none. RVIR on a zero-range series must be bit-identical to the shipped `TA_RVI` on it. Zero tolerance. |
| **Composite differential vs two `TA_RVI` calls** | **Yes, by construction** | **free, in-tree, no server** | none — `TA_RVI` shipped in `8c0fedbcd`. This is the primary gate. |
| **pandas-ta-classic 0.6.52** | **Yes, natively** — `refined=True` is `0.5*(RVI(high)+RVI(low))`; `mamode='rma'`, ties→neither; constrained to `n1 == n2`, so it cannot witness the shipped (14,10) default | **must be added** | one `compute_RVIR` + one `SPECS` row in `pandas_serve/server.py`, the `compute_KC` shape (2 price inputs), ~10 lines. `length ∈ {3,5}`, bars `≥ beg + 168`. |
| **trading-signals 8.3.0** | **Legs only** — ships the close-form RVI; the average is the harness's | **must be added** | ~9-line `ARMS` entry whose `ctor` builds two `RelativeVolatilityIndex` instances and returns `{update(b){ … 0.5*(h+l) }}`, `feed: 'hl'`. **No `bar()` widening** — `hl` already exists in `capture.mjs`. **The only arm that reaches the shipped (14,10) default.** |
| **1995 MetaStock formula reproduction** | **Yes** — a formula witness, not an executable arm | **not an arm** | The strongest evidence on this card; it is what settles the tie rule. |
| **App::Chart 275 (Perl, GPLv3)** | **Yes, natively** — but ties→**down** | **not worth an arm** | A fifth server, and a **negative control**: the flip it represents is MEASURED at 4.64. Value banked by reading the source. |
| **Tulip 0.9.2** | **No, structurally** | **DO NOT ADD** | `beta/rvi.c` takes `inputs[0]` only — one series, so it cannot express a high/low average, on top of TA-Lib/ta-lib#366's MEASURED 36.04 finding that it is a different indicator. |
| **ta4j 0.24.1** | **No** | impossible | Verified on **0.24.1** (already in `~/.m2`, so the pin bump is a pom edit, not a download): 406 indicator `.java` files in the sources jar; no Relative Volatility, no Relative Vigor. Same verdict as the 0.22.6 pin, on the newer jar. |
| **PyneCore / `pine_serve`** | n/a | n/a | No `rvi` in `pynecore/lib/ta.py`; RVI is not a Pine `ta.*` builtin. |

**pandas convergence — MEASURED**, `ta.rvi(close, high=high, low=low, length=L, mamode='rma',
refined=True)` vs two `TA_RVI` calls averaged, max abs deviation from the given offset onward
(pandas-ta-classic 0.6.52 / pandas 3.0.3 / numpy 2.5.1):

| length | skip 0 | skip 56 | skip 112 | skip 168 |
|---|---|---|---|---|
| 3 | 13.2 | 1.66e-9 | 1.41e-11 | **4.72e-12** |
| 5 | 7.12 | 3.35e-5 | 9.96e-11 | **6.96e-12** |
| 10 | 11.4 | 2.81e-2 | 6.97e-5 | 1.55e-7 |
| 14 | 9.33 | 7.56e-2 | 1.36e-3 | 2.70e-5 |

pandas' `overlap/rma.py` seeds from a one-sample mean over an input whose first `length-1` bars are
NaN, so the error decays as `((L−1)/L)^k` and at small `L` the arm converges onto a correctly
SMA-seeded Wilder. It witnesses the **averaging step**, the **high/low leg assignment**, the
smoother, the up/down gate and the tie rule, from another codebase and language. It does **not**
witness `n1 != n2`: only the composed trading-signals arm reaches the shipped default (14,10), and
there the average is the harness's arithmetic. That split is the honest state of the coverage.

## Verification plan

1. **Composite differential, bit-exact `memcmp`**, appended to
   `src/tools/ta_regtest/ta_test_func/test_composite2.c`. Reference: two `TA_RVI` calls, on `inHigh`
   and on `inLow`, **both at RVIR's own `startIdx`**, then `0.5*(a+b)` — bit-exact because the fused
   legs execute the identical operation sequence. Grid `(n1,n2)` over `{2,3,10,14,30} × {1,2,14,30}`.
   Every piece is a shipped function; the only extra arithmetic is one exact multiply by 0.5.
2. **`inHigh == inLow` identity, zero tolerance.** `TA_RVIR(h,h,…)` must `memcmp`-equal `TA_RVI(h,…)`
   over the whole range. MEASURED exact today; free, needs no server.
3. **External golden — trading-signals, composed.** Freeze the arm's hex-of-IEEE-bits at
   `(n2,n1) = (14,10)`, `(30,30)`, `(1,10)`, `(2,2)`, compared with
   `checkOracleValue(got, want, /*relTol*/ 1e-9, /*absTol*/ 1e-12)` (`ta_test_priv.h:294-319`).
   relTol is set by the MEASURED worst relative — 1.80e-14 / 6.09e-15 / 1.42e-16 / **3.65e-12**,
   worst at (2,2) — leaving 274x. absTol is the `ta_test_priv.h` abs-near-zero rule, not a number
   the data forces: at (1,10) `want` is exactly `0.0` on **88** of 243 bars, where relative error is
   undefined, and MEASURED both sides are bit-equal on every one of them. Size it from the (1,10)
   series max abs, **1.42e-14**, so absTol 1e-12 leaves 70x.
4. **Second oracle — pandas converged tail**, `L ∈ {3,5}`, bars `≥ beg + 168`,
   `checkOracleValue(relTol 1e-9, absTol 1e-9)` (MEASURED worst 4.72e-12 and 6.96e-12, ~150x
   headroom). Record that the same arm is 13.2 and 7.12 away at skip 0, so the offset cannot be
   mistaken for arbitrary.
5. **Tie rule — an exact, non-decaying, SIX-bar probe.** **MEASURED** tie census: high 128 up /
   120 down / **3 ties (bars 66, 88, 250)**; low 133 / 115 / **3 ties (bars 81, 176, 191)**; close
   123 / 127 / 1 (bar 101). At `optInTimePeriod = 1` each leg has no memory, so a tie reads a literal
   and the average takes a quarter-point value:

   | bar | ties→neither | ties→down | ties→up |
   |---|---|---|---|
   | 66, 81, 88, 176 | **75.0** | 50.0 | 100.0 |
   | 191, 250 | **25.0** | 0.0 | 50.0 |

   **MEASURED**, the whole 243-value series at (1,10) takes only `{0, 25, 50, 75, 100}` — counts
   `{0:88, 25:2, 50:47, 75:4, 100:102}`, nothing else — and only `{0, 50, 100}` under either flip:
   the *presence* of a 25 or a 75 decides the rule in one assert. Assert the six literals and the
   value set. At (14,10) the flip is far more visible than in the close form — MEASURED max **4.64**
   (ties→down, bar 88) and **2.24e-1 at the last bar**, against TA-Lib/ta-lib#366's 1.97 and 5.33e-5,
   because a tie sits at bar 250.
6. **Streaming-tier aliasing rejection** (TA-Lib/ta-lib#108, TA-Lib/ta-lib#130): the `_OpenAndFill`
   guard must reject non-distinct output buffers. This is the **rejection** case, and it is not the
   same thing as item 7.
7. **Batch in-place aliasing must be bit-identical** — the opposite of item 6. Template exists and
   takes the same `inPriceHL` bundle: `src/tools/ta_regtest/ta_test_func/test_donchian.c:430-460`
   ("each output over each input in turn must match"). Alias `outReal` over `inHigh`, then over
   `inLow`, **separately** — two cases, not one. Both legs must finish reading bar `i` before the
   store; at `n2 == 1` with unstable 0 the slack is exactly zero. ASan cannot see a violation.
8. **Hand-written edge cases** — flat input: `U + D == 0` in both legs, so exactly `50.0` on every
   bar, no NaN (TA-Lib/ta-lib#112). Plus TA-Lib/ta-lib#366's list, inherited.
9. **Free once the YAML lands:** `--codegen`, `--xlang-hash` / `server_verify` at zero tolerance,
   `--fuzz-064`, the TA-Lib/ta-lib#94 boundary sweep under ASan/UBSan, and the unstable sweep as
   `doRangeTestEx(…, TA_STABLE_CONVERGING, TA_FUNC_UNST_RVI, …)`.
10. **No NaN/Inf can escape.** Each leg is bounded to `[0, 100]` by its division-first combine; the
    average of two values in `[0, 100]` is in `[0, 100]`. The averaging adds no division and no
    overflow path.

## Open questions

1. **Name — one word, and it is the only thing outstanding.** Recommend **`TA_RVIR`**: it matches
   pandas' output naming (`RVIr_<length>`) and sorts adjacent to `rvi`. Alternative: `TA_RVIHL`.
   Swapping the pair so `TA_RVI` becomes this function is not available — `TA_RVI` shipped under the
   close form in `8c0fedbcd`. There is no evidence that decides between the two candidates; it is a
   naming call, and the C symbol is part of the contract, so the card stays a proposal until it is
   made. Nothing else about the formula, the contract, the oracles or the gates waits on it.

## Notes

- **Streaming is not a risk.** The shipped `rvi.yaml:4` carries `flags: [unstable_period, stream]`
  and `Core::rvi_open` exists at `ta_codegen/output/rust/library/src/ta_func/rvi.rs:1016` —
  stream-census passes on RVI's IR. If RVI streams, RVIR streams. Each leg is a `TA_RVI` stream; the
  ring is `n1` highs plus `n1` lows, and the up/down gate needs no extra slot.
- **Not `path_dependent`** (TA-Lib/ta-lib#127): both legs seed at `startIdx − lookbackTotal`, exactly
  like RSI.
- **ABI and sorting:** one new function; there is no C `TA_FuncID` — the ordinal ABI is the Rust
  crate's `FuncId` (`ta_codegen/output/rust/library/src/abstract_api.rs:44`), where a new variant
  renumbers every later discriminant. No new unstable id. `rvir` sorts between **`rvi` and `rvol`**
  (`rvol` landed as TA-Lib/ta-lib#370), so confirm `ta_ref_serve` still builds
  (TA-Lib/ta-lib#228 AC).
- **Follow-up, no card exists** (0 hits for "inertia" on the board): the same 1995 article also
  introduced **Inertia**, a 20-period smoothing of RVIR. The sources disagree on the smoother — the
  1995 page's prose says *"a Linear Regression indicator"* and the Chart manual says a 20-day LSQMA,
  while that same page's code says `Mov(Fml("New Relative Volatility Index"),20,SIMPLE)` — so an
  INERTIA card must arbitrate it. `TA_LINEARREG` (the endpoint MA) is already shipped.
- Gotchas: a **full** `cargo run -- generate` (a `--func=RVIR` filter skips `Core.java` and the
  servers); `scripts/build.py servers`, not `build`; register the new test `.c` in **both**
  `CMakeLists.txt` and `src/tools/ta_regtest/Makefile.am`; the CHANGELOG entry; and
  `ta_regtest --function=RVIR` matches the *group tag*, so confirm the filter ran tests instead of
  exiting 0 vacuously.

kevinlincg:
Closes #416.

Implements `TA_RVIR`, Dorsey's 1995 refinement of his own RVI. The proposal card was transferred into this repo as #416 while this PR was open, so the two claim comments on it are mine, carried over from the drafts repo.

`RVIR[i] = 0.5 * ( RVI(high)[i] + RVI(low)[i] )`, both legs at the same parameters.

## Two shape findings that changed the implementation

**The fused body the card specifies cannot carry `flags: [stream]`.** The card asks for one pass driving two copies of RVI's state. That body is expressible and I wrote it first, but it carries two window-start cursors and the streaming analyzer accepts exactly one:

```
YAML declares `streaming: true` but the function is not streamable at stage 1:
extrema automaton: expected exactly one window-start variable,
found ["trailingIdxH", "trailingIdxL"]
```

Streaming is an API-surface contract and fusion is an implementation preference, so I kept the former: composed, like KC, where each leg streams as its own sub-handle. The cost is one scratch buffer sized `endIdx-startIdx+1`.

**The IR cannot express a leg-indexed inner loop.** `Expr::ArrayAccess(String, Box<Expr>)` takes a *name* as its base, not an expression, and `VarType` has no pointer-array member, so `legs[j][i]` — the natural spelling of "do this twice" — has no representation. Worth knowing before anyone else reaches for it.

## What the tests can and cannot falsify

Discrimination was measured, not assumed: each mutation was applied to the generator input and regenerated.

| mutation | differential | composite | degenerate | tie | edges (ASan) |
|---|---|---|---|---|---|
| averaging dropped, high leg returned alone | **RED** | **RED** | green | **RED** | green |
| the two `TA_RVI` calls swapped | green | green | green | green | green |
| lookback raised by one | green | green | **RED** | green | green |
| every scratch allocation one element short | green | green | green | green | **RED** |

Three things that follow, all of them in the file's header comment:

- **The swap is green because the order genuinely does not matter.** `TA_RVI`'s write index trails its read index by the lookback and never overtakes it, so neither call order can corrupt an aliased input. I had written a comment claiming the order was what made aliasing safe; the mutation disproved it and the comment is now the corrected version. The aliasing leg therefore earns its keep against the scratch buffer's extent, not against the order.
- **The composite leg is structurally true against this body**, which calls `TA_RVI` twice, so it cannot fail for an arithmetic reason. It is there for the fused rewrite, which would have to reproduce the same equality. The arithmetic is carried by the differential leg, which rebuilds both legs from `TA_STDDEV` and `TA_RMA` — so a defect shared by `TA_RVI` and `TA_RVIR` cannot hide in it.
- **Degenerate and differential are the pair that covers the warm-up.** The differential leg anchors its reference at whatever `outBegIdx` the function reports, so it follows a lookback drift instead of catching it; the degenerate leg (`inHigh == inLow`, which must reduce to `TA_RVI` exactly) compares `outBegIdx` against an established function and is the only leg that sees it.

The scratch-extent case needs AddressSanitizer to be visible — a plain run of the one-element case is green either way. Control run: clean 0 errors → mutated `heap-buffer-overflow`, abort → restored 0 errors, generated file byte-identical.

## Verified locally

- `ta_regtest` — full run green; `--function=RVIR` exercises all seven legs with literal coverage counts (39757 / 4609 / 13707 / 243 / 106592 / 9218)
- `ta_codegen/generator` — full suite green. Two lists needed RVIR and got it with its reason: `stability_suite`'s inherited-instability set (`RVIR` inherits `TA_FUNC_UNST_RVI` through both legs) and `streaming_suite`'s composed-sub-call set (safe for KDJ's reason — the low leg is handed `outBegIdx`/`outNBElement` themselves and they are returned unmodified, so the final count *is* that callee's count)
- `ta_codegen/output/rust` — 703 tests green
- C library builds clean

## Not verified locally

**The Java and C# outputs were never compiled.** This machine has no JDK that accepts `release 17` and its .NET SDK (7.0.200) cannot target `net10.0`. What I did instead is a syntax-level check against KC, the composed function this one is shaped after: brace and paren balance after stripping comments and string literals, identical to KC's. That is not a compiler. If CI is red on either, it will be in the generated fragment and I will fix it there.

I do not think this blocks review — the same generator paths produce KC's Java and C# today and they are green in CI — but the claim "this compiles" is one I have not earned, so I am not making it.

## Naming

The card left the name open between `RVIR` and alternatives. I took `RVIR`, the name the card recommends. `rvir.md` and `rvi.md` now carry reciprocal notes about the collision, because some vendors reserve the bare name RVI for *this* revision and call the 1993 form RVIorig; this library ships the 1993 form as `RVI`.

`rvir.md` also records the measured distance to five variants published under this name, so the next person to receive a "your RVIR is wrong" report can check which one they are holding: a plain exponential smoother instead of Wilder's, up to 12.9 index points; averaging the prices and taking one index of the result, 13.6; adding the close as a third leg, 6.0; a 9-period deviation, 3.2; routing ties to the down bucket, 4.6. Against `RVI` of the closes at shared defaults, 18.1 — this is not a re-parameterisation.


kevinlincg:
CI is green on all seven checks, including the two I said I could not compile here: Generated Java and Generated C# both pass.

The one that failed was the one I *had* built locally. `Generated C compiles` came back red — not on a compiler error but on `Public C ABI matches ABI.manifest`: a new public function is an ABI addition, and I had not touched the manifest. Fixed in f9d793e: interfaces added and none removed or changed, so the generation goes `c+1 : 0 : a+1` → `3:0:2` by the rule beside it in `configure.ac`, and the soname stays `libta-lib.so.1`.

Twelve entry points added: `TA_RVIR`, `TA_S_RVIR`, `TA_RVIR_Lookback`, and the nine streaming ones — which is the concrete thing the composed shape bought. A fused body would have shipped three.

Worth recording that `check-abi` on this machine reports `EXPORTS UNCHECKED: no ELF shared library (Mach-O/PE not covered)`, so the export-table half of that gate only ever runs in CI. That is a second thing I could not have caught locally, on top of the two I declared.

kevinlincg:
ABI bookkeeping note: this PR and #417 both bump `TALIB_LIBRARY_VERSION` to `3:0:2` and both write `function-count 2463`. The version is right after both land — libtool's generation is per release, so two batches of additions inside unreleased 0.8.2 are one bump — but the manifest count is not: merged together the file lists 24 new entry points under a count of 12, and git takes the identical count edit from both sides without raising a conflict.

`check-abi` does catch it (exit 1, `2463 -> 2475`), so whichever of the two merges second goes red on `Public C ABI matches ABI.manifest` and needs `scripts/build.py check-abi --update`. Details and the locally-merged reproduction are on #417.

kevinlincg:

|---|---|---|---|

kevinlincg:
Rebased onto `86917b38e`. One real defect found and fixed in the process, and it
was mine, not the rebase's.

## The defect

`ta_regtest --codegen --function=RVIR` failed:

```
Fail: doRangeTestFixSize diff data for idx=0 (3.852493e+01,4.077374e+01)
Fail: doRangeTestFixSize (23,23,23,1,1)
Fail: doRangeTestFixSize refOutBeg,refOutNbElement (22,230)
Fail: Diff 5.8373 %
RANGE TEST FAILED (code=161)  (TA_RVIR is post-reference)
```

`UNSTABLE_MAP` had no row for RVIR, so `stability_class()` classified it
EPSILON and the range leg measured it against a ~1e-13 envelope.

This gate is not new — it went in with `52b984724` on 2026-09-01, before this
PR was opened. I had never run `--codegen` against my own function, so I did
not report this as unverified either. It is the `PVO` case the map's own
comment describes, one function later.

## Why RVIR belongs in the map

RVIR averages two RVI legs, one over the highs and one over the lows, and each
leg is the same converging RVI recurrence. Measured at bar 380 with the
defaults, varying only startIdx:

| startIdx | RVIR |
|---|---|
| 0 | 16.803547034309 |
| 40 | 48.093740728249 |
| 80 | 89.681109913933 |
| 120 | 31.063735351054 |

Whole multiples — the converging shape. A genuine EPSILON function moves by
~1e-13 across the same sweep.

One row covers both legs, unlike `KC` directly above it in the map. KC's two
legs carry *different* ids, which is why it needs a row each: sweeping one id
warms one leg and leaves the other cold. RVIR's two legs carry the *same* id,
so sweeping it warms both. The passing run is the evidence for that claim, not
the reasoning.

After the row, the C leg reports:

```
post-cutover range-stability verified: 2 of 2 (2 value-compared, 0 path-dependent: coherency only)
```

## Separately: the new ride-along does not reach RVIR

Not a request, just a measured fact in case it is useful for #425, since the
ride-along is new as of `86917b38e`.

Post-cutover functions return at the `refFuncList` gate in `test_codegen.c`
after the float leg and the range-stability leg, which is above the
`ride_read()` call site. So `rideResponses` stays 0 for them and no ride
verdict is ever recorded.

Probe calibrated both ways rather than read off an exit code:

| function | pre/post cutover | ride-along line |
|---|---|---|
| ATR, NATR | pre | `104 verdict(s) over 2 function(s), 2036 bar(s)` (C; Rust 86/1594, Java 90/1786, C# 90/1786) |
| RVI, RVIR | post | none — not one line, in any of the four backends |

Both runs exit 0, and `RIDE VACUOUS` cannot speak here either: `rideEligible`
is incremented inside the same `else` branch, so the ratio floor never sees
these functions at all.

I first used RVI as the pre-cutover control, which was wrong — the output says
`no frozen-reference baseline (post-cutover): RVI,RVIR` — and a run with only
that control would have read as "the ride-along prints for nobody".

## Verification

- `ta_regtest --function=RVIR` — green
- `ta_regtest --codegen --function=RVIR` — green, four servers, range leg 2 of 2
- Zero drift: `python3 scripts/build.py generate` leaves the tree clean
- Mutation control: the failure quoted above IS the without-the-row run

---

## The security gate now fails on all four of my PRs, for a generated constant

Reporting, not asking for a specific fix — the call on what to do with it is
yours.

After rebasing onto `86917b38e`, `Scan for secrets (Betterleaks)` fails on
#413, #417, #418 and #419. Same finding in all four, one leak each:

```
┌─generic-api-key──○
│ 57291 │         ulong key = REDACTED;
│   path .................. ta_codegen/output/csharp/tools/TaCodegenServe.cs
│   resource .............. git.patch_content
leaks found: 1
```

The redacted value is `0xcbf29ce484222325UL` — the FNV-1a 64-bit offset basis,
the published seed of the hash, not a credential. It comes from the ride-along
emitters that landed in `7782b0373`.

### Why it started now, and why only the PRs

`pr-security-gate.yml` triggers on `pull_request` only and scans the commits a
PR introduces, so pushes to dev never run it. The constant has been on dev
since `7782b0373` without the gate ever looking at it. A branch that adds a
function regenerates the C# server, its own copy of the constant lands in the
PR's patch, and the gate sees it there.

Four PRs, four different functions, same file and same finding — but they are
all my branches, so "every function-adding PR from now on" is an inference from
four samples, not something I have measured on anyone else's branch.

### One thing I could not explain

`TaCodegenServe.java` carries the identical constant 201 times as
`long key = 0xcbf29ce484222325L;`, and the C# file carries it 201 times as
`ulong key = 0xcbf29ce484222325UL;`. Only the C# one is flagged, and only once.
I could not reproduce the scan locally to find out why — the runner image is
`ghcr.io/betterleaks/betterleaks` and this box cannot pull public Docker
images — so I am leaving it as unexplained rather than guessing at an entropy
or length threshold. It matters for how wide any allowlist would need to be: if
the Java form is under a threshold rather than deliberately excluded, an
allowlist written for the C# spelling alone would be one generator change away
from not covering it.

C and Rust do not contain the constant at all.

### What I am not doing

Not touching `.github/workflows/`, and not changing the generator to spell the
constant differently — either one is a decision about your CI, and the second
would also churn dev's brand-new `ride_gen` output. The four branches are
otherwise green: the other six checks pass on all of them.

kevinlincg:
Following up on my own "could not explain" from the comment above: I can now,
and the answer changes what an allowlist would need to cover.

I had said the scan was not reproducible here because this box cannot pull
public Docker images. That was the wrong reason to stop — Betterleaks ships
plain release binaries, so I ran the same version the workflow pins (1.7.4,
`betterleaks_1.7.4_darwin_x64`) with the same invocation
(`git . -v --redact --log-opts=<base>..<head>`) and reproduced the CI finding
exactly: same file, same line 57291, one leak, exit 1.

## Why C# is flagged and Java is not

Not the file extension, not the length. It is the `generic-api-key` rule's
Shannon entropy floor.

Controlled runs, one variable at a time, each a separate file in a scratch repo:

| token | len | uppercase | entropy | flagged |
|---|---|---|---|---|
| `0xcbf29ce484222325L` — the Java spelling | 19 | 1 | 3.4264 | no |
| `0xcbf29ce484222325LL` | 20 | 2, repeated | 3.4414 | no |
| `0xcbf29ce48422232UL` | 19 | 2 | 3.4264 | no |
| `0xcbf29ce484222325cb` | 20 | 0 | 3.3037 | no |
| `0xcbf29ce484222325UL` — the C# spelling | 20 | 2 | **3.5414** | **yes** |

Extension is not the discriminator: the same `ulong ... UL` line saved as
`.java` is flagged, and the `long ... L` line saved as `.cs` is not. Length is
not it either: pure-hex literals from 14 to 24 characters are all clean.

Bracketing the floor with a dense sample of 20-character literals:

| entropy | flagged |
|---|---|
| 3.4414 | no |
| 3.4464 | no |
| 3.4842 | no |
| **3.5037** | **yes** |
| 3.5087, 3.5219, 3.5414, 3.5464 | yes |

So the floor sits in (3.4842, 3.5037] — gitleaks' standard 3.5 for this rule.

## What that means for the fix

The C# form clears the floor by 0.0414 and the Java form misses it by 0.0736.
Java is not excluded by anything; it is one character short of the same
finding. `0xcbf29ce484222325UL` only clears it because `U` and `L` are two
*distinct* characters — `LL` would not, as the second row shows.

So an allowlist written against the C# spelling alone is one generator change
away from not covering this. Whether that argues for a path-scoped allowlist on
the generated servers, a rule-scoped one, or leaving the gate to fail on
function-adding PRs is your call — I am not touching the workflow or the
generator.

C and Rust still do not contain the constant at all, so they are unaffected
either way.

kevinlincg:
reach the language servers.

## What is wired

Six legs: differential, composite, degenerate, tie, flat, and the non-aliased call of the aliasing leg.

reproduce.

## Measured, not assumed

```
SV FAIL [RVIR] (pipe 0, c): BITWISE mismatch vs in-process C
exit 22
```

and a bare `ta_regtest --function=RVIR` is unaffected either way —

## One thing worth flagging in the diff

There are two macros here, not one, and the difference is not cosmetic. The
differential and degenerate legs sweep the unstable period and restore it under
a `done:` label. Returning straight out of the macro there would leave
`TA_FUNC_UNST_RVI` and `TA_FUNC_UNST_RMA` set for every test that runs after
this file, so a failure would corrupt the rest of the run instead of reporting
itself. Those two legs use the `goto` form.

Related: `server_verify()` syncs every changed unstable-period id to the servers
before each request, so the swept legs compare like-for-like rather than C-at-N
against server-at-zero.

kevinlincg:
Correcting my own analysis above. The entropy floor is real but it is only half
the rule, and I presented it as the whole explanation.

The gate is green on all four PRs now. Not from an allowlist — there is none —
and not from a workflow change. `dde309a94` renamed the ride's dedup
accumulator from `key` to `hash`, for a reason that has nothing to do with this
gate, and that rename is what stopped the finding.

## The rule is a conjunction

Same value in every arm — `0xcbf29ce484222325UL`, entropy 3.5414, the one that
was being flagged — varying only the identifier:

| identifier | flagged |
|---|---|
| `key` | yes |
| `secret` | yes |
| `token` | yes |
| `hash` | no |
| `accumulator` | no |
| `seed` | no |
| `mixer` | no |

So `generic-api-key` needs BOTH a credential-shaped identifier AND a value over
the entropy floor. My bracketing runs held the identifier at `key` throughout
and only varied the value, which is why they measured the floor cleanly and
said nothing about the name. Reporting the floor as the reason was an
overreach: it was one necessary condition presented as the sufficient one.

## What that changes

The concern I raised — that an allowlist written against the C# spelling would
be one generator change away from not covering the Java one — was reasoned from
the wrong half. The identifier is what decides, and both backends now emit
`hash`, so neither is flagged and the 0.0736 entropy gap between the two
spellings never mattered.

What does still hold, and is the only part worth keeping: the gate scans
`base..head`, so a constant already on dev is invisible to it while the same
constant arriving in a PR's own patch is not. That is why this fired on four
function-adding PRs and never on dev.

Nothing needed from you — recording it because the earlier comment reads as a
complete diagnosis and it was not one.

mario4tier:
@kevinlincg Just so you know, I am holding to do changes to the ABI until ta-lib-python is released. This is why there is a delay in merging your recent new TA function PR. Ome way or another, I will unblock that within 1 week.

Details
---------
This is because the ta-lib-python dev branch (which hold the code ready for v0.8.1) is tested against the ta-lib dev branch (because convenient to prepare ta-lib-python BEFORE ta-lib C is released).

Now, we are a bit delayed on releasing ta-lib-python, so we can't make ABI change until python is released (would break the python PR intended for their release).

We will change the process for the next release (we will no longer block ta-lib Core if ta-lib-python release is delayed).

kevinlincg:
Understood, and no rush on my side. I will keep the four rebased onto dev and
green so the hold is the only thing gating them.

One thing worth having on hand for whenever it lifts, since it only shows up
when more than one of these lands. All four carry the same edit:

```
-TALIB_LIBRARY_VERSION=2:0:1
+TALIB_LIBRARY_VERSION=3:0:2
```

Each is a pure addition and each bumps from the same base, so the triple is
correct for whichever one merges first and already correct for the ones after
it — one release, one bump, as the rule next to it in configure.ac says. What
does not survive the same way is `function-count` in ABI.manifest: git merges
those cleanly to a value lower than the tree actually has, and `abi.py check`
is what catches it. Running the manifest regeneration once after the last of
them lands settles it. Detail is in #417.

Nothing needed now — noting it here so it is not a surprise at merge time.

mario4tier:
Landing this. One commit added on top:

- the Notes claimed the two legs are computed in one pass; they are two `RVI` calls, and that sentence reached the javadoc, the C# XML docs and the website page.
- `optInTimePeriod`'s hint said "both legs", which means the up/down pair in RVI and the two indices here.
- the tie leg's prose named the wrong bars as the quarter-points: they are the bars where one series is flat, not where the two disagree (measured on the 252-bar corpus: 5 quarter-points, 53 disagreement bars).

Green locally on the new head: `check-source-lists`, C build, `ta_regtest`, `regen-check`, `ta_regtest --codegen` (all four languages, 161 functions each), and the generator suite (962 tests).

mario4tier:
Correction to my note above: the numbers I quoted for the tie leg were measured against `ta_gDataHigh.c`/`ta_gDataLow.c`, but this test's corpus is `TA_SREF_{high,low}_daily_ref_0_PRIV` in `test_data.c`. On the corpus the leg actually uses there are **6** quarter-points (flat highs at bars 66, 88, 250; flat lows at 81, 176, 191) and **47** opposite-direction bars.

The fix itself stands unchanged: a quarter-point is a bar where exactly one of the two series is flat, never one where the two disagree.

Issue #243, opened 2026-08-22
STDDEV/BBANDS: a small non-zero standard deviation produces 0


mario4tier:
`TA_STDDEV` returns exactly `0` for a series whose standard deviation is small but plainly
non-zero. `TA_BBANDS` inherits it: the upper and lower bands land exactly on the middle
band, so the band width is `0` for the whole series.

The failure is a cliff, not a drift: right up to the boundary the result tracks a textbook
two-pass sigma to ~1e-13 relative, and then becomes exactly `0`.

## What happens

`optInTimePeriod = 5`, `TA_BBANDS` with `optInNbDevUp = optInNbDevDn = 2.0`, `TA_MAType_SMA`.
Last output value shown; "BBANDS width" is `upper - lower`.

Series scaled down (the price level falls along with the tick):

```
tick size    price level  TA_STDDEV               two-pass sigma          BBANDS width
1e-02        1.9700e+00   0.0162480768092728      0.0162480768092719      0.0649923
1e-03        1.9700e-01   0.00162480768092714     0.00162480768092719     0.00649923
1e-04        1.9700e-02   0.000162480768092725    0.00016248076809272     0.000649923
1e-05        1.9700e-03   1.6248076809271e-05     1.62480768092719e-05    6.49923e-05
1e-06        1.9700e-04   1.62480768092722e-06    1.6248076809272e-06     6.49923e-06
1e-07        1.9700e-05   1.62480768092714e-07    1.62480768092719e-07    6.49923e-07
1e-08        1.9700e-06   0                       1.62480768092719e-08    0
1e-09        1.9700e-07   0                       1.62480768092719e-09    0
```

The same series held at a $100 price level, so that only the spread shrinks:

```
tick size    price level  TA_STDDEV               two-pass sigma          BBANDS width
1e-02        1.0197e+02   0.0162480768092732      0.0162480768092736      0.0649923
1e-03        1.0020e+02   0.00162480768092817     0.0016248076809283      0.00649923
1e-04        1.0002e+02   0.000162480768092512    0.000162480768092515    0.000649923
1e-05        1.0000e+02   1.62480768077832e-05    1.62480768077822e-05    6.49923e-05
1e-06        1.0000e+02   1.62480768102316e-06    1.62480768102311e-06    6.49923e-06
1e-07        1.0000e+02   1.62480769291796e-07    1.62480769291791e-07    6.49923e-07
1e-08        1.0000e+02   0                       1.62480774539503e-08    0
[...]

PR #417, opened 2026-09-14
feat(kurtosis): add TA_KURTOSIS, rolling excess kurtosis (G2)

kevinlincg:
Implements `TA_KURTOSIS` from the proposal card in [ta-lib-proposal-drafts#72](https://github.com/TA-Lib/ta-lib/issues/433), which I claimed there before starting. That card has no open questions — G2 is selected, the degenerate window is ruled NaN, the parameter floor of 4 is fixed by the `(n−2)(n−3)` denominators — so this PR takes all of it as written and settles only the one thing the card left to implementation: numerical-design item 4.

## Item 4, measured

The card asks whether `S1..S4` can be carried incrementally or whether the window must be rescanned, and is explicit that it be settled by measurement rather than by analogy to `var.c`. Reference is `decimal` at 60 digits, computed per bar for that bar's own window. Two calibrations before trusting it: it agrees with `scipy.stats.kurtosis(bias=False)` to 1.1e-16 on a benign window, and it *disagrees* with scipy on the card's own `[5.0]*7 + [5.0+1e-13]` case, where the exact G2 is **8** and scipy returns 8.0635677113437438.

1200-bar series, `n = 30`, worst relative error over every emitted bar:

| series | 32n (`var.c`'s) | 8n | 2n | n | **n/4** | rescan every bar |
|---|---|---|---|---|---|---|
| walk around 100 | 1.18e-06 | 9.33e-07 | 1.61e-08 | 1.28e-09 | **4.58e-12** | 5.34e-11 |
| walk on 3.1e10 | 4.99e-06 | 4.99e-06 | 6.27e-10 | 1.45e-09 | **1.76e-12** | 4.30e-11 |
| outlier 1e5 every 200 | 2.84e-13 | 2.84e-13 | 2.68e-13 | 1.51e-13 | **4.48e-14** | 2.21e-12 |

**So incremental works, but `var.c`'s period does not transfer.** A fourth moment recovered against a stale shift pays `(u/σ)⁴` where a second pays `(u/σ)²` — the shift goes stale four times faster in the exponent — and at `32n` that leaves 1e-6 on an ordinary random walk. The collapse trigger cannot catch it: the ratio it tests sits around 0.24 there, six orders from firing, so only the periodic rebuild can, and it has to run at `n/4`.

Rebuilding that often also beats rescanning the window every bar, which I did not expect. The rebuild anchors the shift at the window *mean*; a per-bar rescan can only anchor it at a window *value*, which sits further from centre and leaves a larger residue. Better **and** about a tenth of the arithmetic.

**This is the one place the PR departs from the card**, which says to reuse `var.c`'s reseed architecture wholesale. The architecture is reused; the period is not. I raised it on the card before implementing and had no reply, so I went with the measurement — say the word if you would rather it match `var.c` for consistency and I will change the constant and re-baseline.

Not measured: throughput. The flop counts are static (~25/bar against ~120 for the rescan); I have not run either shape through `ta_bench`, so I am not quoting that as a timing.

## What each test leg can falsify

Measured by mutating the generator input and regenerating, not assumed:

| mutation | progression | scale | goldens | near-degenerate | NaN |
|---|---|---|---|---|---|
| `coefA` denominator (n−1) → n | **RED** (144) | green | green | green | green |
| residue correction dropped | **RED** (128) | green | green | green | green |
| rebuild period n/4 → 32n | **RED** (36) | green | green | green | green |
| degenerate window returns 0 | green | green | green | green | **RED** (571) |
| absolute 1e-14 floor on `moment2` | green | **RED** at 2^-40 | green | **RED** | green |

Three notes on that table:

**The progression identity is the strongest leg and consults nothing.** Substituting the discrete-uniform `g2 = −6(n²+1)/(5(n²−1))` into `G2 = [(n−1)/((n−2)(n−3))]·[(n+1)g2 + 6]` leaves `−6(n−2)(n−3)` over `5(n−2)(n−3)`: the n-dependence cancels exactly, so **every** arithmetic progression has excess kurtosis −6/5 at every length and every spacing. That one assertion constrains both coefficients against each other *and* the shift's staleness, which is why three separate defects land on it. 81936 comparisons, no oracle.

**Its premise is checked per series, because it does not always hold.** At base 3.1e10 the ulp is 3.8e-6, so a requested spacing of 1e-4 lands on two different actual differences (9.918e-05 and 1.030e-04) and the series is not a progression at all — asserting −1.2 there would test the spacing's representability, not this function. 36 such combinations are skipped and counted rather than silently expected to hold. (I found this by asserting −1.2 on them first and reading the 9 failures as an implementation bug.)

**The scale-invariance leg's exponent range is load-bearing.** It exists to catch an absolute threshold anywhere in the body — the #243 mistake, where a fixed epsilon is a cliff at a price level rather than a noise floor. At 2^±8 a corpus around 100 still has a second moment near 1e-1, so a 1e-14 floor is nowhere near it and the leg passes straight over the defect. MEASURED: with such a floor inserted, 2^-8 is green and 2^-40 is red. The span has to reach the floor to see it.

**The near-degenerate golden is non-vacuous by a wide margin.** Every value is `base + k·ulp` at base 2^20, so the corpus is exact and the reference is not measuring its own error. On the first window the 60-digit reference says 0.34999999999999998; **scipy 1.15.3 answers 1.6624999999999996 — a 375% miss — and raises `RuntimeWarning: Precision loss occurred`**. This implementation lands 1.59e-15 from the reference.

## Correction to something I said on the card

I wrote there that scipy 1.15.3 "does not raise the `RuntimeWarning` the card quotes from 1.18.1", and drew the conclusion that anyone reproducing item 3 on an older scipy would see the garbage number with nothing flagging it. **The first half is true only of that particular case and the conclusion is wrong.** 1.15.3 has the warning and raises it on the near-degenerate corpus above; it stays silent on the card's `[5.0]*7 + [5.0+1e-13]`. The trigger is case-dependent, not version-dependent, so the warning is not something to lean on as a detector either way.

## Verified locally

- `ta_regtest` — full run green; `--function=KURTOSIS` exercises all seven legs with literal coverage counts (81936 progression comparisons / 36 skipped / 52 golden / 13 near-degenerate / 571 NaN)
- `ta_codegen/generator` — full suite green. One inventory needed updating and got it: `FUSING_INVENTORY`, because the `moment4` expansion is `a*b + c` and the Rust backend emits an FMA dispatch trio for it
- ABI: twelve added entry points, nothing removed or changed, so `c+1:0:a+1` → `3:0:2` and the soname stays `libta-lib.so.1`
- CHANGELOG entry added under 0.8.2

## Not verified locally

**C#.** This box's .NET SDK (7.0.200) cannot target `net10.0`, so that backend is only exercised by CI. The same generator paths produce the other three and they are green here, but "this compiles" is a claim I have not earned for C#.

`check-abi` also reports `EXPORTS UNCHECKED: no ELF shared library (Mach-O/PE not covered)` on macOS, so the export-table half of that gate runs only in CI.


kevinlincg:
Heads-up on something neither PR's own CI can see: **#413 and #417 collide in the ABI bookkeeping**, and GitHub reports both as mergeable because each is mergeable against `dev` *separately*.

I merged both into `dev` locally to find out exactly what happens.

**`configure.ac` conflicts, but only in the comment.** Both PRs bump `TALIB_LIBRARY_VERSION` from `2:0:1` to `3:0:2`, and that value stays correct after both land — libtool's generation is per release, not per change, so two batches of additions inside unreleased 0.8.2 are still one `c+1:0:a+1`. Only the line naming which function was added needs merging:

```
# 3:0:2 -- TA_RVIR and TA_KURTOSIS and their streaming entry points are added
# and nothing is removed or changed, so the soname stays libta-lib.so.1.
```

**`ABI.manifest` does NOT conflict, and that is the part worth knowing.** Each PR writes `function-count 2451 -> 2463`. Git sees the same edit on both sides, takes it once without complaint, and merges the two disjoint blocks of added `function` lines cleanly — leaving a manifest that lists **24** new entry points under a count that says **12**. No conflict marker, nothing to review.

**The gate does catch it.** `scripts/build.py check-abi` exits 1 on the merged tree and names the number:

```
-function-count 2463
+function-count 2475
```

So whichever of the two is merged second will go red on `Generated C compiles` at the `Public C ABI matches ABI.manifest` step, and the fix is one command — `scripts/build.py check-abi --update` and commit the manifest. I am flagging it only so that red reads as expected rather than as a defect in the PR that tripped it.

Four generated files also conflict textually and need nothing but a regenerate: `src/ta_abstract/ta_func_api.c`, `BuildStamp.java`, `metadata/FunctionDescription.java`, `tools/TaCodegenServe.java`.

If you would rather not deal with it at merge time, I can rebase whichever one you plan to take second onto the other and carry the combined manifest in that PR — say which and I will push it.

kevinlincg:

|---|---|---|---|

kevinlincg:

|---|---|---|---|

kevinlincg:
**The pandas arm from the card, run.** Not via `ta_pandas_serve` — `ta-lib-oracles` is not a repo I can reach — but directly against pandas 2.3.3 on this box, same corpus as the frozen goldens.

pandas' identity was checked before trusting it, since a pandas computing the biased `g2` would disagree with everything and look like our defect: `.kurt()` matches `scipy.stats.kurtosis(bias=False)` to 1e-12 on the corpus head and is nowhere near `bias=True` (1.943 against 0.762). So it is G2, as the card states.

**Result: 52 values across periods 10 and 20, worst relative difference 4.767e-12**, at period 10 bar 33.

That figure is pandas', not this function's. Same window, against a 60-digit decimal reference:

| | value | relative error vs reference |
|---|---|---|
| pandas `.rolling(10).kurt()` | −0.3936640054375812 | **4.76e-12** |
| pandas `.kurt()` on the window alone | −0.39366400543544344 | 6.66e-13 |
| scipy `bias=False` | −0.39366400543544344 | 6.66e-13 |
| this implementation | — | ~1e-14 (the 1.14e-14 already reported for this corpus) |

pandas' rolling accumulator is seven times further from the reference than pandas' own non-rolling path on the same numbers, and this implementation is about two orders closer than either. The window is bars 24..33, which straddles the flat run — the region the reseed exists for.

So the arm agrees to what pandas can resolve, and where it does not, the reference says which side is drifting. I am not proposing a golden from it.

kevinlincg:
Rebased onto `86917b38e`.

KURTOSIS passes it:

```
no frozen-reference baseline (post-cutover): KURTOSIS  [sweep skipped 1; all still bitwise-gated by VARIANT]
```

kevinlincg:
reach the language servers.

## What is wired

Six legs: progression, scale, goldens, degenerate, the non-aliased call of the aliasing leg, and the contract leg's accepted periods.

reproduce.

## Measured, not assumed

```
SV FAIL [KURTOSIS] (pipe 0, c): BITWISE mismatch vs in-process C
exit 22
```

and a bare `ta_regtest --function=KURTOSIS` is unaffected either way —

## One leg deliberately left out

The period<4 leg asserts `TA_BAD_PARAM`. `server_verify()` skips reject cases
by design — parameter validation is implementation-specific across languages —
so wrapping that leg would make the vacuity floor fire on a leg doing exactly
what it is meant to do. Left unwrapped on purpose rather than overlooked.

kevinlincg:
Trimmed `kurtosis.md` to match the standard you set in efeda0bf3 and d4586ca78 on RVI/RVIR, rather than leave you to apply it here a third time. Pushed as de7b8d2d7.

Removed two Notes bullets: the reseed-period measurement table, and the "rebuilding beats rescanning" comparison. Both were duplicates — the table and its derivation live in `kurtosis.c:66-87`, which is where a reader who needs them is already looking. Nothing was deleted that exists only in the `.md`.

The degenerate-window behaviour moved into the Formula block, following the `or 50 when Up + Down = 0` line you added to `rvi.md`. The Summary lost its paragraph about `SKEW` being a separate question; that was a record of a decision, not something a caller needs. It is the only one of the three whose Summary changed, so it is the only one whose Rust doc comment changed.

What I kept is the estimator identity (`G2` vs the biased `g2`, with the one number that shows they are far apart), the NaN rationale in one line, and the unboundedness.

kevinlincg:
Traced both defects from 69d68b702 into KURTOSIS, since its rebuild carries the same `moment2 < 1e-6 * total2` shape with the current running sum as the denominator. The short answer is that neither transfers, and the peak-sum fix would be a no-op here. Measurements rather than reading, because the shape is the same and reading it says otherwise.

**The flat-window re-fire does not happen.** Timing `TA_KURTOSIS` over 20000 bars, ordinary series against exactly-flat and flat-to-one-ulp, at periods 10/100/1000: all three flat at 1x. The probe does detect the regime — forcing the rebuild on every bar gives 88x at period 1000 on the same three series. On an exactly flat window `moment2` and `total2` are both zero, and `0 < 1e-6 * 0` is false, so the clause cannot re-fire.

**The stale-shift defect is present but bounded, and your fix does not move it.** Worst relative error per emitted bar against a 60-digit `decimal` reference, 1200 bars, period 30:

| series | as merged | with the peak-sum trigger | collapse trigger disabled |
|---|---|---|---|
| random walk | 9.92e-13 | 9.92e-13 | 9.92e-13 |
| decays back onto the anchor | 9.67e-09 | 9.67e-09 | 9.67e-09 |
| 1e5 spike passing through | 1.16e-13 | 1.16e-13 | 1.05e+13 |

The decay shape is four orders worse than an ordinary walk, so the defect is real. But it is `optInTimePeriod/4` of stale shift, not the collapse trigger: switching to the largest `total2` since the last rebuild changes nothing to the last digit on any shape. The third column is the control that says the clause is live rather than inert — disabling it leaves the walk and the decay untouched and destroys the spike case, which is the job it is actually doing here.

So KURTOSIS's exposure is the `n/4` rebuild period, which is 128 times shorter than the `32n` that let CTI reach 0.91. I have not changed anything; if you would rather the decay shape were tighter than 1e-8 the lever is the period, not the trigger, and that is a number worth you setting rather than me.

The reference is calibrated both ways before it scores anything: it agrees with `scipy.stats.kurtosis(bias=False)` to 1e-15 on a benign window, and returns exactly 8 on `[5.0]*7 + [5.0+1e-13]` where scipy returns 8.0635677113437438.

MEDIAN carries no running sums at all, so neither defect has anything to attach to.

On the two numbers you corrected in #419: 2.864e-01 appears nowhere else — not in #417, #418, #430, #432, #433, and not in the KURTOSIS or MEDIAN sources. The probe that says so found all three occurrences in #419, so it is not reporting an empty search.

`internal_error_ids.yaml`: understood on the duplicate 429. I will regenerate it rather than hand-edit the `next:` line when #419 lands and I rebase these two onto it.

mario4tier:
Thanks Kevin. Three commits on top of your head, then this lands.

**`fix(kurtosis)`: rebuild against the peak moments.** var.c's triggers watch the second moment against the current sums, and a fourth moment loses its digits long before either fires. Measured on your head:

- Period 100, a walk with a 1e5 outlier every 200 bars: 115 bars wrong in total, up to -239 where exact G2 is -0.41. The departure trigger fires while the outlier is still in the window, so the rebuild anchors on a mean that includes it.
- A series decaying back onto the shift: 1.1e-3 from exact at period 20.
- A flat window whose mean does not round back to its level: period 49 at 100, 0.1 or 100.37 returns -3.197 instead of NaN.
- Volatility regimes and near-flat windows at long periods: up to 1e-4.

Each moment now triggers a rebuild when it falls below 1% of the largest shifted sum of its power held since the last rebuild. When the mean-anchored rebuild still fails either test, the shift moves by the residue it just measured, which lands within about an ulp of the mean, so flat windows are exact 0/0.

Your reseed table was right for the old triggers. With a stale shift caught by the triggers, n/4 no longer buys accuracy that matters, so the period is var.c's 32n as the card asked: worst 8.7e-12 against exact rational G2 over twelve corpora (1.1e-12 at n/4), about a tenth of the rebuilds, and about 2x the batch throughput. The periodic rebuild is also what clears a NaN or Inf input, within 32n bars as in VAR.

**`docs(kurtosis)`.** Implementation and Aliases sections. The Summary defined the window's own g2 and read it as a sharper or flatter peak; it now defines G2 as an estimate for the distribution, a tail-weight measure. "Not bounded" is now "at most n", and the estimator note states the G2/g2 identity.

**`test(kurtosis)`.** 42 goldens on the 252-bar close at periods 4/10/30/100: exact rational G2 at 2e-12, scipy 1.18.1 and pandas-ta-classic 0.6.52 at 1e-11, plus Microsoft's KURT example. pandas `rolling().kurt()` is checked but not frozen: it sits 1e-5 from exact at period 4. Five legs hold every window to a no-history recompute and flat windows to NaN, and your original kernel fails each of them alone; two more pin non-finite input recovery and the second-moment trigger. Leg 1 gains a dyadic 2^-13 spacing and a 1e-12 tolerance (measured 1.1e-15). Your 60-digit goldens equal exact rational G2 bit for bit.

C# ran here (.NET 10): all seven suites pass, and KURTOSIS is bit-identical across C, Rust, Java and C# under `--xlang-hash`.

kevinlincg:
My last comment was wrong, and wrong in the direction that matters: I reported that neither defect transferred and that the peak-sum trigger was a no-op on KURTOSIS. Thank you for not taking it.

**Reproduced your flat-window case on my own head before accepting it.** At 3e4ac0faf, every period-49 output on a wholly flat series is -3.19704 instead of NaN, 341 bars of it, at all three of 100, 0.1 and 100.37. Periods 8, 16, 32, 64 and 100 are clean.

**Two separate errors on my side, and the second is the one worth recording.**

The structural one: I applied the peak idea only to the second-moment trigger. Each moment needs its own largest shifted sum, and the fourth is the one that loses its digits first — which is the whole reason a fourth moment cannot inherit `var.c`'s triggers, and I had already written that sentence in the source comment without following it to the trigger.

The measurement one: I sampled four periods, 8/16/32/64, and ran the accuracy table at a single period. The failure is sparse along the period axis, so no coarse sample was going to land on it. I do not have a characterisation to offer — the obvious model, that the flat window's mean does not round back to its level, explains period 49 but over-predicts: it flags 0.1 at period 9 and 100.37 at period 11, and both measure clean on my head. So the honest statement is only that sampling periods cannot find this and the property has to be asserted densely, which is what your leg does.

My "9.67e-09 on a decay" number stands as a measurement of the shape I fed it, but reporting the three columns as if they settled the question was the over-extension. A table of four shapes at one period is not a negative result about a function.

**Verified your merge.** At a57b779a1 all six of those cases return NaN.

**On MEDIAN (#418), the claim I made alongside it does hold, and I checked rather than reasoned.** `median.c` has no accumulators at all — no running sums, no shift, no reseed; the window is a circular buffer plus a by-value copy, and every output is a read of one or two slots. There is nothing for either trigger to attach to.

PR #418, opened 2026-09-14
feat(median): add TA_MEDIAN, the rolling median

kevinlincg:
Implements `TA_MEDIAN` from the proposal card in [ta-lib-proposal-drafts#73](https://github.com/TA-Lib/ta-lib/issues/432), which I claimed there before starting. That card has no open questions — the even-`n` mean is unanimous across NumPy, R, scipy and Excel, there is no original author to arbitrate against, and the fold-into-`PERCENTILE` alternative is closed by that function's own arity ruling.

## The implementation is `percentile.c` with a different read

The machinery is borrowed unchanged: the window carried twice, `ring` by age and `sorted` by value, insertion-sort placement on the way in, shift-down removal on the way out, both hand-written `CIRCBUF`s as the card requires.

What changes is what comes out. At the output store `sorted` holds the window's *other* `n−1` values and `pos` is where the incoming one belongs, so the k-th of the virtual full window is `sorted[k]`, `newValue`, or `sorted[k−1]` depending which side of `pos` it falls on — no materialisation. Odd `n` is one such read at `(n−1)/2`; even `n` is two, at `n/2−1` and `n/2`, averaged. The buffer stores stay below the output store exactly as they do in `percentile.c`, which is what keeps the streaming peek frame able to drop the state update.

**The odd case is a branch, not `(v + v) / 2.0`.** The arithmetic form is exact for any value this library is realistically handed — doubling moves the exponent with the mantissa untouched, halving moves it back — but it overflows to ±inf above `DBL_MAX/2`, and this function does not declare `nan_inf_output`. The condition is loop-invariant, so the branch costs nothing.

## The strongest gate is in-tree and free

At odd `n` the median **is** the nearest-rank 50th percentile: `ceil(50·n/100)` is the central ordinal, and both functions read the same order statistic out of the same sorted window. So this is asserted against shipped `TA_PERCENTILE(x, n, 50)` with `memcmp`, not a tolerance. No oracle server, no capture.

**That leg alone is satisfied by a body that just forwards to `PERCENTILE`** — which is the one thing this function exists not to be. So the even-`n` leg asserts the two must *differ*, and it is mutation-validated:

| mutation | odd identity | even differs | goldens | invented level | exact shapes |
|---|---|---|---|---|---|
| even-`n` average → lower central value (i.e. the body becomes `PERCENTILE`) | green | **RED** — 0 of 7600 differ, against 7322 for the real body | green | green | **RED** |

The failure message names both possibilities rather than just the one, because a corpus with no window whose two central values differ would make leg 1 vacuous too.

**One mutation deliberately did not go red, and I chased it down rather than leaving it as a weak gate.** Flipping the placement scan's `<=` to `<` inserts a new value at the *front* of a run of equal values instead of the back — a real change to internal state. It changes no output at all: 14820 windows on both a 7-distinct-value series and a random walk, bit-identical. Equal members of a window are interchangeable, which is precisely the premise `percentile.c`'s removal relies on when it identifies a value by value and carries no slot array. That measurement is now recorded at the site so the next person does not read the green as a hole.

Other legs: frozen numpy goldens at periods 7, 8 and 20 (asserted bitwise — the definition has no variant to pick); exact assertions on monotone and flat windows at both parities; in-place aliasing; and one that asserts the documented consequence rather than describing it — **at even `n` the output reaches levels the series never traded at** (13 such bars in the 8-period golden corpus), which is the deliberate opposite of `PERCENTILE`'s design property.

## Also in this PR

The card's "Also owed" item: `percentile.md`'s alias list claimed **"Rolling Median" unqualified**, and the generated `website/src/functions/percentile.md` carried it live. True only for an odd window. Removed, with a CHANGELOG entry under Fixed. Worth landing whether or not MEDIAN itself ships.

## Verified locally

- `ta_regtest` — full run green; `--function=MEDIAN` exercises all seven legs with coverage counters (4620 odd-identity comparisons, 4640 even with 4396 differing, 88 golden, 13 invented levels, 29640 exact-shape, 3029 alias)
- `ta_codegen/generator` — full suite green, no inventory needed updating
- ABI: twelve added entry points, nothing removed or changed, so `c+1:0:a+1` → `3:0:2` and the soname stays `libta-lib.so.1`

## Not verified locally

**C#** — this box's .NET SDK (7.0.200) cannot target `net10.0`, so that backend is only exercised by CI.

`check-abi` also reports `EXPORTS UNCHECKED: no ELF shared library (Mach-O/PE not covered)` on macOS, so the export-table half of that gate runs only in CI.

## Note on the ABI line, shared with #413 and #417

All three of my open PRs bump `TALIB_LIBRARY_VERSION` to `3:0:2` and write `function-count 2463`. The version stays correct however many of them land — libtool's generation is per release, so several batches of additions inside unreleased 0.8.2 are one bump — but the count does not: git takes the identical count edit from each side without raising a conflict, leaving a manifest that lists more entry points than it claims. `check-abi` catches it (exit 1, naming the right number), so each PR after the first goes red on `Public C ABI matches ABI.manifest` and needs one `scripts/build.py check-abi --update`. Flagging it so that red reads as expected. Details on #417.


kevinlincg:

|---|---|---|---|

kevinlincg:

|---|---|---|---|

kevinlincg:
**The pandas arm from the card, run.** Not via `ta_pandas_serve` — `ta-lib-oracles` is not a repo I can reach — but directly against pandas 2.3.3 on this box, same corpus as the frozen numpy goldens.

pandas' convention was checked before trusting it: `.rolling(4).median()` on `[1,2,3,4]` gives `2.5`, the mean of the two central values, which is the convention this function implements and the one the card calls unanimous.

**Result: 88 values across periods 7, 8 and 20 — all 88 bit-identical.** Worst relative difference exactly zero.

That is a second independent implementation agreeing bitwise, on a corpus whose 8-bar windows include 13 bars where the output is a level the series never traded at. numpy was the first; pandas reaches the same doubles by a different path (an incremental rolling accumulator rather than a per-window sort), which is the part worth having.

kevinlincg:
Rebased onto `86917b38e`.

MEDIAN passes it:

```
no frozen-reference baseline (post-cutover): MEDIAN  [sweep skipped 1; all still bitwise-gated by VARIANT]
```

kevinlincg:
reach the language servers.

## What is wired

Seven legs: odd, even, goldens, invented, monotone, flat, and the non-aliased call of the aliasing leg.

reproduce.

## Measured, not assumed

```
SV FAIL [MEDIAN] (pipe 0, c): BITWISE mismatch vs in-process C
exit 22
```

and a bare `ta_regtest --function=MEDIAN` is unaffected either way —

What this buys specifically: a port that took the lower of the two central
members instead of their mean would pass every odd-window assertion in this
file untouched, and disagree with C on every even one. That is the defect the
odd/even split exists to catch, and until now it was only ever checked in one
language.

kevinlincg:
Trimmed `median.md` to match the standard you set in efeda0bf3 and d4586ca78 on RVI/RVIR. Pushed as 115866501.

Removed three Notes bullets. The "even-`n` mean is not a variant to choose" bullet is the same shape as the population-vs-sample standard deviation bullet you cut from `rvi.md` — it argues against a variant nobody ships. The odd-case branch and overflow reasoning, and the insertion-order note, are internals: both already live in `median.c:112-134`, and tie order among equal values is not observable in the output at all.

What I kept is the `PERCENTILE` contrast with the concrete 4-bar example, the consequence that an even period can emit a value the series never traded at, and the note that the period is not restricted to odd values.

kevinlincg:
Rebased onto f9a708d92. One thing from that commit is worth a question rather than a change.

You capped KURTOSIS at 10000 because the running sums drift past 1e-9 above ~3e4. MEDIAN declares `[2, 100000]` and that reason does not transfer — it has no running sums, and the output is a slot read, so it stays exact. What the top of its range does cost is time, because the by-value copy shifts O(n) entries per bar:

| period | us/bar | value check |
|---|---|---|
| 100 | 0.11 | 6/6 exact |
| 1000 | 0.77 | 6/6 exact |
| 10000 | 6.92 | 6/6 exact |
| 100000 | 93.51 | 6/6 exact |

150000 bars at period 100000 is 4.7s. Each row is checked against an independent median taken by sorting a copy of that bar's window, so the exactness is not the function checking itself.

So the question is whether MEDIAN's top should follow KURTOSIS's to 10000 for consistency of what the library promises to be usable at, or stay at 100000 because it is merely slow rather than wrong. I have not touched it — it is your call which of those the range is meant to express.

Also in this push: the MEDIAN test group's `DO_TEST` line and its declaration, both of which I had dropped resolving an earlier rebase. The suite stayed green the whole time because it simply was not running those legs — `--function=MEDIAN` said it matched no group and exited 0. Restored, and the group runs.

kevinlincg:
404501f0c answers the question I asked, so I applied it rather than wait: MEDIAN's period top is now 10000, pushed as 6deff1cc6.

Your rule there is cost proportional to the period on every bar, and MEDIAN is the same machinery as PERCENTILE — it shifts a sorted copy of its window. Measured on this box, 6.9us per bar at 10000 against 93.5us at 100000, the same shape as your 3 and 33. The values stay exact at both, so this is a cost cap, not a correctness one.

A new leg pins 10000 as accepted and 10001 as rejected, in the lookback and in the call, following the shape of your PERCENTILE edges. It is mutation-checked rather than merely green: reverting the range to 100000 fails it with `lookback 10000 at period 10001, expected -1`, exit 120.

No CHANGELOG entry, since MEDIAN has not shipped — that is the distinction your PERCENTILE commit drew, and it puts MEDIAN on the KURTOSIS side of it.

If you would rather MEDIAN keep 100000 because being merely slow is not the same as being wrong, this is one line in the yaml plus the leg.

Issue #242, opened 2026-08-22
CORREL: perfectly correlated inputs produce 0 instead of 1


mario4tier:
`TA_CORREL` returns exactly `0` for two series that are perfectly correlated, once the
spread of the input window gets small enough. The correct answer is `1`.

The failure is a cliff, not a drift: right up to the boundary the result tracks a textbook
two-pass Pearson r to ~1e-12 relative, and then becomes exactly `0`.

## What happens

`y = 2x + c` — perfectly correlated by construction, so `r == 1` for every row below.
`optInTimePeriod = 30`, last output value shown.

Series scaled down (the price level falls along with the tick):

```
tick size    price level  sigma(x)     TA_CORREL               two-pass Pearson r
1e-02        1.9700e+00   8.4772e-02   0.999999999999967       1
1e-03        1.9700e-01   8.4772e-03   0.999999999999743       1
1e-04        1.9700e-02   8.4772e-04   0.999999999999437       1
1e-05        1.9700e-03   8.4772e-05   1.00000000000013        1
1e-06        1.9700e-04   8.4772e-06   0                       1
1e-07        1.9700e-05   8.4772e-07   0                       1
1e-08        1.9700e-06   8.4772e-08   0                       1
1e-09        1.9700e-07   8.4772e-09   0                       1
```

The same series held at a $100 price level, so that only the spread shrinks:

```
tick size    price level  sigma(x)     TA_CORREL               two-pass Pearson r
1e-02        1.0197e+02   8.4772e-02   1.00000000094499        1
1e-03        1.0020e+02   8.4772e-03   0.999999932500865       1
1e-04        1.0002e+02   8.4772e-04   1.00000000005832        1
1e-05        1.0000e+02   8.4772e-05   1.00053984751029        1
1e-06        1.0000e+02   8.4772e-06   0                       1
1e-07        1.0000e+02   8.4772e-07   0                       1
1e-08        1.0000e+02   8.4772e-08   0                       0.999999999999944
1e-09        1.0000e+02   8.4772e-09   0                       0.999999999998595
```

Both tables flip at the same place — `sigma(x)` between 8.5e-5 and 8.5e-6 — so this tracks
the **spread of the window, not the price level**. A $100.00 instrument quoted in 1e-6
increments reaches it.

`TA_CORREL` returns `TA_SUCCESS`, with the expected `outBegIdx` / `outNBElement`. Nothing
signals that the values are not a correlation.

## Reproduction

[...]

PR #419, opened 2026-09-14
feat(cti): add TA_CTI, Ehlers' Correlation Trend Indicator

kevinlincg:
Implements `TA_CTI` from the proposal card in [ta-lib-proposal-drafts#74](https://github.com/TA-Lib/ta-lib/issues/430), which I claimed there before starting. That card has no open questions and settles all six of its decisions, so this takes them as written and reports back on the two things it left to measurement.

## Not a literal transcription, and the card is right about why

The author's listing carries `n·Sxx − Sx·Sx` on raw price levels — the exact #242 cancellation that made `TA_CORREL` return `0`, `-1` and `-1.73` from perfectly correlated inputs. I reproduced both arms before building the gate, against a 60-digit `decimal` reference over the exact binary value of each double:

| form | worst absolute error, 1e2 level / 1e-5 spread |
|---|---|
| the listing, transcribed literally | **2.864e-01** |
| this implementation | **5.551e-17** |

Fifteen orders apart, on an indicator whose entire range is two wide. (The card states 5.7e-03 for the naive form on its own corpus; mine is a shorter window on uniform rather than Gaussian noise, so it is worse conditioned — the direction is what matters.)

So the price side carries `correl.c`'s shift-and-reseed with the same three triggers, and the ramp side collapses to exact constants: with `y` as bars-ago, `sumY = n(n−1)/2` and `ssY = n(n²−1)/12`, both computed in double because the integer forms overflow at the top of the parameter range. The y-weighted sum slides in O(1) by `linearreg.c`'s identity — advancing the window raises every retained weight by one and drops the departing value at full weight, which together are exactly `−n·trailingX`, with `sumX` read before it is decremented.

## The sign is the trap, and the mutation table is the evidence

The ramp runs backward in time, so the coefficient is negated once at the output. The card warns that dropping it inverts the whole indicator and that a magnitude assertion is blind to it. Measured, by removing the negation and regenerating:

| leg | negation removed |
|---|---|
| 2 — sign, rising series → positive | **RED** |
| 3 — exact ±1 endpoints | **RED** |
| 1 — differential vs `TA_CORREL(ramp)` | **RED**, worst diff 2.0 — the entire range |
| 5 — `\|CTI\| ≤ 1` | **green** |
| 6 — flat window → 0.0 | green |
| 7 — in-place aliasing | green |

Leg 5 staying green under a defect that reverses every single value is why the sign leg exists as a separate assertion rather than as a bound. I wrote it first, before tuning anything else, for that reason.

A second mutation covers the cancellation gate: replacing the shift with zero (i.e. reverting to the listing's arithmetic) turns leg 4 red with `0` against a reference of `0.0419` — and the `0` is itself the #242 signature, the guard deciding the window has no spread when it has one.

## Cross-language parity came out better than the card expected

The card anticipated a per-function xlang tolerance entry **rather than zero**, reasoning that CTI shares CORREL's kernel and its `sqrt`, and that `test_codegen.c:5877` gives CORREL `{ TOL_REL_OUT, 4e-9 }`. It also marked this "Not measured for CTI".

Measured: `ta_regtest --xlang-hash --function=CTI` gives **1948 Rust cases and 1948 Java cases, zero mismatches, BIT-IDENTICAL at zero tolerance**. No entry added. (C# not in that run — see below.)

Worth noting how I got there, because the first two attempts looked like passes and were not. `--codegen --function=CTI` prints `* All tests succeeded *` while doing **no value comparison at all** — CTI is post-cutover, so the frozen `ta_ref_serve` baseline has no row for it and every match is skipped. The tool says so itself, in a banner that reads `NO VALUE COMPARISON ... This is NOT a pass`, and names `--xlang-hash` as the gate that does compare. Worth knowing for any post-cutover function.

## Verified locally

- `ta_regtest` — full run green; `--function=CTI` exercises eight legs with coverage counters (sign, differential, endpoints, 21 cancellation, bounds, flat, alias, range sweep)
- `ta_regtest --xlang-hash --function=CTI` — bit-identical across C, Rust and Java at zero tolerance
- `ta_codegen/generator` — full suite green, no inventory needed updating
- Leg 1's tolerance is measured against the live kernel rather than carried in from the card (which says as much): worst absolute difference from shipped `TA_CORREL` over an ascending ramp is **5.893e-12** across periods 2..60 on a 400-bar walk, so the gate sits at 1e-9
- ABI: twelve added entry points, nothing removed or changed, `c+1:0:a+1` → `3:0:2`, soname unchanged

## Not verified locally

**C#.** This box's .NET SDK (7.0.200) cannot target `net10.0`. That backend is absent from the `--xlang-hash` run above as well as from the build, so its parity is established only by CI.

`check-abi` reports `EXPORTS UNCHECKED: no ELF shared library (Mach-O/PE not covered)` on macOS, so the export-table half of that gate runs only in CI.

Gate 8 of the card's plan — live pandas and Pine goldens — needs oracle arms I cannot run here.

## Note on the ABI line, shared with #413, #417 and #418

All four of my open PRs bump `TALIB_LIBRARY_VERSION` to `3:0:2` and write `function-count 2463`. The version stays correct however many land — libtool's generation is per release, so several batches of additions inside unreleased 0.8.2 are one bump — but the count does not: git takes the identical count edit from each side without raising a conflict, leaving a manifest that lists more entry points than it claims. `check-abi` catches it (exit 1, naming the right number), so each PR after the first goes red on `Public C ABI matches ABI.manifest` and needs one `scripts/build.py check-abi --update`. Details on #417.


kevinlincg:

|---|---|---|---|

kevinlincg:

|---|---|---|---|

kevinlincg:
**The pandas-ta-classic arm from the card, run.** Not via `ta_pandas_serve` — `ta-lib-oracles` is not a repo I can reach — but against `pandas-ta-classic` 0.8.32 installed into a scratch venv here.

**It turns out to be the card's warning, alive.** `pta.cti` is a wrapper over `linreg(close, length, r=True)`, and that path computes (`overlap/linreg.py:72-76`):

```python
rn = length * xy_sums - x_sum * y_sums
rd = (divisor * (length * y2_sums - y_sums**2)) ** 0.5
```

`length * y2_sums - y_sums**2` is `n·Sxx − Sx²` **on the raw price levels** — exactly the quantity the card says not to transcribe, and exactly #242. So this arm is not an independent check of the arithmetic; it is a specimen of the form this PR avoids.

Measured against a 60-digit decimal reference, worst absolute error over the window:

| corpus | `TA_CTI` | `pandas-ta-classic` |
|---|---|---|
| random walk, spread 7.7 over a level of 96.75 | 1.554e-15 | 3.426e-12 |
| the #242 regime, spread 9.5e-06 over a level of 100 | **5.551e-17** | **4.727e-02** |

On ordinary data the two agree to what pta can resolve. In the regime the card names, pta is off by 0.047 on an indicator whose entire range is two wide — about 2% of full scale — while the shift-and-reseed form holds 5.6e-17.

So: the arm agrees where it can, and where it cannot, the reference says which side moved. I am **not** proposing a golden from it, and I would not treat a future disagreement with pta on tight-spread data as evidence against this implementation.

Worth noting for the card: its oracle table lists pandas-ta-classic without qualification. On this evidence it is usable for structure — window alignment, lookback, sign — and not for values below roughly a 1e-6 spread-to-level ratio.

kevinlincg:
Rebased onto `86917b38e`.

CTI passes it:

```
no frozen-reference baseline (post-cutover): CTI  [sweep skipped 1; all still bitwise-gated by VARIANT]
```

kevinlincg:
reach the language servers.

## What is wired

Six legs: sign, endpoints, cancellation, bounds, flat, and the non-aliased call of the aliasing leg.

reproduce.

## Measured, not assumed

```
SV FAIL [CTI] (pipe 0, c): BITWISE mismatch vs in-process C
  retCode 0/0  begIdx 19/20  nbElem 21/20  hash f847c55ed8c406cc/e49517cb5277e38d
exit 22
```

and a bare `ta_regtest --function=CTI` is unaffected either way —

What this buys specifically: the #242 cancellation corpus — the data written
to break this function's numerics — was being checked in one language only.
A port that transcribed the author's listing literally errs by 2.864e-01 on it,
and nothing here would have noticed.

kevinlincg:
Trimmed `cti.md` to match the standard you set in efeda0bf3 and d4586ca78 on RVI/RVIR. Pushed as 5a8384fd3.

Removed the bars-ago orientation bullet and the shift-versus-raw-levels bullet; both are implementation reasoning and both already live in `cti.c:49-66` and `cti.c:156-162`. Also dropped the "implementations disagree here — the listing holds, CORREL gives 0, pandas NaN, Pine na" list, which is the same shape as the vendor-divergence bullets you cut from both RVI and RVIR.

The two things from those bullets that a caller actually needs moved into the Formula block: that `y` rises with time, so a rising series reads positive, and that a flat window returns 0. That follows the `or 50 when Up + Down = 0` line you added to `rvi.md`.

What I kept is the conditioning floor, the flat-window value with the one-clause reason it is not held over, the clamp, and why the period starts at 2.

mario4tier:
Merging with three commits on top:

- **fix(cti): rebuild against the peak sum of squares.** Two defects, both inherited from `correl.c`:
  - A series that decays back onto the shift left the running sums holding only rounding, and `ssX < 1e-6*sumX2` never fired. Fully flat windows came out nonzero, and near-flat ones were up to 0.91 from the exact correlation.
  - A window flat to within the rounding of its own mean re-fired the rebuild on every bar, so every flat bar cost O(n).
  
  The trigger now compares against the largest `sumX2` since the last rebuild, and such a window is re-anchored on one of its own values. Full-range output on the 252-bar corpus is bit-identical at every period. `CORREL`/`VAR`/`STDDEV` keep both defects: #434.
- **test(cti):** gate 8 of the card. The table holds 46 goldens at periods 5/12/20/50, checked three ways:
  - against an exact rational reference, at 1e-12;
  - against pandas-ta-classic 0.6.52 `ta.cti`, at 1e-10;
  - against Pine `ta.correlation(close, bar_index, n)` on PyneCore 6.8.14, at 1e-10.
  
  ta4j 0.22.6 and a transcription of the listing agree within the same bound. New legs cover the rebuild, a flat stretch after real data, period 2, history independence and the periodic rebuild. Disabling every rebuild used to leave the suite green.
- **docs(cti):** drops the conditioning-floor note (accuracy does not degrade as spread/level narrows: around a fixed level CTI matches the exact correlation to about 1e-15 down to 1e-14; its worst case anywhere is about 2e-10, set by the rebuild trigger), corrects the clamp note (overshoot up to about 1e-10, not a few ulp), adds `## Implementation` and `## Aliases`, and the CHANGELOG cites #430.

Two numbers in the original comments, now corrected: the literal listing errs by 2.05e-01 on the cancellation corpus (2.864e-01 is the oldest-first sum), and leg 1's corpus is the 252-bar close, where the worst is 8.9e-12.

C# parity holds here too: `--xlang-hash --function=CTI` is bit-identical for C#, and the C# suites pass.

For #417 and #418: `CTI.extrema` takes internal-error id 429, the same id as `KURTOSIS.extrema` and `MEDIAN.circbuf.ring`. #417's `next: 429 → 430` edit merges cleanly against this one and would ship a duplicate id, so regenerate `internal_error_ids.yaml` on rebase.

Issue #415, opened 2026-09-13
perf(java): rolling min/max family runs 6-9x slower than C on a 100k random walk


mario4tier:
## Problem

The Java batch tier of the rolling min/max family (MIN, MAX, MINMAX, MIDPOINT, MIDPRICE, WILLR, the #147 Van Herk block scan) runs **6 to 9x slower than C** on a 100k-bar random walk. MFI shows the same signature at 3x. Across the corpus Java batch is 1.13x C at the median, so these rows are the tail a Java user would notice, and the only place a JNI wrapper of the C library would beat the pure-Java jar.

## Measurements

Host: **laptop**, AMD Ryzen 7 PRO 8840U (Zen 4), WSL2, OpenJDK 21.0.12. Not yet measured on the Intel devbox. C is the shipped CMake Release `libta-lib.so`; Java is the public `Core` tier from dev `745247fe5`. One function per process, min of 25 rounds, best of 3 processes, period 14. Both arms produce identical output.

ns/bar:

| function | C 1k | Java 1k | Java/C | C 100k | Java 100k | Java/C |
|---|---|---|---|---|---|---|
| MIN | 0.74 | 2.48 | 3.3 | 0.79 | 7.31 | **9.3** |
| MAX | 0.80 | 2.20 | 2.8 | 0.79 | 7.43 | **9.4** |
| MINMAX | 1.52 | 4.02 | 2.6 | 1.64 | 14.01 | **8.6** |
| MIDPOINT | 1.52 | 3.97 | 2.6 | 1.59 | 13.53 | **8.5** |
| MIDPRICE | 1.52 | 4.06 | 2.7 | 1.63 | 14.03 | **8.6** |
| WILLR | 2.45 | 4.52 | 1.8 | 2.46 | 15.29 | **6.2** |
| MFI | 3.77 | 3.82 | 1.0 | 3.74 | 11.39 | **3.1** |
| SMA (control) | 1.27 | 1.28 | 1.0 | 1.34 | 1.33 | 1.0 |
| BBANDS (control) | 3.27 | 3.64 | 1.1 | 3.38 | 3.61 | 1.1 |

## What the numbers say about the cause

[...]

PR #420, opened 2026-09-14
test(bench): watch every arm's spread, not only the reference

kevinlincg:
This is the first of the two items I offered on #415 and never heard back on. It does not touch that issue's deliverables — the Java emission and its C2 output are untouched — it only makes the tool that measures them say when it cannot be trusted. If you would rather fold it into the #415 work, or not have it, close this and nothing is lost.

## What is wrong

`ta_bench` keeps the minimum of `BENCH_PASSES` per server and reports how far apart those passes landed, so a reader can tell whether the ratio columns mean anything. The comment on that accounting says as much: *"The ratio columns below are only as meaningful as this is small."*

But it only ever watched `cref`. `g_spread_sum/_worst/_n` is fed under `strcmp(LANGUAGES[li].name, "cref") == 0`, so the arm most likely to be moving is the one nothing is looking at — a managed runtime spends its first passes compiling, and its rows swing accordingly.

**MEASURED on this tree.** `MIN`, 100k points, random walk, five repeats of the identical command:

| `--iters` | Java arm run-to-run range | Java/C ratio across the five runs |
|---|---|---|
| 100 (default) | **7.2x** | 4.34, 5.26, 5.82, 8.56, 26.86 |
| 1000 | 1.16x | 4.18 – 4.94 |

At the default the ratio is not a measurement. I reported 11.3x for that row on #415 before I noticed, and the settled figure is 4.5x.

**Correction to an earlier claim of mine.** I first offered this as the explanation for the #415 rows that appeared in the corpus-wide pass and not in isolation (IMI at 1.15x, AD at 1.0x). It is not. Those are two separate invocations — run-to-run variance — and this gate's domain is within a single run, across `BENCH_PASSES`. Covering every arm would not have flagged them. What survives is the narrower code fact below: the accounting was `cref`-only, so no arm's within-run spread but the reference's was ever checked.

## What this changes

Each arm accumulates its own spread; the footer prints one line per active arm; `--max-spread` fires on the **worst** arm rather than on `cref` alone. A ratio is a quotient of two timings and is no better than the noisier of them.

`cref` keeps its separate role as the denominator of the ratio columns — that is untouched, as are the colouring, the `No C-ref column` wording, and `--verify-corpus`, which returns before any of this and is the only `ta_bench` invocation in CI.

## Control run

`--language=c,java --function=MIN --points=100000 --max-spread=10`:

| | exit | footer |
|---|---|---|
| **before**, `--iters=100` | 0 | `No C-ref column ... uncalibrated` — Java's 15% mean spread never examined |
| **after**, `--iters=100` | **1** | `mean Java spread 15% exceeds --max-spread=10%` |
| **after**, `--iters=1000` | 0 | `Java spread ... mean 3%` |

The before-row is the point: with no `cref` in `--language` the old gate had nothing to accumulate, so it passed on a run whose Java column was swinging 15% between passes. Both after-rows come from the same binary, so the difference between them is the data and not the build.

Also checked: with `cref` present the per-arm lines appear alongside it and the ratio columns are unchanged; `--verify-corpus` still exits 0; the full `ta_regtest` run is green.

## Not done here

The second item I offered on #415 — a `--help` note that the managed arms need `--iters` well above the default before their rows settle — is not in this PR. The error message now says it at the point of failure, which seemed like the better place, but say the word if you want it in the usage text too.


mario4tier:
Thanks Kevin, watching every arm is the right call. A few things before merging:

1. The `No C-ref column` warning is now unreachable whenever any other arm timed a row: it hangs off `if( any_arm )`. So `--language=c,java` prints an all-green, uncalibrated table with no notice (your own after-rows show it). Test `!LANGUAGES[0].active` on its own.
2. The stderr hint blames a managed runtime even when the worst arm is C-ref, C or Rust. Either print it only for java/csharp, or use a neutral remedy like `ta_bench_direct`'s "quiet the machine or raise --iters".
3. Comments: please keep only the rule (a ratio is no better than the noisier of its two timings). The history, the pasted measurements and the narration belong in the commit message, and the new loop comment restates the code.
4. Worth stating in the PR: the gate measures spread between passes within one run, so at the default `--max-spread=25` it does not catch the run-to-run swing you measured (Java sits around 15%). Also, a plain local `scripts/regtest.py` (perftest on) now exits 1 on Java noise at the default `--iters`. Fine if intended, but say so, or raise the iters regtest.py uses.

Nits: `timings[li] <= 0` can never be true (`has_timing` already implies > 0), and the `[16]` arrays could be sized from `NUM_LANGUAGES`.

kevinlincg:
All five addressed in 96cb7c4. Thank you — point 1 is a regression I introduced, and it was visible in the after-rows I pasted in this PR's own description. I read those rows for the exit code and not for what was missing from them.

**1. Unreachable notice.** It hung off the same if/else chain as the new per-arm block, so any arm with a timed row suppressed it. Now its own test on `!LANGUAGES[0].active`. Verified: `--language=c,java --max-spread=10` on SMA now prints the spread lines *and* `No C-ref column` — at `--iters=100` exiting 1, at `--iters=800` exiting 0, the notice present in both.

**2. Wording.** Now `ta_bench_direct`'s: *"these numbers are not trustworthy; quiet the machine or raise --iters"*. Checked on a run where the worst arm was C rather than a managed one — `--language=cref,c --max-spread=1` reports `mean C spread 14%` with no mention of a runtime.

**3. Comments.** Trimmed to the rule. The measurements and history are in the commit message. The loop comment that restated the code is gone. Net −17 lines against the first version.

**Nits.** `timings[li] <= 0` dropped (`has_timing` implies it); arrays sized from `NUM_LANGUAGES`.

**4. What the gate measures, and `regtest.py`.**

You are right that these are two different quantities and I conflated them. The gate measures spread *between `BENCH_PASSES` within one run*; the 7.2x I reported on #415 was *run-to-run*, across five separate invocations. At the default `--max-spread=25` the within-run figure often sits under the bar while the run-to-run ratio is still unusable, so **this gate does not catch the thing that produced my 11.3x mistake.** It catches a noisy box, which is a different and narrower claim than the PR made. I have not attempted the run-to-run case; that would need state across invocations.

On `regtest.py`: confirmed, `--test-only --no-regtest --function=SMA --language=c,java` exits 1 with `mean Java spread 31%`. MEASURED on this box, cref+c+java, SMA at 100k points:

| `--iters` | Java within-run spread | exit at the default `--max-spread=25` |
|---|---|---|
| 100 (ta_bench's default) | 64% | 1 |
| 200 | 21% | 0 |
| 400 | 10% | 0 |
| 800 | 7% | 0 |

So 400 is the smallest value with real headroom, and the cost is roughly 4x the perftest wall-clock for everyone running the loop.

I have **not** changed `regtest.py`, because that trade is yours rather than mine — and note this is not new behaviour in kind: `ta_bench_direct` already fails the same way on a busy machine, which is where the wording in point 2 comes from. Say which you want and I will push it:

- leave it, and I will add a line to the PR description saying a busy box or a default-`--iters` managed arm will now fail the local loop; or
- `regtest.py` passes `--iters=400` when the caller did not specify one; or
- `regtest.py` passes a looser `--max-spread` for its own quick-loop runs.

mario4tier:
Thanks Kevin, merged. I added one commit (1e500a09e) instead of asking you for another round.

The per-arm spread lines stay, so a noisy column is visible right under the table. `--max-spread` gates on cref only again. I measured a full-corpus perftest on my laptop: the Java arm's mean spread was 42% at `--iters=100` (6.6 min) and still 30% at `--iters=200` (15 min). Gating the worst arm would make every local `regtest.py` perftest fail, and raising `--iters` enough to clear it costs 30+ minutes, so `regtest.py` is unchanged.

